Issue #54, opened 2020-05-28
[doctor-visits] Use Jeffreys estimator for binomial proportion


capnrefsmmat:
The doctor visits pipeline should estimate the binomial proportion using the Jeffreys estimate, where

$$
\hat p = \frac{x + 0.5}{n + 1}
$$

That's the posterior mean if you put a Beta(0.5, 0.5) prior on p; see [here](https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Jeffreys_interval). This would be consistent with the Facebook community question and the Google survey pipeline.

When we report standard errors internally, these should be reported based on this estimate, to avoid ever reporting SE = 0.

We should use this for internal reports immediately (cc @huisaddison), but probably should not alter public estimates without a release note, unless you find the changes in estimates are very small compared to e.g. normal backfill.

PR #155, opened 2020-07-20
FB-survey validation with a generic design to include other pipelines

### LINKED ISSUE #59: Port the Facebook validation pipeline to be generic and automatable
Currently, `covid-19/facebook/prepare-extracts/covidalert-io-funs.R` contains a validation pipeline for Facebook. As I understand it, it does the following checks when we prepare a new day of data to upload.

1. Ensure that the *old* data in the API matches the newly generated old data; that is, if it's currently June 1, make sure we didn't unexpectedly change data for May 25th.
2. Do sanity checks on the new data, such as the geography types being reasonable, the geo_ids having the right format, the values and SEs being in the correct range, sample sizes are present, dates aren't missing, etc.
3. Verify the number of geographical regions reporting hasn't suddenly changed.
4. Verify the average variable values haven't suddenly changed.

Many of these checks can be made generic to multiple data sources and applied to our new pipeline. This would require

- [x] adapt the script to work on Taylor's directory structure where data files are placed
- [ ] provide for configuration files that specify the checks that apply to each data source
- [x] adapt the script to report all errors, rather than dying on the first one
- [ ] make it easy to automatically run for each ~all~ data source as a part of its automation job

amartyabasu:
Closes #59 

- The new validator reads the daily-files from the disk after pipeline execution. The validations for a given signal-type and geo-value  on given day are done against a range of previous days data, that get fetched from the API.
- The validator does basic checks related to different columns, column types, missing_dates, numeric-constraints, null values.
- It also checks for upward and downward spikes of data, variations in mean differences of column values grouped by geo_ids


validator/delphi_validator/datafetcher.py
```diff
@@ -1,95 +1,145 @@
...
-    # Todo - Include weighted whh_cmnty_cli and wnohh_cmnty_cli
-    for sig in unique_signals:
-        if "community" in sig:
-            unique_signals.remove(sig)
+    if data_source == 'fb-survey':
+        # Currently metadata returns --*community*-- signals that don't get generated
+        # in the new fb-pipeline. Sieving them out for now.
+        for sig in unique_signals:
+            if "community" in sig:
+                unique_signals.remove(sig)
```
krivard, line 66: (this is no longer the case)

```suggestion
```

validator/delphi_validator/validate.py
```diff
@@ -1,237 +1,721 @@
...
+    read_filenames, load_csv, get_geo_sig_cmbo, \
+    read_geo_sig_cmbo_files, fetch_api_reference
+
+# Recognized geo types.
+geo_regex_dict = {
+    'county': '^\d{5}$',
+    'hrr': '^\d{1,3}$',
+    'msa': '^\d{5}$',
+    'state': '^[a-z]{2}$',
+    'national': '^usa$'
```
krivard, line 24: `national` requires a two-character country code

```suggestion
    'national': '^[a-z]{2}$'
```

validator/delphi_validator/validate.py
```diff
@@ -1,237 +1,721 @@
...
+                ("check_max_date_vs_reference", checking_date.date(), geo, sig),
+                (df_to_test["time_value"].max(),
+                 df_to_reference["time_value"].max()),
+                'reference df has days beyond the max date in the =df_to_test=; ' +
+                'checks are not constructed to handle this case, and this situation ' +
+                'may indicate that something locally is out of date, or, if the local ' +
+                'working files have already been compared against the reference, ' +
+                'that there is a bug somewhere'))
+
+    def check_rapid_change(self, df_to_test, df_to_reference, checking_date, geo, sig):
```
krivard, line 452: Rename to specify that it's a change in num_rows? this is not the only rapid change we might test for.

validator/delphi_validator/validate.py
```diff
@@ -1,237 +1,721 @@
...
+            mean_type_diff=("type_diff", "mean"),
+            mean_abs_type_diff=("abs_type_diff", "mean"),
+            mean_test_var=("test", "mean"),
+            mean_ref_var=("reference", "mean")
+        ).assign(
+            mean_stddiff=lambda x: 2 *
+            x["mean_type_diff"] / (x["mean_test_var"] + x["mean_ref_var"]),
+            mean_stdabsdiff=lambda x: 2 *
+            x["mean_abs_type_diff"] / (x["mean_test_var"] + x["mean_ref_var"])
+        )[["variable", "mean_stddiff", "mean_stdabsdiff"]]
```
krivard, line 535: This is a super-gnarly bit of code; make sure it's thoroughly tested

validator/delphi_validator/validate.py
```diff
@@ -1,237 +1,721 @@
...
+                geo_sig_cmbo,
+                export_dir,
+                [name_match_pair[0] for name_match_pair in validate_files],
+                date_slist):
+
+            match_obj = smooth_option_regex.match(sig)
+            smooth_option = match_obj.group(1)
+
+            if smooth_option not in ('raw', 'smoothed'):
+                smooth_option = 'smoothed' if '7dav' in sig or 'smoothed' in sig else 'raw'
```
krivard, line 648: This is likely to be brittle -- need to keep in mind when naming future signals that the validator will assume raw unless one of these keywords is present. Probably fine for now but we may want to make the smoothed keyword and/or default behavior configurable some time in the future.

validator/delphi_validator/validate.py
```diff
@@ -1,237 +1,721 @@
...
+            - max_date: date of most recent data to be validated; datetime format.
+            - weighted_option: str; selects the "reasonable" threshold based on signal name
+            - geo: str; geo type name (county, msa, hrr, state) as in the CSV name
+            - sig: str; signal name as in the CSV name
+
+        Returns:
+            - None
+        """
+        switcher = {
+            'unweighted': timedelta(days=1),
```
krivard, line 395: Is this essentially the assumed lag? doctor-visits, hospital-admissions, and quidel will all have much longer minimum lags than this.

validator/delphi_validator/validate.py
```diff
@@ -1,237 +1,721 @@
...
+                [name_match_pair[0] for name_match_pair in validate_files],
+                date_slist):
+
+            match_obj = smooth_option_regex.match(sig)
+            smooth_option = match_obj.group(1)
+
+            if smooth_option not in ('raw', 'smoothed'):
+                smooth_option = 'smoothed' if '7dav' in sig or 'smoothed' in sig else 'raw'
+
+            weight_option = 'weighted' if 'wili' in sig or 'wcli' in sig else 'unweighted'
```
krivard, line 650: If this is about expected lag and nothing else, rename. Surveys is the only indicator with weighted/unweighted variants, but it's _not_ the only indicator with differences in expected lag between signals.

Also: better make the signal name keywords configurable, since even Surveys has more than just wili/wcli now.

validator/delphi_validator/validate.py
```diff
@@ -1,237 +1,721 @@
...
+        """
+        If any not-suppressed exceptions were raised, print and exit with non-zero status.
+        """
+        if self.raised_errors:
+            suppressed_counter = 0
+            subset_raised_errors = []
+
+            for val_error in self.raised_errors:
+                raised_check_id = tuple(item.strftime("%Y-%m-%d") if isinstance(
+                    item, (date, datetime)) else item for item in val_error.check_data_id)
```
krivard, line 702: Add parentheses to clarify whether this is `(str, <generator>)` or `(str, str, ..., str)` -- operator precedence in this context varies across languages

validator/params.json.template
```diff
@@ -1,10 +1,13 @@
...
   "validation": {
-  "data_source": "fb_survey",
-  "start_date": "2020-06-13",
-  "end_date": "2020-06-19",
-  "ref_window_size": 7,
-  "data_folder": "C:/Covidcast/covidcast-indicators/validator/data",
-  "minimum_sample_size": 100
+    "data_source": "usa-facts",
+    "start_date": "2020-09-05",
+    "end_date": "2020-09-08",
```
krivard, line 5: We discussed this adequately this morning so just noting here for posterity: prefer to configure end_date and span_length (doctor-visits produces updates on ~6 weeks of data at each issue), with special end_date value like "latest" or "new" to indicate should just use date.today().

krivard (commented):
Just based on a read-through, not running or testing

validator/delphi_validator/validate.py
nmdefries: I'll add this and the new "lag_option" (or similar name) as user settings.

validator/delphi_validator/validate.py
nmdefries: My understanding is that there are two different types of "lag".

1. One is the number of days behind a data source is from being current. So, for example, a centered 7-day average will always be 3-4 days behind *now* because of how it's calculated.
2. The other type of lag, to be addressed with the new "scan_length" setting, happens when case (or other metric) reports are delayed compared to when they happen, causing recent already-released data to be updated.

 We're dealing with the first type here (going to be renamed to "lag_option" or similar), so I'm surprised that it varies widely. I thought the goal was to have all the indicators update on a daily basis and not be more than a few days out of date.

I'll plan to add this as a user setting so it can be set on a per-signal basis.

validator/params.json.template
nmdefries: Does it make sense to check the not-yet-finalized (i.e. still in the *x* week update window) data? Presumably, it will raise a lot of errors, especially the most-recent data, for being very different than historical values. And we'll also end up checking those dates multiple times, as they move from less-finalized to more-finalized.

validator/README.md
```diff
@@ -1,36 +1,44 @@
...
-   source. 
-2) Its does a comparative analysis with recent data from the API 
-   to detect any anomalies such as spikes, significant value differences
+1) Sanity checks on daily data generated from the pipeline of a specific data
+   source.
+2) Comparative analysis with recent data from the API
+   to detect any anomalies, such as spikes or significant value differences
 
-The validator validates against daily data thats already written in the disk
+The validator validates new source data against daily data that is already written to disk,
```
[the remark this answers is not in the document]
nmdefries: Whoops, thanks for the catch!

validator/delphi_validator/validate.py
krivard: Ah! Good note.

In the COVIDcast API, we define lag to be the number of time units between when an event occurred and when data about those events were published. It's a purely operational definition: `issue - time_value`.

Delphi seems to collectively define "backfill" as the effect where previously-reported values are updated as additional data about those events becomes available. Backfilled data by definition has greater lag than what was first reported.

These two terms do not exactly match the two definitions you're familiar with, but they have roughly the same coverage (with the possible exclusion of filter lag, but then the API hasn't had to explicitly handle that).

krivard: > I thought the goal was to have all the indicators update on a daily basis and not be more than a few days out of date.

That was the goal in early April, but it quickly became clear that some sources wouldn't have data, or wouldn't have enough data, available quickly enough. iirc doctor-visits first publishes at 5 days out with 70% of the data it will ultimately have about that day; safegraph data has been up to 7 days out, etc.

nmdefries: Perfect, thanks for the clarification!

validator/params.json.template
krivard: I see what you mean, but if a 100x magnitude change happens during the backfill window, we'll want to know about it as soon as possible. One thing we can do when it comes time to install this in each indicator is _run_ the validator but not _believe_ the validator for the first week or so, just to see what kinds of false positives are raised. Then we can revise accordingly.

nmdefries: > Does it make sense to check the not-yet-finalized (i.e. still in the x week update window) data? Presumably, it will raise a lot of errors

Sounds like this is a non-issue. I was imagining most-recent data starting off with only, say 5% of the data ultimately available for that day, but you said above it's more like 70%. So high missingness, low sample size leading to biased values, etc are unlikely.

validator/delphi_validator/validate.py
nmdefries: User can now specify a list of `smoothed_signals` in the params file. If a signal name appears in the list, the lower threshold is used when comparing test and reference (API) avg values; otherwise, the signal is assumed to be "raw" and the higher threshold is used.

validator/delphi_validator/datafetcher.py
```diff
@@ -0,0 +1,138 @@
...
+from datetime import datetime
+from itertools import product
+import pandas as pd
+import numpy as np
+
+import covidcast
+from .errors import APIDataFetchError
+
+filename_regex = re.compile(
+    r'^(?P<date>\d{8})_(?P<geo_type>\w+?)_(?P<signal>\w+)\.csv$')
```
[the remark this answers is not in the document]
benjaminysmith: Or maybe the validator should accept the regex as a parameter?

validator/README.md
```diff
@@ -0,0 +1,112 @@
+# Validator
+
+The validator performs two main tasks:
```
benjaminysmith, line 3: Thanks for writing such through documenation. How are you planning on integrating this with the rest of the system -- will this be added as an extra automation step?

validator/delphi_validator/validate.py
```diff
@@ -0,0 +1,742 @@
...
+geo_regex_dict = {
+    'county': '^\d{5}$',
+    'hrr': '^\d{1,3}$',
+    'msa': '^\d{5}$',
+    'state': '^[a-z]{2}$',
+    'national': '^[a-z]{2}$'
+}
+
+
+def reldiff_by_min(x, y):
```
benjaminysmith, line 28: A style question: It looks like some functions have abbreviations and some not (e.g., reldiff vs relative_difference) is there a delphi coding standard on this? If not would suggest making consistent within the file.

validator/delphi_validator/validate.py
```diff
@@ -0,0 +1,742 @@
...
+        elif self.missing_se_allowed:
+            result = df_to_test.query(
+                '~(se.isnull() | ((se > 0) & (se < 50) & (se <= se_upper_limit)))')
+
+            if not result.empty:
+                self.raised_errors.append(ValidationError(
+                    ("check_se_missing_or_in_range", nameformat),
+                    result, "se must be NA or in (0, min(50,val*(1+eps))]"))
+
+        result_jeffreys = df_to_test.query('(val == 0) & (se == 0)')
```
benjaminysmith, line 339: Is "jefferys" an established term for this?

validator/delphi_validator/validate.py
```diff
@@ -0,0 +1,742 @@
...
+        result_alt = df_to_test.query('se == 0')
+
+        if not result_jeffreys.empty:
+            self.raised_errors.append(ValidationError(
+                ("check_se_0_when_val_0", nameformat),
+                None,
+                "when signal value is 0, se must be non-zero. please "
+                + "use Jeffreys correction to generate an appropriate se"))
```
benjaminysmith, line 347: Maybe add a reference here if "Jeffreys correction" is not widely understood.

validator/delphi_validator/validate.py
```diff
@@ -0,0 +1,742 @@
...
+            reference_rows_per_reporting_day)
+
+        if abs(compare_rows) > 0.35:
+            self.raised_errors.append(ValidationError(
+                ("check_rapid_change_num_rows", checking_date, geo, sig),
+                (test_rows_per_reporting_day, reference_rows_per_reporting_day),
+                "Number of rows per day (-with-any-rows) seems to have changed " +
+                "rapidly (reference vs test data)"))
+
+    def check_avg_val_diffs(self, df_to_test, df_to_reference, checking_date, geo, sig):
```
benjaminysmith, line 486: This looks pretty complicated, but also seems like a technique that other people may be doing too. Do you know if there are there any anomaly detection/comparison packages we could take advantage of for this?

validator/delphi_validator/validate.py
```diff
@@ -0,0 +1,742 @@
...
+                ("check_test_vs_reference_avg_changed",
+                 checking_date.date(), geo, sig),
+                (mean_stddiff_high, mean_stdabsdiff_high),
+                'Average differences in variables by geo_id between recent & reference data '
+                + '(either semirecent or from API) seem large --- either large increase '
+                + 'tending toward one direction or large mean absolute difference, relative '
+                + 'to average values of corresponding variables.  For the former check, '
+                + 'tolerances for `val` are more restrictive than those for other columns.'))
+
+    def validate(self, export_dir):
```
benjaminysmith, line 592: It looks like this will only print output on failed checks (maybe I'm missing something though). Did you consider also printing out stats on success? It would be reassuring to see that N out N checks passed, for example, to confirm they are being run.

validator/delphi_validator/validate.py
```diff
@@ -0,0 +1,742 @@
...
+                if raised_check_id not in self.suppressed_errors:
+                    subset_raised_errors.append(val_error)
+                else:
+                    self.suppressed_errors.remove(raised_check_id)
+                    suppressed_counter += 1
+
+            print(len(subset_raised_errors), "messages")
+            print(suppressed_counter, "suppressed messages")
+
+            if len(subset_raised_errors) == 0:
```
benjaminysmith, line 734: It looks like you can remove this block since it will be handled by the last one anyway

validator/PLANS.md
```diff
@@ -0,0 +1,53 @@
+# Validator checks and features
+
+## Current checks for indicator source data
```
benjaminysmith, line 3: Just curious -- have you tried running this on existing imports already?

Specifically I'm wondering if you have a sense of how often this generates false positives (if any) so that the aggressiveness of the validations can be tuned.

validator/delphi_validator/datafetcher.py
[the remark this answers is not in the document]
krivard: Shared with epidata; there will be a lot of things duplicated between this package and the validation done by the epidata acquisition code. We should pull it out into a library as soon as we have a [private package repo](https://github.com/cmu-delphi/covidcast-indicators/issues/261) up and running.

validator/README.md
krivard: Yes. The intended pipeline is:
1. Run indicator
2. Run validator <- this package
3. Run archiver <- [completed but not yet integrated](https://github.com/cmu-delphi/covidcast-indicators/pull/295)

validator/delphi_validator/validate.py
krivard: No standard except snake_case and appeasing pylint :)

validator/delphi_validator/validate.py
krivard: Yes; see [prior work](https://github.com/cmu-delphi/covidcast-indicators/issues/54) and [documentation]9https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/hospital-admissions.html); ["Jeffreysized"](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/indicator-combination.html) is also used

validator/delphi_validator/validate.py
nmdefries: >  are there any anomaly detection/comparison packages we could take advantage of for this?

Good thought! I'm aware of a few packages, but they are focused on modeling-based outlier detection, which might be a bit heavy-weight for our purposes here. I'll look into this and see if anything is promising.

validator/delphi_validator/validate.py
nmdefries: That's right; currently, validation outputs the # of failed checks, the # of suppressed checks (would have failed, but they've been verified as false positives or otherwise not worth blocking data upload for), and messages from failed checks. It's a good idea to add the number of passed and total checks, as well.

validator/PLANS.md
nmdefries: Yes! I have run this on the USAFacts indicators.

The old Facebook-specific validator had false positives from the check comparing average values of source data vs of the API. I increased the static threshold used in this check to require a larger difference in averages to trigger the error, and haven't yet had problems with it being too sensitive. 

I do often see the "val column can't have any cell smaller than 0" error. Although we expect to see negative values *occasionally*, used to adjusted for previous overreporting, it turns out they are fairly common. These aren't false positives per se, but perhaps given the frequency of the error, we would only want to report it if there were a very large negative value or many small negative values.

krivard: I'm going to put @chinandrew on surveying the other signals.

PR #176, opened 2020-07-31
Add the smoothing utility

dshemetov:
This PR adds the smoothing utility and its tests as part of #171. The utility has not replaced the previous smoother implemetations in the rest of the code base, as I received the suggestion to do those in separate PRs.


_delphi_utils_python/delphi_utils/smooth.py
```diff
@@ -0,0 +1,434 @@
...
+import numpy as np
+
+
+class Smoother:
+    """
+    This is the smoothing utility class. This class holds the parameter settings for its smoother
+    methods and provides reasonable defaults. Basic usage can be found in the examples below.
+
+    The smoother function takes numpy arrays as input, expecting the values to come from a
+    regularly-spaced time grid. NANs are ok, as long as the array does not begin with a NAN. The
```
nmdefries, line 24: Are you expecting to see time series beginning with NaN? If so, it might be worth explicitly trimming them off rather than worrying about how they might cause problems.

_delphi_utils_python/delphi_utils/smooth.py
```diff
@@ -0,0 +1,434 @@
...
+            A vector of coefficients of length nl that determines the savgol
+            convolution filter.
+        """
+        coeffs = self.savgol_coeffs(
+            -len(signal) + 1, 0, self.poly_fit_degree, self.gaussian_bandwidth
+        )
+        return signal @ coeffs
+
+    @classmethod
+    def savgol_coeffs(cls, nl, nr, poly_fit_degree, gaussian_bandwidth=100):
```
nmdefries, line 284: I can't see that this function is ever used as a class method (although I'm not familiar with this functionality).

_delphi_utils_python/delphi_utils/smooth.py
```diff
@@ -0,0 +1,434 @@
...
+        if gaussian_bandwidth is None:
+            mat_inverse = np.linalg.inv(A.T @ A) @ A.T
+        else:
+            weights = np.exp(-((np.arange(nl, nr + 1)) ** 2) / gaussian_bandwidth)
+            mat_inverse = np.linalg.inv((A.T * weights) @ A) @ (A.T * weights)
+        window_length = nr - nl + 1
+        basis_vector = np.zeros(window_length)
+        coeffs = np.zeros(window_length)
+        for i in range(window_length):
+            basis_vector = np.zeros(window_length)
```
nmdefries, line 332: Defined basis_vector twice; first is unused.

_delphi_utils_python/delphi_utils/smooth.py
dshemetov: I think I have an error if the series begins with a NaN. I wanted the smooth methods to preserve the length of the array, expecting the user to insert the output back into a dataframe with a fixed index.

dshemetov: Line 415. Though now I think it makes sense to pull that check out to `__init__`.

_delphi_utils_python/delphi_utils/smooth.py
dshemetov: Thanks for catching the different default. I think the `@classmethod` is left over from when I was testing this class and wanted to call this function without instantiating an object. At this point, it's probably unlikely anyone else will call it independently, so we can probably remove that.

nmdefries (commented):
Looks good to me. Nice documentation -- appreciate the example uses!

A couple tests are failing:

```
FAILED test_smooth.py::TestSmoothers::test_left_gauss_linear_smoother - ValueError: The window_length must be smaller than the length of the signal.
FAILED test_smooth.py::TestSmoothers::test_causal_savgol_smoother - ValueError: The window_length must be smaller than the length of the signal.
```

dshemetov:
@nmdefries incorporated your suggestions, thanks!

krivard:
I'm still showing the following failing test; do you have changes you haven't pushed up to github yet? It looks like there was conversation after your last commit.

```
self = <test_smooth.TestSmoothers object at 0x7fe665caeb50>

    def test_impute(self):
        # test the nan imputer
        signal = np.array([i if i % 3 else np.nan for i in range(1, 40)])
        assert np.allclose(Smoother(impute_method=None).impute(signal), signal, equal_nan=True)

        # test the zeros imputer
        signal = np.array([i if i % 3 else np.nan for i in range(1, 40)])
        assert np.allclose(
            Smoother(impute_method="zeros").impute(signal),
            np.array([i if i % 3 else 0.0 for i in range(1, 40)])
        )

        # make a signal with periodic nans to test the imputer
        signal = np.array([i if i % 3 else np.nan for i in range(1, 40)])
        # test that the non-nan values are unchanged
        not_nans_ixs = np.bitwise_xor(np.isnan(signal, where=True), np.full(len(signal), True))
        smoothed_signal = Smoother().savgol_impute(signal)
        assert np.allclose(signal[not_nans_ixs], smoothed_signal[not_nans_ixs])
        # test that the imputer is close to the true line
>       assert np.allclose(range(1, 40), smoothed_signal, atol=0.5)
E       assert False
E        +  where False = <function allclose at 0x7fe67033dee0>(range(1, 40), array([ 1.        ,  2.        ,  2.40533001,  4.        ,  5.        ,\n        6.47948675,  7.        ,  8.        , ...      , 32.        , 33.01427941, 34.        , 35.        ,\n       36.00357408, 37.        , 38.        , 38.99659138]), atol=0.5)
E        +    where <function allclose at 0x7fe67033dee0> = np.allclose
E        +    and   range(1, 40) = range(1, 40)

test_smooth.py:133: AssertionError
```

dshemetov:
This is really strange... I have that test, it's passing on my end, and I have pushed everything.

dshemetov:
Checked out the conflict. The incoming change is the correct one, the rest was refactored into functions.

dshemetov:
The commit hash I'm on locally is the same as the one here. Stumped.

dshemetov:
While this is blocked, I added an interface for handling pd.Series. This should let indicators use the [transform method](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.DataFrameGroupBy.transform.html) and not worry about numpy conversions.

krivard:
What version of python are you using? Prod is 3.8.2; my local is 3.8.5.

dshemetov:
I'm using 3.8.5.

krivard:
Dmitry and I debugged this offline. Reproduced test failure on a fresh clone and fresh python environment on my machine; reproduced passing condition on a fresh clone and fresh python environment (two different kinds!) on his with matching Python version; reproduced passing condition on a fresh clone and fresh python environment on prod with `indicator`'s Python. Prod wins; I give up. 👿 👻 👺

Issue #215, opened 2020-08-12
geomap refactor refactor


dshemetov:
Some outstanding fixes for the geomap utility refactor #137 
- [x] documentation
- [x] remove full flag
- [x] cross files
- - [x] zip -> county : population weighted
- - [x] zip -> fips : population weighted
- - [x] zip -> hrr : unweighted
- - [x] county -> state : unweighted
- - [x] county -> msa : unweighted
- - [x] county -> megacounty
- - [x] fips -> hrr
- - [x] zip -> state
- - [x] zip -> msa
- - [x] state code -> hhs region number
- - [x] state code -> us national
- [x] rethink nan handling. In many to many mappings (like zip to fips), we zero-fill nans in the aggregation step. This conflicts with smoothing, where it would best to distinguish between true zeros and zero filled nans.
- - Update: no better solution found than to zero-fill when aggregating.

PR #217, opened 2020-08-13
Further refactoring for the geo coding utility

dshemetov:
The work for this #215


chinandrew (approved):
Tests pass for me now. Linter yelled at some lines >100char, though not sure if we're using 100 or 120... otherwise looks good.

dshemetov:
@chinandrew I think those are the long URL lines. Is there are a preferred way to handle those?

chinandrew:
> @chinandrew I think those are the long URL lines. Is there are a preferred way to handle those?

if its a URL it's probably fine to leave it. might be able to do backslashes but that might actually make things less readable. Some of thse look like normal dosctrings though

Screenshot from 2020-09-30 14-24-57

dshemetov:
Oh oops, thanks for catching those. Just fixed those up.

dshemetov:
Just realized there are some renamings in this refactor that need to be smoothed out with the previous extension of the utility to JHU. Should be just naming fixes.

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -1,825 +1,1252 @@
...
+        state_codes = ["state_code", "state_id", "state_name"]
+
+        if not is_string_dtype(df[from_col]):
+            if from_code in ["fips", "zip"]:
+                df[from_col] = df[from_col].astype(str).str.zfill(5)
+            else:
+                df[from_col] = df[from_col].astype(str)
+
+        # Assuming that the passed-in records are all United States data
+        if new_code is "national":
```
chinandrew, line 274:
```suggestion
        if new_code == "national":
```

krivard:
Merge is pending completion of [corresponding documentation](https://github.com/cmu-delphi/delphi-epidata/pull/231)

jhu/delphi_jhu/geo.py
```diff
@@ -27,22 +27,18 @@ def geo_map(df: pd.DataFrame, geo_res: str):
...
-        df = gmpr.county_to_hrr(df, fips_col="fips", hrr_col="geo_id", date_col="timestamp")
-        df['geo_id'] = df['geo_id'].astype(int)
-    elif geo_res == 'county':
+    if geo_res == "fips":
         df.rename(columns={'fips': 'geo_id'}, inplace=True)
+    elif geo_res == "state":
+        df = gmpr.replace_geocode(df, "fips", "state_id", new_col="geo_id", date_col="timestamp")
+    else:
+        df = gmpr.replace_geocode(df, "fips", geo_res, new_col="geo_id", date_col="timestamp")
+
```
krivard, line 41:
```suggestion

```
(remove trailing whitespace)

krivard (approved):
* delphi_utils lints and tests
* jhu lints (with python3.6; 3.7+ still has a pytest recursion problem) and tests
* emr_hosp is a retired signal, is exempt from linting, but tests pass
* verified no other signals are currently using GeoMapper

PR #260, opened 2020-09-02
fb-package ready for production

krivard:
This is the version that will begin running Thursday 3 September.

This may warrant a rebase to clean up the commit history, since it's been a very long process.


capnrefsmmat:
@nmdefries I think it'd be good for us to pick up this PR and get fb-package merged. You'll want to coordinate with @krivard to ensure her build scripts start using main instead of fb-package, and then we can do further improvements with PRs off of main like normal people.

PR #287, opened 2020-09-28
[sir-complainsalot] Add grace period and dry run

krivard:
* A method for sir-CAL to complain that's not blasting Slack, for testing purposes
* The ability to tell sir-CAL not to bug anybody until results are N days late, for configurable N (currently global, might consider per-source or per-signal versions if we need them)


krivard:
That, but also to distinguish between when we usually expect a signal to become available, when it's nominally late but not worrisome, and when somebody should really take a look.

PR #294, opened 2020-10-06
Fb package test rebase

krivard:
Would replace #260.

Rebased to only two commits, but kept a large number of the commit notes for posterity.

Questions are:

* did this go too far?
* would it be better to split the R package changes off from the fetch/publish pipeline changes?


dshemetov:
@krivard just noticed this in the backlog. This is probably even harder to read now, isn't it? Is this code still waiting to be released or is this an outdated PR?

capnrefsmmat:
We'd probably want to redo the squashing (or just merge) from the current tip of fb-package.

krivard:
> We'd probably want to redo the squashing (or just merge) from the current tip of fb-package.

agreed

PR #297, opened 2020-10-07
Propagate geomapper changes into JHU

krivard:
#217 merged into main because delphi_utils doesn't have an associated deployment branch. This PR applies those changes to JHU automation.


korlaxxalrok (approved):
Tests pass in Jenkins (though we are ignoring linting because of that weird issue) so looks good to me 👍 :shipit:

PR #301, opened 2020-10-08
Fix JHU's fips-population source file

dshemetov:
* use geocode utility source file and remove old static source file
* update tests
* update validation notebook to verify prop signals in NYC


krivard:
It could've gone either way, but I switched the base to `deploy-jhu` because this PR primarily concerns the JHU indicator. Once we merge it, we'll propagate the changes to `main`.

testing_utils/indicator_validation.ipynb
```diff
@@ -172,7 +172,130 @@
...
+     "output_type": "stream",
+     "name": "stdout",
+     "text": [
+      "Total difference:  2.3797075421327918e-12\n"
+     ]
+    },
+    {
+     "output_type": "display_data",
+     "data": {
+      "text/plain": "<Figure size 1152x432 with 1 Axes>",
```
sarah-colq, line 241: Should we be able to see the remote line or are they identical here?  I cannot see the remote line on any of the new charts.

sarah-colq (changes requested):
Please check the charts in the python notebook.  Otherwise, looks good! :)

dshemetov:
That notebook isn't really supposed to be tracked. It was meant to be a template, but I kept working in the template and committed changes. I've reverted it back to its template version and renamed it accordingly.

krivard (approved):

PR #304, opened 2020-10-09
Fix JHU bug that renamed fips to county in receiving

dshemetov:
* change valid geo_res
* change GEO_RESOLUTIOSN in run.py
* update tests to use "county"


krivard (approved):
Lints (with python3.6), tests pass, ran it and all the `receiving` files look good. Huzzah!

PR #307, opened 2020-10-12
Add a stopgap fix to USAFacts deaths from October 2

dshemetov:
* set all county deaths values on Oct 2nd to the values from Oct 1st


krivard:
Issue date, not event date, turns out to have been the problem -- closing.

Issue #273, opened 2020-09-11
Add weekday/weekend SafeGraph signals


RoniRos:
- [ ] A smooth "weekday mobility" signal, e.g. a trailing average of the last 5 weekdays (skipping over weekend days).  Using 5 will average out variations due to differences among M/T/W/R/F.  This signal should not be available for Sat or Sun.
- [ ] A smooth "Saturday mobility" and "Sunday mobility" signals.  These are tougher to design because weekend days are more separated in time from one another.   Perhaps an exponential decay with a decay constant of several weeks?  Alternatively we can ditch these signals, and let people derive their own the way they like it.

PR #309, opened 2020-10-13
Add 7-day average of each signal to Safegraph indicator

### LINKED ISSUE #271: Add 7-day average versions of SafeGraph signals
Larry says forecasting has been using something similar, and they're extremely useful.

sgsmob:
Fixes #271.

For each signal at a geo resolution output as part of the `delphi_safegraph` module, create another version of the signal that is averaged over the previous 7 days.

Also cleans up test and comments.


safegraph/delphi_safegraph/process.py
```diff
@@ -1,13 +1,51 @@
...
 
 from .constants import HOME_DWELL, COMPLETELY_HOME, FULL_TIME_WORK, PART_TIME_WORK
-from .geo import FIPS_TO_STATE
+from .geo import FIPS_TO_STATE, VALID_GEO_RESOLUTIONS
 
 # Magic number for modular arithmetic; CBG -> FIPS
 MOD = 10000000
 
+
+def validate(df):
```
benjaminysmith, line 15: Is this (and below) something that can be shared across multiple indicators?

safegraph/tests/test_process.py
```diff
@@ -62,5 +68,47 @@ def test_handle_wip_signal(self):
         assert signal_names[0].startswith("wip_")
         assert all(not s.startswith("wip_") for s in signal_names[1:])
 
+    def test_files_in_past_week(self):
```
benjaminysmith, line 71: Are failure cases tested too? (e.g., missing date files)

safegraph/delphi_safegraph/process.py
```diff
@@ -1,13 +1,51 @@
...
+    """
+    path, year, month, day, _ = current_filename.rsplit('/', 4)
+    current_date = datetime.date(int(year), int(month), int(day))
+    one_day = datetime.timedelta(days=1)
+    for _ in range(1, 7):
+        current_date = current_date - one_day
+        date_str = current_date.isoformat()
+        date_path = date_str.replace('-', '/')
+        new_filename = f'{path}/{date_path}/{date_str}-'\
+            'social-distancing.csv.gz'
```
benjaminysmith, line 45: Should this be a constant?

benjaminysmith (commented):
It looks like there are bunch of changes in structure and logic here -- do you have a sense of the test coverage?

safegraph/delphi_safegraph/process.py
sgsmob: Possibly.  I'm working on a survey of the indicators to see what sorts of libraries we can or would want to refactor out.  Would prefer to do in a followon PR.

safegraph/delphi_safegraph/process.py
sgsmob: You mean that "social-distancing.csv.gz" should be refactored as a top-level constant that is interpolated here too?

safegraph/tests/test_process.py
sgsmob: This function does not do any testing of file existence.  I should add file existence testing to `process()`

safegraph/delphi_safegraph/process.py
benjaminysmith: Yeah.

safegraph/tests/test_process.py
benjaminysmith: Cool -- yeah because it looks like this is new logic it's hard to tell if this introduces any new edge cases so would suggest thorough testing if it is not there already.

safegraph/delphi_safegraph/process.py
benjaminysmith: +1 that's exactly what I was wondering. Agree definitely not for this PR.

safegraph/tests/test_process.py
sgsmob: Added more test cases across year boundary and leap day checking

safegraph/delphi_safegraph/process.py
sgsmob: Done

sgsmob:
Just added a test case for process which should give us some more confidence on the correctness.

safegraph/tests/test_process.py
```diff
@@ -62,5 +70,112 @@ def test_handle_wip_signal(self):
...
+        expected = pd.DataFrame(data={
+            'geo_id': [1053, 1073],
+            'val': [0.04, 0.14],
+            'se': [0.02, 0.10],
+            'sample_size': [2, 2]
+        })
+        actual = pd.read_csv(
+            export_dir / '2020-02-14_county_completely_home_prop.csv')
+        print(expected)
+        print(actual)
```
benjaminysmith, line 133: Were these prints for testing the tests?

safegraph/delphi_safegraph/run.py
```diff
@@ -24,11 +24,14 @@ def run_module():
     aws_endpoint = params["aws_endpoint"]
     wip_signal = params["wip_signal"]
 
-    process_file = partial(process,
```
benjaminysmith, line 27: I noticed this used `partial` before and no longer -- does that make any difference? (I don't understand why you would use that over a regular function)

safegraph/delphi_safegraph/run.py
```diff
@@ -43,6 +46,7 @@ def run_module():
             'AWS_DEFAULT_REGION': aws_default_region,
         },
         shell=True,
+        check=True,
```
benjaminysmith, line 49: What does this do?

safegraph/delphi_safegraph/run.py
```diff
@@ -24,11 +24,14 @@ def run_module():
     aws_endpoint = params["aws_endpoint"]
     wip_signal = params["wip_signal"]
 
-    process_file = partial(process,
-                           signal_names=add_prefix(SIGNALS, wip_signal, prefix='wip_'),
-                           geo_resolutions=GEO_RESOLUTIONS,
-                           export_dir=export_dir,
-                           )
+    def process_file(current_filename):
```
benjaminysmith, line 27: It looks like this is changing the default behavior -- I don't know how Delphi handles changes like this but we would typically put it behind a flag -- is something like that worth doing, or are Delphi ok with just manually running and verifying?

safegraph/delphi_safegraph/run.py
sgsmob: I don't think `partial` is useful here anymore.  It was used to fix all the other arguments to `process()` since because `pool.map()` requires the function to have a single argument.  With the addition of the smoothing, we need two variable arguments to `process()`:  the current filename and the list of previous filenames (derived from the current filename).  It would be possible to refactor the call to `files_in_past_week` to live inside of `process()` but that strikes me as harder to test and not as easily extendable to #273.

safegraph/delphi_safegraph/run.py
sgsmob: It causes an exception to be raised if the subprocess exits with non-zero status (as opposed to failing silently).  Pylint requested this be added.

safegraph/delphi_safegraph/run.py
sgsmob: Unsure what the protocol is

safegraph/tests/test_process.py
sgsmob: yes, removed.

safegraph/delphi_safegraph/run.py
benjaminysmith: @krivard can you help advise?

safegraph/delphi_safegraph/run.py
benjaminysmith: Thanks for the explanation!

safegraph/delphi_safegraph/run.py
krivard: Protocol is: If it's going to change the behavior of the _output_, then it needs to be more broadly reviewed for statistical quality, and either mentioned in the public changelog (if it's a minor change) or under a new signal name (if it's a major change). If it's a change in how the code works, then if it gets called by another system, put it behind a flag. If it's just called internally, document it, but otherwise anything goes.

It looks like this particular change falls mostly under the latter. The new 7dav signals will get statistically reviewed before we release them, but just changing `process_file` to include a batch in addition to a single file is fine. Since we are promoting it to a full function instead of a partial/lambda, would you add a docstring that notes the constraints imposed by `process()`?

sgsmob: Done.

benjaminysmith (dismissed):

krivard:
What are the signal names for the 7-day average outputs? I'll need to add them to the wip list before we can merge.

sgsmob:
Ah, I have not actually written them to different files.  Let me get on that and get back to you

korlaxxalrok (approved):
Looks good to me 👍

sgsmob:
@krivard Smoothed signals are named `{signal}_7d_avg`.  They have been added the appropriate field in the `params.json` file

krivard (approved):

PR #310, opened 2020-10-13
Extend geocode utility to actually support the state to state mappings

dshemetov:
Add support for the state_x -> state_y where x,y are in {code, id, name} mappings


_delphi_utils_python/tests/test_geomap.py
```diff
@@ -382,6 +382,11 @@ def test_add_geocode(self):
         new_data2 = gmpr.add_geocode(new_data, "state_code", "hhs_region_number")
         assert new_data2["hhs_region_number"].unique().size == 2
 
+        # state_name -> state_id
+        new_data = gmpr.add_geocode(self.zip_data, "zip", "state_name")
+        new_data2 = gmpr.add_geocode(new_data, "state_name", "state_id")
+        assert new_data2.shape == (12, 6)
```
chinandrew, line 388: should there also be an assert statement for `new_data`?

dshemetov: Nah, new_data is just the test data for state_name to state_id, but it starts in zip form

chinandrew:
not specifically related to this PR, I just missed this in the previous one: `date_cols` is undocumented in the `replace_geocode()` docstring

dshemetov:
Ahhh, thanks, might as well take care of that here.

_delphi_utils_python/tests/test_geomap.py
```diff
@@ -278,11 +278,13 @@ def test_zip_to_state_id(self):
     def test_add_population_column(self):
         gmpr = GeoMapper()
         new_data = gmpr.add_population_column(self.fips_data_3, "fips")
-        assert new_data["population"].sum() == 274963
+        assert new_data.shape == (5, 5)
         new_data = gmpr.add_population_column(self.zip_data, "zip")
-        assert new_data["population"].sum() == 274902
```
chinandrew, line 283: any reason not to keep both of these checks?

Also, if it's 5x5 output it may be simpler juts to do a direct df comparison on values

_delphi_utils_python/tests/test_geomap.py
```diff
@@ -411,6 +420,33 @@ def test_add_geocode(self):
...
+        # hrr -> nation
+        with pytest.raises(ValueError):    
+            new_data = gmpr.replace_geocode(self.zip_data, "zip", "hrr")
+            new_data2 = gmpr.replace_geocode(new_data, "hrr", "nation")
+
+        # hrr -> nation
+        with pytest.raises(ValueError):    
+            new_data = gmpr.replace_geocode(self.zip_data, "zip", "hrr")
+            new_data2 = gmpr.replace_geocode(new_data, "hrr", "nation")
+
```
chinandrew, line 437: duplicated tests?

dshemetov: Oops, this is a new one 😄

_delphi_utils_python/tests/test_geomap.py
dshemetov: This is an old test I don't see anymore. Pull again?

chinandrew: I was just asking why it was deleted

dshemetov: Ah. I decided to move away from tests based on data-derived population counts. I figured the tests should catch whether the underlying logic or arithmetic breaks, not whether the data file changed. Am open to reasons for keeping though.

_delphi_utils_python/tests/test_geomap.py
```diff
@@ -411,6 +420,23 @@ def test_add_geocode(self):
             )
         )
 
+        # hrr -> nation
+        with pytest.raises(ValueError):    
+            new_data = gmpr.replace_geocode(self.zip_data, "zip", "hrr")
```
chinandrew, line 425: I'm pretty sure if the first line valueerrors but the second one doesnt, it still passes, so may want to split this into two `with pytest raises....` statements

dshemetov: Wouldn't that have been caught in the zip -> hrr test some lines before that though?

chinandrew: Maybe im understanding this test wrong. I read it as testing that both 

```
            new_data = gmpr.replace_geocode(self.zip_data, "zip", "hrr")
```
and
```
            new_data2 = gmpr.replace_geocode(new_data, "hrr", "nation")
```
raise valueerrors. Is that right?

dshemetov: It's testing the second one, since we should not be mapping hrr -> nation (it's an incomplete mapping, so it's unsupported).

chinandrew: Ohhh got it, didn't see the second line calls new_data. Thanks.

chinandrew (approved):
lgtm

cdc_covidnet/delphi_cdc_covidnet/update_sensor.py
```diff
@@ -85,9 +85,15 @@ def update_sensor(
...
-    hosp_df = geo_map.state_name_to_abbr(hosp_df)
-
+    gmpr = GeoMapper()
+    hosp_df = gmpr.add_geocode(hosp_df,
+                               from_col=APIConfig.STATE_COL,
+                               from_code="state_name",
+                               new_code="state_id",
+                               dropna=False)
+    # To use the original column name, reassign original column and drop new one
+    hosp_df[APIConfig.STATE_COL] = hosp_df["state_id"].str.upper()
```
dshemetov, line 95: @chinandrew btw, this may be cdc_covidnet specific, but the state abbreviation in other indicators (like JHU) is assumed to be lower case. Are we sure this is what we want?

chinandrew: For some reason cdc covidnet was uppercase, so i kept it consistent. if it's something we can change, would definitely recommend we standardize, but haven't looked into it.

dshemetov: @krivard thoughts?

chinandrew: Slacked Katie on this while discussing a similar issue, apparently downstream ingestion will standardize everything and accepts lower or uppercase, so this can be removed and we can just move to lowercase for the indicator code.

Issue #306, opened 2020-10-09
Refactor indicators to use delphi_utils library calls


krivard:
We now/will soon have the following pipeline utilities in delphi_utils:

* geographic aggregation: delphi_utils.geomap
* smoothing: pending #176
* validation: pending #155 

These behaviors are currently implemented individually in each indicator. This ticket will track refactors on all indicators to use the library versions instead.

indicator | geo | smoothing | validation |
-------- | ---- | ----------- | ---------- |
cdc_covidnet |[in progress](https://github.com/cmu-delphi/covidcast-indicators/pull/311) (@chinandrew) | | |
~cds~[1] | | | |
claims_hosp | [in progress](https://github.com/cmu-delphi/covidcast-indicators/pull/322) (@chinandrew)| | |
combo_cases_and_deaths |✅ (no change necessary) | | |
doctor_visits | in progress (@chinandrew) | | |
~emr_hosp~[2] | ✅  | | |
facebook[3] | ❌  | ❌  | |
~google_health~[4] | | | |
jhu | ✅  | [in progress](https://github.com/cmu-delphi/covidcast-indicators/pull/177) | |
nchs_mortality | ✅   | | |
quidel | ✅   | | |
quidel_covidtest | ✅   | | |
safegraph | [in progress](https://github.com/cmu-delphi/covidcast-indicators/pull/313) (@chinandrew)| | |
safegraph_patterns | ✅  | | |
usafacts |[in progress](https://github.com/cmu-delphi/covidcast-indicators/issues/316) (@sgsmob)  | | |

[1] - we don't plan to continue development on CDS; don't refactor it
[2] - the emr-hosp indicator is no longer active; low priority
[3] - the facebook indicator is in R; python utilities can't be a part of the main computation (smoothing, aggregation) but validation can be added as a separate step
[4] - deprecated in favor of a new indicator using the new symptoms search product; don't refactor it

krivard:
[Relevant for the validation refactors](https://github.com/cmu-delphi/covidcast-indicators/pull/155#discussion_r504218414)

chinandrew:
I can take the covidnet geo refactor. Updated table accordingly

chinandrew:
I believe for Google Health we need DMA methods with aren't available yet, so made an issue and updated the table. Going down the list, will take safegraph next (which is also a one liner since all it's doing is mapping state abbr and FIPS with a dict, which honestly isn't the worst -- just not consistent)

[...]

PR #311, opened 2020-10-14
Refactor cdc_covidnet to use geo utils

chinandrew:
For #306. Depends on #310 (which is currently the base in this PR for easier review).

Summary of changes:
- Convert cdc_covidnet indicator to use the GeoMapper utility. 
- Remove old cdc_covidnet specific geo_maps file and code


chinandrew:
Unrelated to this refactor but I noticed the `static_path` argument doesn't appear to be used in `update_sensor()`, is there any reason it's still included? @eujing

eujing:
@chinandrew `static_path` was only used with the original implementation for `GeoMaps(static_path)`, but since that has been refactored I dont think its needed anywhere else.
I agree I think it should be removed!

chinandrew:
> @chinandrew `static_path` was only used with the original implementation for `GeoMaps(static_path)`, but since that has been refactored I dont think its needed anywhere else.
> I agree I think it should be removed!
Cool thanks for verifying, will remove.

cdc_covidnet/delphi_cdc_covidnet/update_sensor.py
```diff
@@ -85,9 +85,15 @@ def update_sensor(
...
-    geo_map = GeoMaps(static_path)
-    hosp_df = geo_map.state_name_to_abbr(hosp_df)
-
+    gmpr = GeoMapper()
+    hosp_df = gmpr.add_geocode(hosp_df,
+                               from_col=APIConfig.STATE_COL,
+                               from_code="state_name",
+                               new_code="state_id",
+                               dropna=False)
+    # To preserve column order, reassign original column and drop new one
```
sgsmob, line 94: Is there a test that fails if this line is omitted?

sgsmob (commented):
Looks good to me pending an answer on my question above.

cdc_covidnet/delphi_cdc_covidnet/update_sensor.py
chinandrew: So if `hosp_df[APIConfig.STATE_COL] = hosp_df["state_id"].str.upper()` is omitted, tests fail because APIConfig.STATE_COL isn't the two letter abbreviation.

if `hosp_df.drop("state_id", axis=1, inplace=True)` is omitted, tests fail because dimensions are wrong.

I also actually just realized column order for these fields doesn't matter since there's a `hosp_df.set_index(["date", "geo_id"], inplace=True)` right after, so this is actually just to avoid name collisions since there isn't currently functionality to replace a column while keeping the name the same. I've updated the comment to reflect that.

sgsmob (approved):

PR #314, opened 2020-10-15
fix broken usafacts tests to read from the proper directories

sgsmob:
`usafacts` was trying to read/write to/from `receiving` when run in the `tests` directory rather than `../receiving`.  The tests now pass.


jingjtang:
Hi @sgsmob, I think a better way to solve this is to add a `./receiving` folder under tests, then we will stored the output from tests under `usafacts/tests/receiving` and check those file there. 

The current solution seems weird. Since the receiving dir in `usafacts/tests/params.json` is stilled defined as `usafacts/tests/receiving`, but in `test_run.py` we check the output from `usafacts/receiving`.

sgsmob:
Changed to add the `./receving` directory.

jingjtang (dismissed):
All unit tests passed

krivard (dismissed):

krivard:
Oh ffs. @korlaxxalrok this PR seems to trigger the same pytest/asteroid "recursion limit exceeded" problem as we were having in JHU. Tests seem to pass in Jingjing's environment so I think we're okay to disable them here the same way. If you agree, go ahead and do that so we can unblock Mike.

Afterwards, could you work with @benjaminysmith to figure out a more permanent solution?

sgsmob:
Yes I observed the recursion limit issue.  FWIW I have a pylint extension on my IDE and I corrected the mistakes there.

korlaxxalrok:
I'll disable linting for now, and then spend some time looking at an actual fix.

korlaxxalrok:
@krivard @sgsmob Linting is disabled.

krivard (approved):

PR #322, opened 2020-10-16
Refactor claims_hosp to use new geomapper functions

chinandrew:
Part of #306 

Summary of changes:
- Use the latest geomapper functions
- remove unused zips static file

Couple things to note here:

The previous functions, e.g. `county_to_state()`, don't exist in the GeoMapper package. 
- They do exist in [emr_hosp](https://github.com/cmu-delphi/covidcast-indicators/blob/9f0c22865441da2d12e5a8357918de7569074c63/emr_hosp/delphi_emr_hosp/geo_maps.py), but with different arguments.
- There are deprecated functions like `convert_fips_to_state_id()` in the GeoMapper package which contain the arguments called by this code, so maybe they used to be named `county_to_state()` but were renamed later. _However_, if you use these functions, the code doesn't work due to a non-unique multi-index.

After switching to the new functions all tests pass, but I can't verify against previous behaviour since the original code didn't run.

### DIFF
--- a/claims_hosp/delphi_claims_hosp/update_indicator.py
+++ b/claims_hosp/delphi_claims_hosp/update_indicator.py
@@ -127,7 +127,6 @@ class ClaimsHospIndicatorUpdater:
         assert (
                 len(multiindex) <= (GeoConstants.MAX_GEO[self.geo] * len(self.fit_dates))
         ), "more loc-date pairs than maximum number of geographies x number of dates"
-        print(data_frame)
         # fill dataframe with missing dates using 0
         data_frame = data_frame.reindex(multiindex, fill_value=0)
         data_frame.fillna(0, inplace=True)
--- a/claims_hosp/delphi_claims_hosp/update_indicator.py
+++ b/claims_hosp/delphi_claims_hosp/update_indicator.py
@@ -96,13 +96,21 @@ class ClaimsHospIndicatorUpdater:
         """
         geo_map = GeoMapper()
         if self.geo == "county":
-            data_frame = geo_map.county_to_megacounty(
-                data, Config.MIN_DEN, Config.MAX_BACKWARDS_PAD_LENGTH,
-                thr_col="den", mega_col=self.geo)
+            data_frame = geo_map.fips_to_megacounty(data,
+                                                    Config.MIN_DEN,
+                                                    Config.MAX_BACKWARDS_PAD_LENGTH,
+                                                    thr_col="den",
+                                                    mega_col=self.geo)
         elif self.geo == "state":
-            data_frame = geo_map.county_to_state(data, state_id_col=self.geo)
+            data_frame = geo_map.replace_geocode(data,
+                                                 from_code="fips",
+                                                 new_col=self.geo,
+                                                 new_code="state_id")
+            data_frame[self.geo] = data_frame[self.geo].str.upper()
         elif self.geo == "msa":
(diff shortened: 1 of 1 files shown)


claims_hosp/delphi_claims_hosp/update_indicator.py
```diff
@@ -96,13 +96,21 @@ def geo_reindex(self, data):
...
         elif self.geo == "state":
-            data_frame = geo_map.county_to_state(data, state_id_col=self.geo)
+            data_frame = geo_map.replace_geocode(data,
+                                                 from_code="fips",
+                                                 new_col=self.geo,
+                                                 new_code="state_id")
+            data_frame[self.geo] = data_frame[self.geo]
         elif self.geo == "msa":
-            data_frame = geo_map.county_to_msa(data, msa_col=self.geo)
+            data_frame = geo_map.replace_geocode(data,
```
mariajahja, line 111: I'm seeing this error when running. I don't know how the new geomap functions work, hopefully it's just a simple `new_code='X'` addition?

```
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/mjahja/Delphi/covidcast-indicators/claims_hosp/delphi_claims_hosp/__main__.py", line 11, in <module>
    run_module()  # pragma: no cover
  File "/Users/mjahja/Delphi/covidcast-indicators/claims_hosp/delphi_claims_hosp/run.py", line 89, in run_module
    updater.update_indicator(params["input_file"], params["export_dir"])
  File "/Users/mjahja/Delphi/covidcast-indicators/claims_hosp/delphi_claims_hosp/update_indicator.py", line 152, in update_indicator
    data_frame = self.geo_reindex(data)
  File "/Users/mjahja/Delphi/covidcast-indicators/claims_hosp/delphi_claims_hosp/update_indicator.py", line 114, in geo_reindex
    new_col=self.geo)
TypeError: replace_geocode() missing 1 required positional argument: 'new_code'
```

mariajahja (changes requested):
The original code is broken in `main` due to the changes introduced in GeoMapper. I can run a comparison off commit 64eb2276452244fc860cba0e8053e67b1b57d307 (when the package was originally introduced, and before GeoMapper changes). The results line up for `state`, `hrr`, and `county`, but `msa` has an error.

claims_hosp/delphi_claims_hosp/update_indicator.py
chinandrew: Oops meant to be `new_code` instead of `new_col`, should be fixed now.

mariajahja: Works great!

mariajahja (approved):

krivard (approved):

krivard:
I don't believe this indicator makes any use of the GeoMapper population functions. Is that correct? If we use population data here, I should block this merge until we fix #325.

mariajahja:
Yes, this indicator does not use population. So there shouldn't be any problem merging.

dshemetov (commented):
Late but lgtm

PR #327, opened 2020-10-17
Add a gap detector to Sir Complains-a-lot

capnrefsmmat:
Now complains if there is a gap in the reported data, e.g. if 2020-10-10 is missing but 2020-10-11 is present. This seems like it'd be very useful to have.

Also, add missing sources to the template.


krivard (approved):
xlnt

capnrefsmmat:
One note: it now takes like 9 minutes to run, because of all the API calls. But that will be much better when we figure out API streaming.

Issue #318, opened 2020-10-15
GeoMapper produces invalid counties >80000 for JHU


krivard:
Specifically, `88888` and `99999`. COVIDcast fails any county file with a geo over 80000, so that's not allowed.

This might be related to the 800XX and 900XX codes referenced [here](https://github.com/cmu-delphi/covidcast-indicators/issues/179#issuecomment-689137773) and [here](https://github.com/cmu-delphi/covidcast-indicators/issues/254#issuecomment-683290965); JHU plays pretty fast and loose with their geo coding. It's totally fine for GeoMapper to use 8xxxx and 9xxxx codes internally for geo aggregation, we just can't let them escape into CSV files destined for COVIDcast ingestion.

@dshemetov would you introduce @sgsmob  to the relevant code and documentation here and on JHU's github?

dshemetov:
Ah these are likely coming from the Puerto Rico 72888 and 72999 UIDs, will take a look.

dshemetov:
I've been meaning to make a unified map of all the hand modifications we use in JHU, it might be time.

sgsmob:
So do we need to enforce this in GeoMapper or do we need to have some/all indicators check this later?

dshemetov:
@sgsmob Since this a global requirement for all indicators, I think we should enforce it once in GeoMapper.

dshemetov:
Ah I see, these are not Puerto Rico, but are the Princess Cruises ships. @sgsmob this is the [JHU UID lookup file](https://raw.githubusercontent.com/CSSEGISandData/COVID-19/master/csse_covid_19_data/UID_ISO_FIPS_LookUp_Table.csv) from [this repo](https://github.com/CSSEGISandData/COVID-19/tree/master/csse_covid_19_data).
```
84088888,US,USA,840,88888,,Diamond Princess,US,,,"Diamond Princess, US",
84099999,US,USA,840,99999,,Grand Princess,US,,,"Grand Princess, US",
```
We just need to make sure they are dropped from the JHU UID -> FIPS mapping file. Should be a few lines.

dshemetov:
Also the high-level GeoMapper docs are [here](https://github.com/cmu-delphi/covidcast-indicators/tree/main/_delphi_utils_python/data_proc/geomap). The section on geo codes provides general info on the codes we support and the section on source files describes the authorities we get the lookup tables from.

dshemetov:
[...]

PR #328, opened 2020-10-18
Fix JHU prop signal 

### LINKED ISSUE #325: GeoMapper doubles the effective population in state-level prop signals
More information [in this Slack thread](https://delphi-org.slack.com/archives/C0130CSQRN3/p1602953129060800?thread_ts=1602938403.054000&cid=C0130CSQRN3), but the TL;DR is that JHU state-level `prop` signals are half the magnitude they should be, according to USAFacts and NYT. Since GeoMapper includes population for both state and FIPS, it may be summing both of them when computing state `prop` instead of just using one or the other. 

Might be related to Unassigned handling.

dshemetov:
Current solution zeros out the state FIPS population in the new population source file. Don't quite understand why that occurred, yet. Fixes #325.


capnrefsmmat:
Is there an opportunity for unit tests to prevent this from regressing again?

dshemetov:
That's really worth reflecting on. What sort of tests could have caught this sort of issue? First thought are the indicator validation and anomaly detection projects, which are aimed at related problems. Second thought is the qualitative indicator validation notebook I built [here](https://github.com/cmu-delphi/covidcast-indicators/blob/deploy-jhu/testing_utils/indicator_validation.template.ipynb). 

One challenge in designing tests here is the lack of ground truth - since all reference data may be faulty, we will need to accept sensors with non-zero false positive rates. These sensors will ideally draw our attention to possible flaws for further analysis, but will not block merging. Ideally it will be easy to go from a sensor alert to an interactive REPL/notebook session exploration the suspect data. Another thought is to add correlation tests with a trusted set of indicators.

I would like to learn more about the established best practices for this area. @nmdefries @jsharpna do you have thoughts on this? I'd love to compare notes on your related projects.

capnrefsmmat:
In a test case, you'd provide known input -- for example, you'd provide input that says there are 1,000,000 cases, and compare the answer to what you'd get. The ideal way to facilitate this would be if the proportion calculation were factored out into its own function that takes counts and populations as arguments; then you could provide it some known counts and populations and ensure it returns the right proportions.

dshemetov:
Ah looks like we did have a prop test with synthetic population and values, but it didn't have any population for the megaFIPS.

_delphi_utils_python/data_proc/geomap/geo_data_proc.py
```diff
@@ -391,6 +391,10 @@ def create_fips_population_table():
     df_pr = df_pr.groupby("fips").sum().reset_index()
     df_pr = df_pr[~df_pr["fips"].isin(census_pop["fips"])]
     census_pop_pr = pd.concat([census_pop, df_pr])
+
+    # Zero out the populations for the state FIPS codes XX000 to avoid double counting
+    megafips_codes = [str(x).zfill(2) + "000" for x in range(1, 73)]
```
krivard, line 396: Won't this result in inf results for megafips prop signals?

dshemetov: I don't think so. The denominator in JHU is obtained my merging a FIPS population dataframe on the FIPS-level data, which is then aggregated to the state level. I verified this in [this updated notebook](https://github.com/cmu-delphi/covidcast-indicators/blob/296d75da2488d70672a7db30418ccad313c1c045/testing_utils/indicator_validation.template.ipynb), where I compare the new JHU state prop values with USA facts. Rhode Island is the largest outlier because we are now using values from Unassigned that we weren't before.

Let me double check how other indicators like Safegraph do this calculation, though.

krivard: I meant county files -- don't we include unassigned as a megafips for county?

dshemetov:
I have identified why the state FIPS codes started being added into the population totals now. It is because the GeoMapper updated the JHU UID -> FIPS mapping to keep the state FIPS (for the purposes of aggregating Unassigned and Out of State data). Previously, most of them were filtered out, prior to joining the FIPS -> population file.

So while the state FIPS codes are important to keep in JHU UID -> FIPS mapping file and in the FIPS -> state mapping file (so that the Unassigned and Out of State values don't get filtered out), we need to set the state FIPS population to zero to avoid double counting when aggregating up to state.

dshemetov:
Right, so Safegraph performs the aggregation to [state similarly](https://github.com/cmu-delphi/covidcast-indicators/blob/0783ebd6d3ca6cdb07a512dea08f4ca08a0e901f/safegraph/delphi_safegraph/process.py#L138). Namely, it truncates the county FIPS code to the first two digits and then aggregates.

It also appears that Safegraph provides [Census block group](https://en.wikipedia.org/wiki/Census_block_group) mappings, which should probably move to the geocoding util. That's a separate issue though.

_delphi_utils_python/data_proc/geomap/geo_data_proc.py
krivard: > don't we include unassigned as a megafips for county?

Turns out we don't! So this won't be a problem.

krivard: We do, in fact:
```
$ zgrep -m1 "^[0-9][0-9]000" *_county_* | head
20200220_county_confirmed_7dav_cumulative_num.csv.gz:60000,0.0,NA,NA
20200220_county_confirmed_7dav_incidence_num.csv.gz:60000,0.0,NA,NA
20200220_county_confirmed_7dav_incidence_prop.csv.gz:01000,0.0,NA,NA
20200220_county_confirmed_7day_avg_cumulative_num.csv.gz:72000,0.0,NA,NA
20200220_county_confirmed_7day_avg_cumulative_prop.csv.gz:72000,0.0,NA,NA
20200220_county_confirmed_7day_avg_incidence_num.csv.gz:72000,0.0,NA,NA
20200220_county_confirmed_7day_avg_incidence_prop.csv.gz:72000,0.0,NA,NA
20200220_county_confirmed_cumulative_num.csv.gz:60000,0.0,NA,NA
20200220_county_confirmed_incidence_num.csv.gz:60000,0.0,NA,NA
20200220_county_confirmed_incidence_prop.csv.gz:01000,0.0,NA,NA
```

krivard:
I've confirmed that the state prop signals are now on par with usa-facts, but we are now getting inf values for megafips in the county files:

```
$ grep "^[0-9][0-9]000" *county*prop* | head
20200310_county_confirmed_cumulative_prop.csv:49000,inf,NA,NA
20200310_county_confirmed_cumulative_prop.csv:53000,inf,NA,NA
20200310_county_confirmed_incidence_prop.csv:49000,inf,NA,NA
20200310_county_confirmed_incidence_prop.csv:53000,inf,NA,NA
20200311_county_confirmed_cumulative_prop.csv:49000,inf,NA,NA
20200311_county_confirmed_incidence_prop.csv:53000,-inf,NA,NA
20200312_county_confirmed_cumulative_prop.csv:49000,inf,NA,NA
20200312_county_confirmed_cumulative_prop.csv:53000,inf,NA,NA
20200312_county_confirmed_incidence_prop.csv:53000,inf,NA,NA
20200313_county_confirmed_cumulative_prop.csv:01000,inf,NA,NA
```

capnrefsmmat:
Definitely a sign that we need a few unit tests for the different combinations: state, county, megacounty, etc.

dshemetov:
@krivard I see what you mean. I wonder if we should filter those state FIPS out from the county .csvs, since for JHU specifically they currently represent a combination of Unassigned and Out of State signals only, which is not something a user would expect.

krivard:
@dshemetov We don't have a clear picture of what users truly expect from a megafips cases signal. We started publishing Unassigned cases under the megafips some months ago because we wanted there to be a way for API users to retrieve those values. Merging in the Out Of State might be counterintuitive in some sense, but it does maintain access through the API to counts of less-certain location, and it does maintain the presence of megafips in the county signals.

Two options:
* Keep megafips population as 0, and drop megafips from JHU county reports. Anyone who wants Unassigned/OOS can get it by subtracting the sum of all actual-county cases from the state case count. If they need prop, they can reconstruct the state population by doing state num/prop. Manually delete all issues of megafips from JHU in the API.
* Restore megafips population, but exclude it when aggregating to state. This requires more complex handling in GeoMapper, but will more closely match existing JHU behavior.

I have a slight preference for the latter, but would yield to argument.

dshemetov:
@krivard I see your point about not knowing what users expect there. I have not been able to get a clear picture from the JHU documentation with regard to what actually goes in the Out of State bucket; Unassigned appears to contain a combination of counts that could not be localized to a particular county FIPS and probable deaths released by state estimates. I can see the value in releasing the Unassigned category to users, but am less sure about the Out of State category. I'm also open to separating Unassigned and Out of State into separate megaFIPS from XX000, if we can find a reasonable code for them (e.g. XX888 and XX999 appear to be unclaimed for all FIPS codes; Puerto Rico out of state and unassigned cases are in 72888 and 72999, respectively).

If we do not remove the megafips codes XX000, the GeoMapper changes wouldn't be too bad:
```
elif geo_res == "state":
  df = df.set_index("fips")
  # Zero out the state FIPS population to avoid double counting.
  state_fips_codes = {str(x).zfill(2) + "000" for x in range(1,73)}
  subset_state_fips_codes = set(df.index.values) & state_fips_codes
  df.loc[subset_state_fips_codes, "population"] = 0
  df = df.reset_index()
  df = gmpr.replace_geocode(df, "fips", "state_id", new_col="geo_id", date_col="timestamp")
```

dshemetov:
That said, what do you think, @krivard? I could either keep the XX000 codes like above or do the splitting into XX888/XX999 now (should be a small change in the mapping files, but would require updating the API mailing list and the docs).

dshemetov:
To fill the XX888/XX999 idea out some more: we would drop the XX000 county codes from the JHU -> FIPS map. The population file would stay the same and have state FIPS population as well. This would be fine for JHU, since it doesn't normally store any data in XX000.

Other indicators: 
* USAFacts will be unaffected regardless of what we do here. USAFacts supports XX000 codes and prop signals, but the [XX000 codes are excluded from the prop signals](https://github.com/cmu-delphi/covidcast-indicators/blob/8ce66a28dd428729db5de9fe0ad136d65a8541eb/usafacts/delphi_usafacts/geo.py#L98), despite providing [XX000 populations in the static file](https://github.com/cmu-delphi/covidcast-indicators/blob/8ce66a28dd428729db5de9fe0ad136d65a8541eb/usafacts/static/fips_population.csv). 
* Safegraph will be unaffected too. It has different prop signals that are not dependent on population. They appear to be [proportional per device count](https://github.com/cmu-delphi/covidcast-indicators/blob/0783ebd6d3ca6cdb07a512dea08f4ca08a0e901f/safegraph/delphi_safegraph/process.py#L97) and [standard errors depend on sample size](https://github.com/cmu-delphi/covidcast-indicators/blob/0783ebd6d3ca6cdb07a512dea08f4ca08a0e901f/safegraph/delphi_safegraph/process.py#L159). 
* Safegraph patterns will be unaffected too. The data files start coded by ZIP before being converted to FIPS. ZIP doesn't contain any mega values and the translation ZIP to FIPS does not translate to any state FIPS values.

krivard:
@dajmcdon What is most useful for forecasting, in the handling of Unassigned and Out-of-state cases/deaths in JHU county level signals? We currently put Unassigned into the megafips/megacounty, but it's not clear (see above) whether Out-of-state should go in there too or be moved to a separate unused FIPS for that state.

dshemetov:
[This table](https://github.com/cmu-delphi/covidcast-indicators/issues/318#issuecomment-713052143) may be helpful context for the whole JHU -> FIPS custom mapping.

dajmcdon:
Our goal is to use JHU as the "truth". The short answer, I think, if @dshemetov has the bandwidth would be to check the script [here](https://github.com/reichlab/covid19-forecast-hub/blob/master/data-truth/get-truth-data.py) to try to figure out what Reich Lab is doing with them. I'm also curious how the 000 codes relate to the roll-up from county level (do they include the megacounty or out-of-state?).

dshemetov:
The Reich lab looks to be pulling their geocodes from [this file](https://github.com/reichlab/covid19-forecast-hub/blob/master/data-locations/locations.csv), which contains 5-digit county FIPS codes and 2-digit state FIPS codes. The 2-digit state codes are transformed into the XX -> XX000 pattern [here](https://github.com/reichlab/covid19-forecast-hub/blob/6bd90b9ce1bd026a0504ef6b57e9d304e16b740d/data-truth/get-truth-data.py#L77). What is mapped to those 2-digit state codes? Well, they cleverly use the 'Province_State` column [here](https://github.com/reichlab/covid19-forecast-hub/blob/6bd90b9ce1bd026a0504ef6b57e9d304e16b740d/data-truth/get-truth-data.py#L162) to aggregate up to the state level, which lets them avoid a good chunk of our custom tables. 

The result is that it pools counties, Out of State, Unassigned, and another state category all into XX000. They do not appear to provide a separate category for Out of State and Unassigned.

As a personal aside, I like `get_truth` as a [function name](https://github.com/reichlab/covid19-forecast-hub/blob/6bd90b9ce1bd026a0504ef6b57e9d304e16b740d/data-truth/get-truth-data.py#L157). Has a very powerful energy.

dajmcdon:
It does have a nice energy/optimism. 

So if we want to predict the truth. Do we target the xx000 signal, or do we need to try and grab the "Province_State" sums? Or do these magically equal each other?

dshemetov:
Right, so it's the latter. Just to be clear, XX000 is a code pattern of our own devising that currently holds the Out of State and Unassigned categories from JHU. The Unassigned bin typically contains state gov's estimates of probable cases/deaths, though sometimes it contains confirmed values that just couldn't be localized to a county, like in Rhode Island. Out of State is another bin like that, but less well documented. Do you foresee having any need for these categories separate from their aggregation into states?

dajmcdon:
Ok, so I'm still not sure what the right answer is then. We want to predict the "truth". If the truth is XX000, then great. But if XX000 \= JHU's province_state, and we don't have JHU's province_state signal, then we need a fix. I don't believe that we actually need the bins separately. (@bnaras or @capolitsch can be more definitive)

Short term (like before Friday), we need something reasonable (whatever was there before is fine). But long term, wrapped up in some of the anomaly detection or data checking, we should make sure that our truth and their truth is the same (unless Delphi doesn't believe their truth)

krivard:
Excellent! Thank you both. 

For now, let's keep the XX000 codes and their previous definition of Unassigned + Out of State. Add the logic to drop XX000 population when aggregating up to state.

I've created [an issue](https://github.com/cmu-delphi/covidcast-indicators/issues/350) to track a future effort to figure out whether we're doing the right thing long-term.

jhu/delphi_jhu/pull.py
```diff
@@ -28,87 +76,39 @@ def pull_jhu_data(base_url: str, metric: str, pop_df: pd.DataFrame) -> pd.DataFr
     may be negative.  This is wholly dependent on the quality of the raw
     dataset.
 
-    We filter the data such that we only keep rows with valid FIPS, or "FIPS"
+    We filter the data such that we only keep rows with valid FIPS or "FIPS"
     codes defined under the exceptions of the README.  The current  exceptions
     include:
```
krivard, line 81:
```suggestion
    codes defined under the exceptions of the README.
```

krivard (approved):
japprove!

krivard (approved):

PR #336, opened 2020-10-20
Revert "Add 7-day average of each signal to Safegraph indicator"

krivard:
Reverts cmu-delphi/covidcast-indicators#309 to see if that fixes [this error](https://delphi-org.slack.com/files/UT124VC8P/F01CZ9AJXB4/untitled?origin_team=TF95CN37E&origin_channel=C013AH5N01E)


sgsmob:
Why are there so many files changed/deleted?

krivard:
Because the `deploy-*` branch structure means that every time we merge a new feature into a deployment branch, it brings with it all the changes from `main` that were done to every other indicator since the last time `deploy-safegraph` was updated. It's a known problem with this workflow, currently a P1 in the fall OKR to ameliorate.

sgsmob (approved):
This is hard to reason about but looks ok

PR #343, opened 2020-10-20
Make combo cases/deaths robust to source outages

### LINKED ISSUE #330: Make combo cases & deaths robust to outages
Currently, the combo cases and deaths signal quits if it can't get data from both jhu-csse and usa-facts. Instead, it should output whatever data it _can_ find, and just log a warning.

* [Relevant code](https://github.com/cmu-delphi/covidcast-indicators/blob/0783ebd6d3ca6cdb07a512dea08f4ca08a0e901f/combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py#L22)

krivard:
- okay to return just one of the data frames if both aren't available
- do something sensible if neither one is available
- unit tests to verify all possible cases are handled appropriately

Fixes #330


combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
```diff
@@ -9,46 +9,68 @@
...
+                  "sample_size": "sample_size"}
+def combine_usafacts_and_jhu(signal, geo, date_range, fetcher=covidcast.signal):
     """
     Add rows for PR from JHU signals to USA-FACTS signals
     """
-    usafacts_df = covidcast.signal("usa-facts", signal, date_range[0], date_range[1], geo)
-    jhu_df = covidcast.signal("jhu-csse", signal, date_range[0], date_range[1], geo)
-    check_not_none(usafacts_df, "USA-FACTS", date_range)
-    check_not_none(jhu_df, "JHU", date_range)
+    print("Fetching usa-facts...")
```
sgsmob, line 48: Are these necessary?

sgsmob, line 48: Are these necessary?

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
```diff
@@ -9,46 +9,68 @@
...
+    usafacts_df = fetcher("usa-facts", signal, date_range[0], date_range[1], geo)
+    print("Fetching jhu-csse...")
+    jhu_df = fetcher("jhu-csse", signal, date_range[0], date_range[1], geo)
+
+    if (not check_not_none(usafacts_df, "USA-FACTS", date_range)) and \
```
sgsmob, line 53: I would change the polarity here (return True if it is None) since the double negative is hard to parse.

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
```diff
@@ -9,46 +9,68 @@
...
-from delphi_utils import read_params, create_export_csv
-from .constants import *
-from .handle_wip_signal import *
+from delphi_utils import read_params
+from .constants import METRICS, SMOOTH_TYPES, SENSORS, GEO_RESOLUTIONS
+from .handle_wip_signal import add_prefix
 
 
 def check_not_none(data_frame, label, date_range):
     """Exit gracefully if a data frame we attempted to retrieve is empty"""
```
sgsmob, line 22: update comment to change new behavior

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
```diff
@@ -83,15 +105,12 @@ def sensor_signal(metric, sensor, smoother):
...
-    return sensor_name, signal
-
-def run_module():
-    """Produce a combined cases and deaths signal using data from JHU and USA Facts"""
-    variants = [tuple((metric, geo_res)+sensor_signal(metric, sensor, smoother))
-                for (metric, geo_res, sensor, smoother) in
-                product(METRICS, GEO_RESOLUTIONS, SENSORS, SMOOTH_TYPES)]
+    return sensor_name, "_".join([metric, sensor_name])
 
+def configure(variants):
```
sgsmob, line 110: Is the `variants` argument used here at all?

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
```diff
@@ -112,30 +131,38 @@ def run_module():
...
+    """Produce a combined cases and deaths signal using data from JHU and USA Facts"""
+    variants = [tuple((metric, geo_res)+sensor_signal(metric, sensor, smoother))
+                for (metric, geo_res, sensor, smoother) in
+                product(METRICS, GEO_RESOLUTIONS, SENSORS, SMOOTH_TYPES)]
+    params = configure(variants)
+    for metric, geo_res, sensor_name, signal in variants:
+        df = combine_usafacts_and_jhu(signal, geo_res, extend_raw_date_range(params, sensor_name)) # pylint: disable=invalid-name
 
-        df = df.copy()
+        df = df.copy() # pylint: disable=invalid-name
```
sgsmob, line 165: We should find a way to globally do this (not in the scope of this PR).

sgsmob, line 165: Why is this copy necessary?

combo_cases_and_deaths/tests/test_run.py
```diff
@@ -49,6 +50,33 @@ def test_handle_wip_signal():
     assert signal_names[0].startswith("wip_")
     assert all(not s.startswith("wip_") for s in signal_names[1:])
 
+def test_unstable_sources():
+    placeholder = lambda geo: pd.DataFrame(
```
sgsmob, line 54: Can you please add comments for what these functions are supposed to be doing?

sgsmob (changes requested):

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
krivard: covidcast.signal prints warnings about individual days not being available, but it doesn't say what signal it was trying to fetch. Without the print statement, it's impossible to tell from the logs which signal was failing on which days.

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
krivard: [Yes](https://github.com/cmu-delphi/covidcast-indicators/blob/9e46bc28248295a1b5c2db9a3cb9bd54291f63fe/combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py#L125)

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
krivard: No idea -- will remove.

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
capnrefsmmat: https://github.com/cmu-delphi/covidcast/issues/139

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
chinandrew: can achieve this through a pylint rc

krivard: @chinandrew yeah, though we typically only want to accept one or two otherwise invalid variable names and still check the rest

chinandrew: Never used this functionality before, but pylint lets you [enter custom regex to filter invalid names](http://pylint.pycqa.org/en/latest/user_guide/options.html#custom-regular-expressions), so might be able to add certain ones as exceptions through that.

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
sgsmob: Ah, I missed a code fold.

sgsmob (approved):

PR #344, opened 2020-10-21
Pipeline for the new google symptoms indicator

jingjtang:
- Only state and county level data provided. No aggregation to hrr/msa.
  - state level data derived based on the dataset [here](https://github.com/google-research/open-covid-19-data/blob/master/data/exports/search_trends_symptoms_dataset/United%20States%20of%20America/2020_US_daily_symptoms_dataset.csv) since there are 51 states & regions available including `District of Columbia`. There are only 78 dates available for `District of Columbia` in terms of `Anosmia` related searches. No data provided for `District of Columbia` in terms of `Ageusia` related searches.
  - county level data derived from [datasets for subregions](https://github.com/google-research/open-covid-19-data/tree/master/data/exports/search_trends_symptoms_dataset/United%20States%20of%20America/subregions) which provides daily data at county level with FIPS codes. 
  - The state level data in both of the two sources(as stated above) are the same. The only concern is that there is no county level data for `District of Columbia`.
  - All counties used by GS are included in our geo utils.
- Signals for `Anosmia` and `Ageusia` related searches respectively.
- `smoothed_search` is set to be the 7-day moving average. (Might change later)
- ~4-day lag
- NANs are ignored in `raw_search` signals. 
- In `smoothed_search`, NANs are filled with zero.


google_symptoms/delphi_google_symptoms/__main__.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+"""Call the function run_module when executed.
+
+This file indicates that calling the module (`python -m MODULE_NAME`) will
```
nmdefries, line 4: module_name

google_symptoms/DETAILS.md
```diff
@@ -0,0 +1,26 @@
+# Google Symptoms
+
+We import the confirmed case and deaths data from the Google Research's
+Open COVID-19 Data project and export the county-level and state-level data
+as-is.
```
nmdefries, line 5:
```suggestion
We import the symptom search term popularity data from the Google Research's
Open COVID-19 Data project and export the county-level and state-level data
as-is.  
```

google_symptoms/DETAILS.md
```diff
@@ -0,0 +1,26 @@
...
+Recoveries are _not_ reported.
+
+## Metrics, Level 2 (`m2`)
+* `raw_search`:  Google search volume reported as-is
+* `smoothed_search`:  Google search volume using 7-day moving average
+
+This data reflects the volume of Google searches mapped to symptoms such Anosmia
+and Ageusia. The resulting daily dataset for each region showing the relative frequency
+of searches for each symptom.  This signal is measured in arbitrary units that are normalized
+for population. Larger numbers represent higher numbers of symptom-related searches.
```
nmdefries, line 26:
```suggestion
of searches for each symptom.  This signal is measured in arbitrary units that are normalized
for population and for the most popular symptom search term within a geographic region. Thus, 
values are not comparable between geographic regions. Larger numbers represent higher 
numbers of symptom-related searches.
```

google_symptoms/DETAILS.md
```diff
@@ -0,0 +1,26 @@
+# Google Symptoms
+
+We import the confirmed case and deaths data from the Google Research's
+Open COVID-19 Data project and export the county-level and state-level data
```
nmdefries, line 4: How important is it to have this signal available at other geo types (msa, hrr) @krivard ?

Values aren't directly comparable across geo regions due to the normalization process. Search volume is normalized by population of the region and by the highest symptom search volume in a region at some set point in time. So, effectively, we have symptom search terms ranked by popularity. It's not obvious how to aggregate this from counties up to other geo types.

Maybe to aggregate, we could average search term popularity from member regions, weighting by population, which would produce an average per-capita search term rank.

google_symptoms/DETAILS.md
```diff
@@ -0,0 +1,26 @@
...
+* `county`: reported using zero-padded FIPS codes.  The county level data is derived 
+from `/subregions/state/2020_US_state_daily_symptoms_dataset.csv`.
+* `state`: reported using two-letter postal code. The state level data is derived from
+`2020_US_daily_symptoms_dataset.csv` which includes data for District of Columbia.
+
+## Metrics, Level 1 (`m1`)
+* `Anosmia`: Google search volume for Anosmia-related searches
+* `Ageusia`: Google search volume for Ageusia-related searches
+
+Recoveries are _not_ reported.
```
nmdefries, line 17: ? This looks leftover from something else.

google_symptoms/README.md
```diff
@@ -0,0 +1,61 @@
+# Google Symptoms
+
+We import the confirmed case and deaths data from the Google Research's
```
nmdefries, line 3:
```suggestion
We import the normalized symptom search term popularity data from the Google Research's
```

google_symptoms/DETAILS.md
capnrefsmmat: Our map also shows MSA, so if we put the symptoms signal on the map, it would be missing from that level. It's not the end of the world, though, so if aggregating takes more work, we could let it wait.

(And maybe we have a contact from google who can tell us the best aggregation strategy, if we choose to do so?)

google_symptoms/delphi_google_symptoms/run.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
+"""Functions to call when running the function.
+
+This module should contain a function called `run_module`, that is executed
+when the module is run with `python -m MODULE_NAME`.
```
nmdefries, line 5: Please fill in module_name.

google_symptoms/delphi_google_symptoms/pull.py
```diff
@@ -0,0 +1,113 @@
...
+    geo_list = df["geo_id"].unique()
+    date_list = pd.date_range(start=df["date"].min(),
+                              end=df["date"].max(),
+                              freq='D')
+    index_df = pd.MultiIndex.from_product(
+        [geo_list, date_list], names=['geo_id', 'date']
+    )
+    df = df.groupby(
+            ["geo_id", "date"]
+        ).sum().reindex(
```
nmdefries, line 108: The raw series are ending up with missing values overwritten by 0. Looks like it's coming from the `sum()` here, which defaults to ignoring `NaN`s. Since this is desired behavior in the smoothed case, so we'll probably need an `if level == "state"` around here.

nmdefries (commented):
We should add anosmia and ageusia signals together to make this indicator directly comparable with the old Google Health Trends indicator. GHT only has raw and smoothed search; individual search terms are not provided. This will also simplify the GS indicator if we decide to add many more symptom search topics.

Expected behavior of adding anosmia and ageusia: The raw sum should only be NA (the smoothed sum should be 0) if both/all search term values are missing. If any are not missing, the sum should ignore any NA (since they are presumed to be small relative to available values).

Tests and linting look good.

It takes a lot of memory (~10 GB at the moment) to load the full 2020 dataset. We want to make sure that won't be a problem when we're rolling the indicator out, so we should give the automation team a heads up.

nmdefries:
> The only concern is that there is no county level data for District of Columbia.

It looks like DC doesn't have counties, exactly, but wards. How is this treated in other indicators? Are the wards reported at the county level?

google_symptoms/DETAILS.md
huisaddison: We do have a few folks at Google who are our primary contacts for Google Symptoms.  @jsharpna do you think this is something appropriate to bring to them (and in general, what do you think of using population-weighted averaging as an aggregation scheme)?

My two cents: I think population-weighted averaging is totally reasonable, since the signal is already on the same scale for each county.  Also, population-weighted averaging was the default for aggregating such signals in March/April - it's intuitive, preserves the scale, etc...

The ideal would be to obtain MSA-level signals directly from Google.  (As we discussed >1 month ago, this is the kind of ask from Google that would be reasonable (along with, e.g., obtaining "union / intersection" signals), but we shouldn't make such an ask empty handed (i.e., we should demonstrate the usefulness of these additional features to motivate the additional work that we're asking them to do for us).

krivard:
> It takes a lot of memory (~10 GB at the moment) to load the full 2020 dataset.

🤔  would it be reasonable to use a streaming approach instead?

jingjtang:
Thanks @capnrefsmmat . @nmdefries Now the code is updated to have both dc states and dc county (FIPS 11001). 

Aggregation to HRRs/MSAs are finished. The solution I used is population-weighted averaging as mentioned by @huisaddison. For MSA with part of counties  unavailable, we will fill those NaNs with 0. So for example, if MSA xxxxx consists of fips 00001, 00002, 00003, 00004 with vals [0.2, 0.5, nan, nan], the final value for MSA xxxxx would be the weighted sum of [0.2, 0.5, 0, 0]. We have ~100 counties out of ~1k counties with corresponding MSAs and ~58 MSAs with non-nan values assigned out of ~400 MSAs.

Here at MSA/HRR level, since I use a "transition matrix" (made based on population distribution), there would be no NANs but only 0s. 

Here are plots showing the comparison at MSA level. Each line are re-scaled using min-max normalization.
[raw_search-combined.pdf](https://github.com/cmu-delphi/covidcast-indicators/files/5459441/raw_search-combined.pdf)

jingjtang:
@nmdefries Another signal named `combined_symptoms` is added for `A + A`. A comparison between the `combined_symptoms_raw_search` with `ght-raw-search` and `ght-smoothed-search` is shown below. 
[msa_combined_symptoms_raw_search-combined.pdf](https://github.com/cmu-delphi/covidcast-indicators/files/5459556/msa_combined_symptoms_raw_search-combined.pdf)

jingjtang:
> > It takes a lot of memory (~10 GB at the moment) to load the full 2020 dataset.
> 
> 🤔 would it be reasonable to use a streaming approach instead?

Yes we could. The code has been updated to not store the full datasets. Now it requires ~2-3 mins to finish running the pipeline.

nmdefries:
> Aggregation to HRRs/MSAs are finished. 

Thanks!

> For MSA with part of counties unavailable, we will fill those NaNs with 0. So for example, if MSA xxxxx consists of fips 00001, 00002, 00003, 00004 with vals [0.2, 0.5, nan, nan], the final value for MSA xxxxx would be the weighted sum of [0.2, 0.5, 0, 0].

That sounds reasonable, and is in line with the sum procedure used for the raw combination indicator.

> Here at MSA/HRR level, since I use a "transition matrix" (made based on population distribution), there would be no NANs but only 0s.

Can we make the non-smoothed aggregated values be NaN where all component geo areas are NaN? That way it matches the raw values for the non-aggregated geo regions.

jingjtang:
@nmdefries Updated. Now if all counties have value NaN in a certain MSA/HRR, the aggregated value would be NaN instead of 0. 

The pdfs uploaded previously have errors with ght `smoothed_search`. And the points with NaN are incorrectly shown. They are re-uploaded now.

nmdefries (approved):
Looks good!

Issue #258, opened 2020-09-02
IMPROVEMENT: remove wildcard imports


melange396:
using the wildcard import can introduce problems when debugging. it puts objects into scope without being declared or assigned, and hunting those down can be difficult. this can be even worse if multiple imports have name collisions.  

for example, instead of: ```delphi_emr_hosp.constants import *```
we can use: ```delphi_emr_hosp.constants import NA, SIGNALS```
or even: ```from delphi_emr_hosp import constants as c # then use c.NA and c.SIGNALS```

the choice is up to the implementer for brevity's or readability's sake.
there are many of these in the repo

chinandrew:
emr_hosp is no longer active per footnote in #306, and the only other occurrences of these in non-test code are in the combo signal which I've hopefully addressed in #346. I see 3 other occurences: 2 in claims_hosp tests and 1 in cdc_covidnet tests, which should also be a quick fix

krivard:
Mmmm we should convert this to a project + individual issues, so that citing "closes #X" in a PR doesn't close this one prematurely.

PR #346, opened 2020-10-21
Combo signal: Remove wildcard imports and fix some linting

chinandrew:
Addresses #258 for combo signal

Summary of changes
- Remove wildcard imports
- Fix some linting stuff with newlines and exception handling


chinandrew:
I think this is fixed by #343

PR #349, opened 2020-10-21
Remove wildcard imports from test

### LINKED ISSUE #258: IMPROVEMENT: remove wildcard imports

chinandrew:
Closes #258 by removing remaining non emr_hosp (deactivated) wildcard imports. 

Summary of changes
- Remove wildcard imports
- Remove an unused import

### DIFF
--- a/cdc_covidnet/tests/test_handle_wip_signal.py
+++ b/cdc_covidnet/tests/test_handle_wip_signal.py
@@ -1,6 +1,6 @@
 import unittest
 from delphi_cdc_covidnet.update_sensor import add_prefix
-from delphi_cdc_covidnet.constants import *
+from delphi_cdc_covidnet.constants import SIGNALS
 
 def test_handle_wip_signal():
     # Test wip_signal = True, add prefix to all signals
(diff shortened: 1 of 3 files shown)


krivard:
This also removes from cdc covidnet -- was that intentional? it's probably fine since both are needed...

chinandrew:
> This also removes from cdc covidnet -- was that intentional? it's probably fine since both are needed...

Yea since it was a one liner in a 3 line PR I just batched them :see_no_evil:

krivard (approved):
xlnt

PR #352, opened 2020-10-21
CHC SFTP Downloads

rumackaaron:
Download CHC files through SFTP to allow for automation of CHC signal


rumackaaron:
@krivard @chinandrew The script now downloads files from the CHC ftp server and generates the CHC signal from those files. I'm not sure what we want to do about testing since I don't want to push the credentials to this repo.

krivard:
You are right not to upload them to the repo in plaintext; we'll add them to the [Ansible vault](https://github.com/cmu-delphi/covidcast-indicators/blob/main/ansible/vault.yaml) instead. @korlaxxalrok can get you set up.

changehc/delphi_changehc/run.py
```diff
@@ -25,6 +26,10 @@ def run_module():
 
     logging.basicConfig(level=logging.DEBUG)
 
+    ## download recent files from FTP server
+    logging.info("downloading recent files through SFTP")
+    download(params["cache_dir"], params["ftp_conn"])
```
krivard, line 31: Can this be mocked out for testing?

changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def download(out_path, ftp_conn):
+
+    # open client
+    client = paramiko.SSHClient()
+    client.set_missing_host_key_policy(AllowAnythingPolicy())
+
+    client.connect(ftp_conn["host"], username=ftp_conn["user"],
+                   password=ftp_conn["pass"][1:] + ftp_conn["pass"][0],
```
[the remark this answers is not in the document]
rumackaaron: This is my super hi-tech encryption method. I store the password in plaintext as "dpasswor" and then the client connects using "password." It's almost as foolproof as RSA, just don't tell anyone!

changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -0,0 +1,71 @@
...
+    for infile, outfile in filepaths_to_download.items():
+        callback_for_filename = functools.partial(print_callback, infile)
+        sftp.get(infile, outfile, callback=callback_for_filename)
+
+
+def download(out_path, ftp_conn):
+
+    # open client
+    client = paramiko.SSHClient()
+    client.set_missing_host_key_policy(AllowAnythingPolicy())
```
chinandrew, line 57: Would it be better to use `.set_missing_host_key_policy(paramiko.AutoAddPolicy())` here?

changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -0,0 +1,71 @@
...
+def get_files_from_dir(sftp, out_path):
+    current_time = datetime.datetime.now()
+    seconds_in_day = 24 * 60 * 60
+
+    # go through files in recieving dir
+    files_to_download = []
+    for fileattr in sftp.listdir_attr():
+        file_time = datetime.datetime.fromtimestamp(fileattr.st_mtime)
+        time_diff_to_current_time = current_time - file_time
+        if time_diff_to_current_time.total_seconds() <= seconds_in_day:
```
chinandrew, line 33: For readability, would suggest doing

```
if current_time - file_time > datetime.timedelta(days=1)
```

changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -0,0 +1,71 @@
...
+            files_to_download.append(fileattr.filename)
+
+    filepaths_to_download = {}
+    for file in files_to_download:
```
chinandrew, line 37: Is this checking if we already have the file? low pri, but for brevity this could get added to the end of the above loop where it checks the <24 hr condition, so it becomes
```
for files in sftp
    if <24hrs old and not exists already
        add to list to download
```

changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -0,0 +1,71 @@
...
+                   allow_agent=False, look_for_keys=False)
+    sftp = client.open_sftp()
+
+    sftp.chdir('/dailycounts/All_Outpatients_By_County')
+    get_files_from_dir(sftp, out_path)
+
+    sftp.chdir('/dailycounts/Covid_Outpatients_By_County')
+    get_files_from_dir(sftp, out_path)
+
+    client.close()
```
chinandrew, line 71: This is probably fine for now, but to handle connections being left open due to some error which prevents this line from running, could either make a [context manager](https://extsoft.pro/safely-destroying-connections-in-python/) or wrap in a [try/except block](https://stackoverflow.com/questions/7159644/does-paramiko-close-ssh-connection-on-a-non-paramiko-exception)(easier option)

changehc/delphi_changehc/run.py
chinandrew: +1

changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -0,0 +1,71 @@
...
+    filepaths_to_download = {}
+    for file in files_to_download:
+        full_path = path.join(out_path, file)
+        if path.exists(full_path):
+            print(f"{file} exists, skipping")
+        else:
+            filepaths_to_download[file] = full_path
+
+    # make sure we don't download more than 2 files per day
+    assert len(files_to_download) <= 2, "more files dropped than expected"
```
chinandrew, line 45: Just wondering, do we have any way to know if no files are dropped? Maybe this is done by the validator @nmdefries ?

changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -0,0 +1,71 @@
...
+
+# third party
+import paramiko
+
+class AllowAnythingPolicy(paramiko.MissingHostKeyPolicy):
+    def missing_host_key(self, client, hostname, key):
+        return
+
+
+def print_callback(filename, bytes_so_far, bytes_total):
```
chinandrew, line 18: Add docstrings for each function

changehc/delphi_changehc/download_ftp_files.py
nmdefries: Yes, the validator checks if there are any dates missing within the date range being checked. At the moment, this is on a data source-basis (if a date is missing from all but one signal, it will still count as "present"), but we will be checking on a signal-basis in the future.

changehc/delphi_changehc/download_ftp_files.py
rumackaaron: Resolved

changehc/delphi_changehc/download_ftp_files.py
rumackaaron: Changed

changehc/delphi_changehc/download_ftp_files.py
rumackaaron: Did try/except

changehc/delphi_changehc/download_ftp_files.py
rumackaaron: Added

changehc/delphi_changehc/download_ftp_files.py
rumackaaron: Changed

changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -0,0 +1,74 @@
...
+
+    current_time = datetime.datetime.now()
+
+    # go through files in recieving dir
+    filepaths_to_download = {}
+    for fileattr in sftp.listdir_attr():
+        file_time = datetime.datetime.fromtimestamp(fileattr.st_mtime)
+        filename = fileattr.filename
+        if current_time - file_time < datetime.timedelta(days=1) and \
+                not path.exists(filename):
```
chinandrew, line 36: does this need to be `path.join(out_path, filename)`? not familiar with the file structures

rumackaaron: Yeah you're right, it was redownloading the file

changehc/delphi_changehc/download_ftp_files.py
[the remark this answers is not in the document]
chinandrew: Do we need to do this if credentials are encrypted at rest in ansible?

chinandrew:
This looks good, just the open question on testing and the password obfuscation

changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -0,0 +1,74 @@
...
+                not path.exists(path.join(out_path, filename)):
+            filepaths_to_download[filename] = path.join(out_path, filename)
+
+    # make sure we don't download more than 2 files per day
+    assert len(filepaths_to_download) <= 2, "more files dropped than expected"
+
+    # download!
+    for infile, outfile in filepaths_to_download.items():
+        callback_for_filename = functools.partial(print_callback, infile)
+        sftp.get(infile, outfile, callback=callback_for_filename)
```
chinandrew, line 45: callback is cool, didn't know of this feature in paramiko

rumackaaron:
Sorry for the delay in creating the tests, I've never worked with mocks before. Let me know what else you'd like me to do

chinandrew:
I'm getting the following test errors
```
ERROR test_load_data.py - KeyError: 'input_covid_file'
ERROR test_sensor.py - KeyError: 'input_covid_file'
ERROR test_update_sensor.py - KeyError: 'input_covid_file'
```

rumackaaron:
Can you copy the current `params.json.template` to `params.json` and see if the errors go away?

chinandrew:
> Can you copy the current `params.json.template` to `params.json` and see if the errors go away?

Thanks, that did the trick. Any reason we can't just keep it in the `params.json` (the one in the tests folder)?

rumackaaron:
> > Can you copy the current `params.json.template` to `params.json` and see if the errors go away?
> 
> Thanks, that did the trick. Any reason we can't just keep it in the `params.json` (the one in the tests folder)?

Theoretically no, but I assume there's some reason that `params.json` is included in `.gitignore`

capnrefsmmat:
Mostly to prevent production params from getting into the repo. I think test params are fine to force-commit.

chinandrew:
So I had an old params.json present which I think was causing the issue. Normal behavior will just have it read the template if a params isn't present, which I think is fine. Just fix linting and should be good to go.

chinandrew (approved):
Linting still raising some issues on things like line length, but I think they snuck in the previous PR and are not due to this one, so I'm fine having those fixed in a separate PR. Tests are passing.

PR #354, opened 2020-10-22
cdc_covidnet: change states to lowercase to align with geo_util default

chinandrew:
Merge to main after #310 (which is currently the base for simple reviewing)

Summary of changes:
- remove .upper() from state abbr column and just use geo_util default.

cc @dshemetov


dshemetov:
Looks good!

PR #345, opened 2020-10-21
Fix safegraph 7d averaged signals

sgsmob:
Fixes #337.

Change summary
* map the `process()` function using `functools.partial` rather than a local function.
* Revert the output date format to be YYYYMMDD rather than YYYY-MM-DD.

PR #356, opened 2020-10-22
Return of "fix safegraph 7d averaged signals"

krivard:
Supersedes #345 

* Branch starts at current contents of `deploy-safegraph`
* Reverts d243e3f to bring us back to where we were after merging #309 (with broken safegraph)
* Cherry-picks @sgsmob 's fixes

Package lints and passes all tests. @sgsmob do you want to pull this branch and make sure it runs? In theory it should but I don't have the keys to be sure.


sgsmob (approved):
Runs on my machine, LGTM.

PR #363, opened 2020-10-23
Gate pulling/syncing data from AWS in safegraph behind a flag in params

sgsmob:
Summary:
* Adds a flag around pulling/syncing data from AWS in `safegraph.run_module()`.
* Adds testing for `safegraph.run_module()` that skips over any interaction with AWS.
* Update `safegraph_patterns` to use similar variable naming and other paradigms.


chinandrew:
Oh didn't even see I had a review requested on this one, sorry. I should add an inbox filter for github emails....

safegraph/delphi_safegraph/run.py
```diff
@@ -15,50 +16,60 @@
...
     n_core = int(params["n_core"])
+
+    # AWS credentials
     aws_access_key_id = params["aws_access_key_id"]
     aws_secret_access_key = params["aws_secret_access_key"]
     aws_default_region = params["aws_default_region"]
     aws_endpoint = params["aws_endpoint"]
-    wip_signal = params["wip_signal"]
+    # Whether to sync `raw_data_dir` with an AWS backend.
+    sync = bool(params["sync"])
```
chinandrew, line 34: If `params["sync"]` is just getting read in as a str, Python will treat all non-empty str as true:

```
In [1]: bool("false")                                                                                                                                                
Out[1]: True

In [2]: bool("False")                                                                                                                                                
Out[2]: True

In [3]: bool("")                                                                                                                                                     
Out[3]: False
```

Would have to document the params clearly if this is the case. Alternatively could make the key optional the use `.get`, but that also requires equal documentation.

safegraph/delphi_safegraph/run.py
```diff
@@ -15,50 +16,60 @@
...
+            shell=True,
+            check=True,
+        )
 
     files = glob.glob(f'{raw_data_dir}/social-distancing/**/*.csv.gz',
                       recursive=True)
 
+    files_with_previous_weeks = []
+    for fname in files:
+        previous_week = [fname]
```
chinandrew, line 70: Previously only `files_in_past_week(current_filename)` was being passed to the `single_arg_process` function (though it was called `process), and now we're adding fname to the front. is that intended?

safegraph_patterns/tests/params.json.template
```diff
@@ -8,5 +8,5 @@
   "aws_secret_access_key": "",
   "aws_default_region": "",
   "aws_endpoint": "",
-  "test": "True"
+  "sync": false
```
chinandrew, line 11: worth adding a test where sync is true and testing by mocking and checking with `assert_called*`?

safegraph/delphi_safegraph/run.py
sgsmob: I noticed that in development, so I have set the value to be true/false in JSON file.  I'm not sure whether removing the explicit cast to bool makes things better or not.  I can add a check here to verify that the value is bool.

safegraph/delphi_safegraph/run.py
sgsmob: This change was made in #345; I don't know why it is showing up as a diff here.

safegraph/delphi_safegraph/run.py
chinandrew: I misread one of the examples and thought it was being read in as a str instead of bool (used to be `  "test": "True"`). I think it's fine as is. I think the cast to bool is potentially confusing since it's redundant. Could type annotate or comment that line if it's something we want to highlight

safegraph_patterns/tests/params.json.template
sgsmob: In the near future, yes; as part of this PR, I'm ambivalent.  Was looking to revamp the data fetching package-wide in #366, but I can put something temporary in here now if you think it would be bad to leave untested now.

chinandrew: I don't feel too strongly, it's a single `if sync` statement so fine to leave it for now, and the data fetching thing makes sense as well.

safegraph/delphi_safegraph/run.py
```diff
@@ -15,50 +16,60 @@
...
-        f'{raw_data_dir}/social-distancing/ --endpoint {aws_endpoint}',
-        env={
-            'AWS_ACCESS_KEY_ID': aws_access_key_id,
-            'AWS_SECRET_ACCESS_KEY': aws_secret_access_key,
-            'AWS_DEFAULT_REGION': aws_default_region,
-        },
-        shell=True,
-        check=True,
-    )
+    if (sync):
```
chinandrew, line 52:
```suggestion
    if sync:
```

sgsmob: Done

chinandrew (approved):

PR #365, opened 2020-10-23
add pylintrc file to each indicator

sgsmob:
This should make the linter complain less about:

* Short variable names
* Pytest naming/argument conventions


krivard:
It looks like this is using an identical .pylintrc for all indicators; the following modules had existing .pylintrc that have been overwritten. Was that intentional?
* `_delphi_utils_python`
* `cds` (can probably ignore this; cds is no longer being pursued)
* `google_health`
* `jhu`
* `nchs_mortality`
* `quidel_covidtest`
* `safegraph`
* `safegraph_patterns`
* `usafacts`

_delphi_utils_python/.pylintrc
```diff
@@ -1,8 +1,21 @@
...
 
-min-public-methods=1
+[MESSAGES CONTROL]
 
+disable=logging-format-interpolation,
+    too-many-locals,
+    too-many-arguments,
+    # Allow pytest functions to be part of a class.
+    no-self-use,
+    # Allow pytest classes to have one test.
```
chinandrew, line 9: this isn't scoped just to the test dir right? so adding these lines to have tests lint correctly potentially allows these into the actual module code?

secondarily, how strongly do we feel we need to lint unit tests?

_delphi_utils_python/.pylintrc
```diff
@@ -1,8 +1,21 @@
...
+
+# Allow arbitrarily short-named variables.
+variable-rgx=[a-z_][a-z0-9_]*
+argument-rgx=[a-z_][a-z0-9_]*
+
+[DESIGN]
 
-disable=R0801, C0330, E1101, E0611, C0114, C0116, C0103, R0913, R0914, W0702
+# Don't complain about pytest "unused" arguments.
+ignored-argument-names=(_.*|run_as_module)
```
chinandrew, line 21: why is it complaining about `run_as_module`?

sgsmob: It is never referenced in the function body.

sgsmob:
> It looks like this is using an identical .pylintrc for all indicators; the following modules had existing .pylintrc that have been overwritten. Was that intentional?
> 
> * `_delphi_utils_python`
> * `cds` (can probably ignore this; cds is no longer being pursued)
> * `google_health`
> * `jhu`
> * `nchs_mortality`
> * `quidel_covidtest`
> * `safegraph`
> * `safegraph_patterns`
> * `usafacts`

Yes, I felt that we should have a consistent style across all indicators.

_delphi_utils_python/.pylintrc
sgsmob: Yes, they would allow it into the module code, but they seem like narrow enough and innocuous enough problems that they could be caught in review.

If one uses an integrated linter in their IDE, it will read from these files and complain about the tests.  FWIW we enforce style the same across implementation and test files.

_delphi_utils_python/.pylintrc
chinandrew: Ah, it's in a conftest. Got it.

_delphi_utils_python/.pylintrc
chinandrew: I'm not a huge fan of relying on code review to catch problems that could have been linted, but agree they're pretty innocuous. One option is to add an additional pylintrc to the tests module, so any weird test things can get captured without worrying about affecting the module code. Not sure how clean that is though.

sgsmob: I thought about doing this but thought it created a lot of lint files (2 per indicator rather than the necessary 1).  Can do if you think it is worthwhile.

chinandrew: Yea that's kind of annoying...I think they're pretty harmless for now, lets just leave as is.

chinandrew (approved):

krivard:
> Yes, I felt that we should have a consistent style across all indicators.

Sounds good! In that case, would you do a pass through the ones below and make sure they lint successfully with the new configuration? Otherwise the next PR on their `deploy-*` branch will have surprise failures unrelated to the feature being added.

* google_health
* jhu
* quidel_covidtest
* safegraph
* usafacts

jingjtang:
A quick check for `google_health` and `quidel_covidtest`, both of them run into linter errors. Most of the errors are due to `C0103` (invalid-name), a few of them are due to `E1101`, `W0212` etc. 

Specially, `W0212` results from protected-access to to member `_load_crosswalk` in GeoMapper. This should be discussed with @dshemetov to make `_load_crosswalk` public (or not).

jingjtang (approved):

sgsmob:
> > Yes, I felt that we should have a consistent style across all indicators.
> 
> Sounds good! In that case, would you do a pass through the ones below and make sure they lint successfully with the new configuration? Otherwise the next PR on their `deploy-*` branch will have surprise failures unrelated to the feature being added.
> 
> * google_health
> * jhu
> * quidel_covidtest
> * safegraph
> * usafacts

@krivard should I include the lint compliance changes in this PR?

krivard (approved):
lgtm

PR #368, opened 2020-10-26
Migrate indicators to a common `add_prefix()` library

sgsmob:
This PR refactors a series of identical `add_prefix()` and `public_signal()` definitions duplicated across indicators into single definitions in `delphi_utils`.

Also fixes lint errors in the files already touched by this PR.


chinandrew:
@sgsmob  FYI you have a merge conflict

_delphi_utils_python/tests/test_signal.py
```diff
@@ -0,0 +1,27 @@
...
+        assert signal_names[0].startswith("wip_")
+        assert all(not s.startswith("wip_") for s in signal_names[1:])
+        # Test wip_signal = False
+        signal_names = add_prefix(["xyzzy", SIGNALS[0]], False, prefix="wip_")
+        assert signal_names[0].startswith("wip_")
+        assert all(not s.startswith("wip_") for s in signal_names[1:])
+
+    def test_public_signal(self):
+        """Tests that `public_signal()` identifies public vs. private signals."""
+        assert not public_signal("junk")
```
chinandrew, line 26: Is this test making a network call since the function has `epidata_df = covidcast.metadata()` in it?

sgsmob: yes.  This method was previously untested across all (?) of its instances

chinandrew: can we mock out the metadata call?  Should be able to do something like 

```
from unittest.mock import patch
@patch("covidcast.metadata")  # maybe needs to be delphi_utils.metadata or something, patch references are weird
def test_public_signal(self, metadata):
    metadata.return_value ={...}
```

Now that I think of it, might be good to just make all these things into a conftest so everyone can just import the fixtures....

quidel_covidtest/tests/cache/pulled_until_20200817.csv
```diff
@@ -0,0 +1,13 @@
+timestamp,zip,totalTest,numUniqueDevices,positiveTest
```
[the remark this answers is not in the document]
sgsmob: not sure, I thought I removed it.  Removed

changehc/tests/test_update_sensor.py
```diff
@@ -99,15 +99,16 @@ def test_update_sensor(self):
...
 
-            assert len(os.listdir(td.name)) == len(su_inst.sensor_dates), f"failed {geo} update sensor test"
+            assert len(os.listdir(td.name)) == len(su_inst.sensor_dates),\
+                f"failed {geo} update sensor test"
             td.cleanup()
 
 class TestWriteToCsv:
+    """Tests for writing output files to CSV."""
     def test_write_to_csv_results(self):
+        """Tests that the computed data written."""
```
chinandrew, line 111:
```suggestion
        """Tests that the computed data are written correctly."""
```

changehc/delphi_changehc/update_sensor.py
```diff
@@ -159,6 +106,12 @@ def __init__(self,
             wip_signal=read_params()["wip_signal"])
         self.updated_signal_names = signal_names
 
+        # initialize members set in shift_dates().
+        self.burnindate = None
```
chinandrew, line 110: are these supposed to be part of this pr?

sgsmob: I was trying to get the linter to pass in every file I touched in this PR.  Updated description to explain.

_delphi_utils_python/tests/test_signal.py
sgsmob: Done.

chinandrew:
`covidcast` needs to be added as a dependency in setup.py

_delphi_utils_python/tests/test_signal.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @patch("covidcast.metadata")
+    def test_handle_wip_signal(self, metadata):
+        """Tests that `add_prefix()` derives work-in-progress signals."""
+        metadata.return_value = SIGNALS_FRAME
+
+        # Test wip_signal = True
+        signal_names = SIGNALS
+        signal_names = add_prefix(SIGNALS, True, prefix="wip_")
+        assert all(s.startswith("wip_") for s in signal_names)
```
chinandrew, line 22: This is a refactor of existing stuff so I don't consider this blocking (could do in a later PR), but these tests would probably be easier if it was just a direct list comparison, e.g. `assert signal_names == ['wip_sig1', 'wip_sig2', 'wip_sig3']`. Maybe it wasn't done this way because things weren't mocked out so signal names changes?

sgsmob: That probably was the reason.  I have refactored them into separate tests and made it clearer with a direct comparison.

chinandrew (approved):

Issue #333, opened 2020-10-20
Pylint/Astroid gives "RecursionError: maximum recursion depth exceeded" in CI tests


benjaminysmith:
Recent discussion in PR review: https://github.com/cmu-delphi/covidcast-indicators/pull/314#issuecomment-712877800

Example of error: https://jenkins-ci-prod-01.delphi.cmu.edu/blue/organizations/jenkins/covidcast-indicators/detail/PR-314/3/pipeline/

Discussion on previous PR: https://github.com/cmu-delphi/covidcast-indicators/pull/272#issuecomment-691215885

chinandrew:
Looks to be an open issue https://github.com/PyCQA/pylint/issues/3836. Been banging my head on this for a while to no avail, including upgrading astroid directly from their github, trying to align package versions, and changing the python version. 

I've been able to reproduce this by pulling the `pip list` from the CI and intsalling those locally, so going to try to narrow down the suspect packages.

chinandrew:
This appears to be due to pandas _and_ the ubuntu version. Locally downgrading to 1.1.1 makes things works, 1.1.2 or above breaks it. However when I pin this on the CI it didn't work :/. The `pip list`s locally and on CI were identical. So then I learned `ubuntu-latest` on Github Actions is ubuntu 18.04, not 20.04 which is what I'm running. Upgraded that and now things are working. @korlaxxalrok FYI

sgsmob:
It looks like `df = pd.merge(df, other)` is a common cause of this problem and replacing it with the equivalent `df = df.merge(other)`.  Going through and seeing if this is sufficient to at least stop the issue in the near-term.

chinandrew:
> It looks like `df = pd.merge(df, other)` is a common cause of this problem and replacing it with the equivalent `df = df.merge(other)` removes the error. Going through and seeing if this is sufficient to at least stop the issue in the near-term.

That's interesting, thanks for looking into it :+1:

chinandrew:
As another confusing data point, I just hit this error on the covidcast repo where the linter didn't complain on a pd.merge(x,y), but then complained when i moved that line into a new file..... :angry:

PR #369, opened 2020-10-26
[DRAFT/TEST] try out CI config for testing packages

chinandrew:
Just testing out some github actions, don't merge. cc @korlaxxalrok @benjaminysmith @krivard.

Aside from that issue, something like this should satisfy the testing need for PRs. See checks below for summaries. I'm only running 2 indicators and the utils package at the moment for demonstration.

I've brought up the subject of costs for this before. Even though it's not super high it's still worth keeping in mind. If it is an issue, one thing we could do if we move to a gitflow branching approach is only run tests on merge with the develop branch. This way we can catch things before they hit main without needing to run 10 checks each time someone commits changes to a PR. It unfortunately gives up the ability to see all the checks passing in PR phase. There may be some intermediate approach where only checks on altered packages are run, but that's going to be a bit more complex and not foolproof (e.g. if i change the utils i can still break other packages).

I'm still trying to figure out the linter. On some packages (usafacts), I hit this [weird recursion error](https://stackoverflow.com/questions/36496192/pylint-infinite-recursion-in-astriod-package) that I can't reproduce locally and have been unable to debug.


chinandrew:
Also just went through the `jenkins` folder and that seems perfectly fine to extend to this application as well. I don't personally have any experience with it though.

korlaxxalrok:
@chinandrew This looks promising 👍 

>There may be some intermediate approach where only checks on altered packages are run

Is this referring to being able to run CI actions based on what is being "worked on" (ie changes to a single indicator)? That would be pretty useful, and was something I was thinking about trying to figure out for Jenkins.

chinandrew:
> @chinandrew This looks promising 
> 
> > There may be some intermediate approach where only checks on altered packages are run
> 
> Is this referring to being able to run CI actions based on what is being "worked on" (ie changes to a single indicator)? That would be pretty useful, and was something I was thinking about trying to figure out for Jenkins.

Yea, so something like changing the code for `cdc_covidnet` doesnt run the `safegraph` tests. We just have to be careful that utility changes always run all indicators.

chinandrew:
Alright linting is in place now (which is why checks aren't passing). See #333  for more details.

chinandrew:
opening a new pr in #449

PR #372, opened 2020-10-27
Update quidel docstrings to pass pydocstyle

chinandrew:
Summary of changes:
- geomapping and data tools (smoothing) functions are excluded until refactors go through


chinandrew:
going to close this until we figure out branching

PR #374, opened 2020-10-27
Update cdccovidnet docstrings to pass pydocstyle

chinandrew:
Summary of changes:


cdc_covidnet/delphi_cdc_covidnet/covidnet.py
```diff
@@ -41,7 +38,8 @@ def download_mappings(
     @staticmethod
     def read_mappings(infile: str) -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]:
         """
-        Reads the mappings JSON file from disk to produce formatted
+        Read the mappings JSON file from disk to produce formatted.
```
sgsmob, line 41: This is one sentence.

sgsmob (changes requested):
Mostly good, minus the one comment.  Should be approve once it is fixed.

cdc_covidnet/delphi_cdc_covidnet/covidnet.py
chinandrew: Done

sgsmob (approved):

PR #316, opened 2020-10-15
Refactor usafacts to use geo utils

sgsmob:
The USAFacts indicator no longer uses localized libraries that have been superseded by `delphi_utils.GeoMapper`.

Fixes #338, which itself addresses part of #306.  Synced to and #309 and #314, so should be submitted after them.

PR #377, opened 2020-10-27
Update usafacts docstrings to pass pydocstyle

chinandrew:
Summary of changes:
- geomapping functions are excluded since they'll be removed once the refactor goes through.


jingjtang (dismissed):

krivard (dismissed):
Changes to USAFacts docstrings are good; changes to geomapper are blocked on #316.

chinandrew:
@krivard Is the base branch supposed to be main or deploy-usafacts?

krivard:
We're going with `main` for the moment to keep folks unblocked.

PR #379, opened 2020-10-27
Update claims hosp docstrings to pass pydocstyle

chinandrew:
Summary of changes:


mariajahja (approved):
Looks good.

PR #389, opened 2020-10-27
Fix linter issue due to try/except in Quidel_covidtest

jingjtang:
Specify the error type in try/except to prevent unintended exceptions


chinandrew (approved):
lgtm

PR #392, opened 2020-10-28
Propagate #363 to safegraph (whoops)

krivard:
This should've gone in the other direction :(

@chinandrew you were approver on #363; apologies for the duplicate


chinandrew (approved):
so I still don't quite understand the deploy process yet, but I have no reason to suspect the changes will break as long as the new params file template is used.

krivard:
Good catch; I would've missed that 😱

PR #398, opened 2020-10-28
Propagate 328 (JHU population fix and function refactor) to main

krivard:


dshemetov:
Oops, meant to find a way to see the conflicts instead of that.

dshemetov:
Checked out the conflict. The incoming change is the correct one, the rest was refactored into functions.

PR #400, opened 2020-10-28
Fix geomapper add_population_column function

dshemetov:
Fixed a hasty function signature change I introduced `GeoMapper.add_population_column`!


krivard (approved):
Good catch on the `data.copy()` too!

PR #401, opened 2020-10-28
Refactor doctor visits to use geo utils package

### LINKED ISSUE #385: Refactor doctor_visits to use geo utils
Convert doctor_visits indicator to use GeoMapper utility

chinandrew:
Refactor doctor visits to use the new geoutils package. Closes #385 

Summary of changes:
- county to msa is now strings and not floats, so outputs are '11500' and not 11500.0, which should ingest fine.
- count to state is now lowercase, which should also ingest fine.
- county to HRR has an updated mapping table from the utils package so some weights are slightly different
- county to megacounty now removes some unnecessary columns and adjusts the order of the rows (megacounty now at the top instead of bottom), both of which I don't believe affect anything downstream. Also, I think there was a bug where rows that were meant to be excluded were not, which I've added a comment for below.

I branched this off dv-package but had to merge main in to get the most up to date geo utils, so the changes are...unreadable. Could remedy this by merging main into dv-package. Pretty sure there are only 2 files I touched, doctor_visits/delphi_doctor_visits/geo_maps.py and doctor_visits/tests/test_geomap.py, both of which I've left a comment on below so they can be found easily.


doctor_visits/delphi_doctor_visits/geo_maps.py
```diff
@@ -160,37 +165,11 @@ def county_to_megacounty(self, data, threshold_visits, threshold_len):
...
-            lambda f: str((int(f) // 1000) * 1000).zfill(5)
-        )
-
-        megacounty_df = (
-            data[data["ToExclude"]]
-            .groupby(["ServiceDate", "StateFIPS"])
-            .sum()
-            .reset_index()
-        )
-        megacounty_df["ToExclude"] = False
```
chinandrew, line 190: I think the values where ToExclue=True were meant to be filtered out but weren't here.

doctor_visits/tests/test_geomap.py
```diff
@@ -26,34 +26,28 @@ def test_county_to_msa(self):
...
 
         out, name = GM.county_to_hrr(DATA)
         assert name == "hrr"
-        assert set(out.groups.keys()) == set(["1", "134", "144", "2", "6", "7", "9"])
+        assert set(out.groups.keys()) == {"1", "134", "144", "146", "2", "5", "6", "7", "9"}
 
     def test_county_to_megacounty(self):
 
-        out, name = GM.county_to_megacounty(DATA, 10, 10)
+        out, name = GM.county_to_megacounty(DATA, 100000, 10)
```
chinandrew, line 45: Previous threshold led to no megacounties being made

mariajahja:
@chinandrew Thanks for the breakdown. Are any state-level changes expected? Most of the values are the same, but I do see some differences (I noticed Alaska). Also, we have more HRRs than before. I don't think these changes are the result of this PR, but maybe the geo utils themselves?

doctor_visits/delphi_doctor_visits/geo_maps.py
mariajahja: I'm not sure. In line 185, we subset by the ["ToExclude"] column, which should have removed all the cases we wanted to filter out. The line at 190 may just be redundant?

chinandrew:
> @chinandrew Thanks for the breakdown. Are any state-level changes expected? Most of the values are the same, but I do see some differences (I noticed Alaska). Also, we have more HRRs than before. I don't think these changes are the result of this PR, but maybe the geo utils themselves?

For HRR's, @dshemetov and I discussed how the mapping files were updated and some very minor changes were to be expected (e.g. fips 01011 got converted to 146,0.003024 and 7,0.996976 (HRR,weight) in the new mapping, while the original gets converted to 7,1.0000). Not sure about the state, it's possible that mapping file was updated? how large is the difference?

mariajahja (approved):
Hmm, OK. The differences are 0.01-0.15 (I see another state is also different). Again, I don't think it's this PR, so I'll hit approve with the understanding that the only changes I've checked are in the two files.

PR #404, opened 2020-10-29
JHU: Appease the linter

krivard:
Fix some linting errors that were introduced in #397 & #398 during the conflict resolution process (bad github; no biscuit)

@dshemetov I'm pretty sure the `r` prefix in that regex in `pull.py` is good but wanted to get your eyes on it. Because it's a network call, we don't have unit tests for it. I ran it by hand and it seemed to generate sensible output in the `timestamp` column.


dshemetov:
@krivard wasn't sure about the `r` prefix before, thanks for catching! I have a test for that regex [here](https://github.com/cmu-delphi/covidcast-indicators/blob/e246208c8dace61628f572c9a002a3ceedb85268/jhu/tests/test_pull.py#L14) (that has an `r` prefix...) so it should be fine.

dshemetov (approved):

PR #422, opened 2020-11-02
Retain gitkeep file for JHU

### LINKED ISSUE #403: Make JHU tests retain .gitkeep instead of deleting it
When JHU runs tests, it deletes the entire tests/receiving directory, and not just the .csv files. When this happens, it also deletes `tests/receiving/.gitkeep`, which should be retained. We should fix the tests so that `.gitkeep` doesn't wind up getting removed from the repository by accident.

chinandrew:
closes #403

Summary of changes:
- Add a statement to prevent the .gitkeep file from being removed when tests run.


krivard (approved):
lgtm

PR #426, opened 2020-11-03
Find Unexpected Values for geo_id

JedGrabman:
- Script for finding geo_ids
- Resulting CSV files with lists of geo_ids broken down by geo_type
- Update template to allow explicitly setting validator directory
- Implement check for finding unexpected geo_id values

Note: Due to technical issues I was not able to run the validator to test this. Looking for code to be run in addition to reviewed.


JedGrabman:
@nmdefries as suggested reviewer, since I cannot set it directly.

validator/delphi_validator/validate.py
```diff
@@ -244,9 +245,26 @@ def check_df_format(self, df_to_test, nameformat):
...
+        """
+        file_path = join(self.validator_static_file_dir, geo_type + '_geo.csv')
+        valid_geo_df = pd.read_csv(file_path, dtype = {'geo_id': str})
+        valid_geos = valid_geo_df['geo_id'].values
+        unexpected_geos = [geo for geo in df_to_test['geo_id'] if geo not in valid_geos]
+        if len(unexpected_geos) > 0:
+            self.raised_errors.append(ValidationError(
+                ("check_bad_geo_id_value", filename),
+                unexpected_geos, "Unrecognized geo_ids (not in historical data)"))
+
```
nmdefries, line 264: Increment check counter with `self.increment_total_checks()`.

validator/delphi_validator/validate.py
```diff
@@ -244,9 +245,26 @@ def check_df_format(self, df_to_test, nameformat):
 
         self.increment_total_checks()
 
-    def check_bad_geo_id(self, df_to_test, nameformat, geo_type):
+    def check_bad_geo_id_value(self, df_to_test, geo_type):
```
nmdefries, line 248:
```suggestion
    def check_bad_geo_id_value(self, df_to_test, filename, geo_type):
```

validator/delphi_validator/validate.py
```diff
@@ -720,8 +738,9 @@ def validate(self, export_dir):
             data_df = load_csv(join(export_dir, filename))
 
             self.check_df_format(data_df, filename)
-            self.check_bad_geo_id(
+            self.check_bad_geo_id_format(
                 data_df, filename, match.groupdict()['geo_type'])
+            self.check_bad_geo_id_value(data_df, match.groupdict()['geo_type'])
```
nmdefries, line 743:
```suggestion
            self.check_bad_geo_id_value(data_df, filename, match.groupdict()['geo_type'])
```

validator/delphi_validator/validate.py
```diff
@@ -244,9 +245,26 @@ def check_df_format(self, df_to_test, nameformat):
...
+            - geo_type: string from CSV name specifying geo type (state, county, msa, etc.) of data
+        """
+        file_path = join(self.validator_static_file_dir, geo_type + '_geo.csv')
+        valid_geo_df = pd.read_csv(file_path, dtype = {'geo_id': str})
+        valid_geos = valid_geo_df['geo_id'].values
+        unexpected_geos = [geo for geo in df_to_test['geo_id'] if geo not in valid_geos]
+        if len(unexpected_geos) > 0:
+            self.raised_errors.append(ValidationError(
+                ("check_bad_geo_id_value", filename),
+                unexpected_geos, "Unrecognized geo_ids (not in historical data)"))
```
nmdefries, line 263: Do you have an idea of how often new `geo_id`s appear? If it's pretty frequent, maybe we want to make this a warning instead of an error (append to `self.raised_warnings` instead of `self.raised_errors`) so it doesn't have to be manually suppressed all the time.

nmdefries (commented):
Looks good! There are just a few cleanup items left.

* Write some tests to check behavior of this function. You can put tests in tests/test_checks.py
* Apply any check name changes to method calls in tests/test_checks.py. There are a few trivial failures right now.

validator/delphi_validator/validate.py
JedGrabman: I initially ran unique_geoids.R by retrieving data from every date, but I found that the list was extremely stable (as in, every geo_id we've ever seen was also seen on October 1). I didn't check every source though.

nmdefries: Okay, that's good to know. So new `geo_id`s should show up rarely and in small numbers, and thus be easy to manually deal with.

validator/delphi_validator/validate.py
krivard: Does this still need to be fixed?

JedGrabman: Fixed in latest commit

nmdefries: Yep, waiting on this and a few other changes!

JedGrabman:
Added changes in response to comments. Please verify by running tests before merging.

validator/tests/test_checks.py
```diff
@@ -146,21 +146,21 @@ def test_expected_groups(self):
         assert pattern_found["signal"] == "signal_signal"
 
 
-class TestCheckBadGeoId:
+class TestCheckBadGeoIdFormat:
     params = {"data_source": "", "span_length": 0,
               "end_date": "2020-09-02", "expected_lag": {}}
```
nmdefries, line 151:
```suggestion
              "end_date": "2020-09-02", "expected_lag": {},
              "validator_static_file_dir": "../static"}
```

Default path to `static` dir doesn't work for tests. Specify explicitly.

validator/tests/test_checks.py
```diff
@@ -230,6 +230,87 @@ def test_invalid_geo_id_national(self):
...
+        assert "88888" in validator.raised_errors[0].expression
+        assert "99999" in validator.raised_errors[0].expression
+
+    def test_invalid_geo_id_hrr(self):
+        validator = Validator(self.params)
+        df = pd.DataFrame(["1", "11", "111", "8", "88", "888"], columns=["geo_id"])
+        validator.check_bad_geo_id_value(df, "name", "hrr")
+
+        assert len(validator.raised_errors) == 1
+        assert "check_geo_id_value" in validator.raised_errors[0].check_data_id
```
nmdefries, line 273:
```suggestion
        assert "check_bad_geo_id_value" in validator.raised_errors[0].check_data_id
```

Update due to check string/method name change.

validator/tests/test_checks.py
```diff
@@ -230,6 +230,87 @@ def test_invalid_geo_id_national(self):
...
+        assert "88" in validator.raised_errors[0].expression
+        assert "888" in validator.raised_errors[0].expression
+
+    def test_invalid_geo_id_state(self):
+        validator = Validator(self.params)
+        df = pd.DataFrame(["aa", "ak"], columns=["geo_id"])
+        validator.check_bad_geo_id_value(df, "name", "state")
+
+        assert len(validator.raised_errors) == 1
+        assert "check_geo_id_value" in validator.raised_errors[0].check_data_id
```
nmdefries, line 288:
```suggestion
        assert "check_bad_geo_id_value" in validator.raised_errors[0].check_data_id
```

Update due to check string/method name change.

validator/tests/test_checks.py
```diff
@@ -230,6 +230,87 @@ def test_invalid_geo_id_national(self):
...
+        assert "check_geo_id_lowercase" in validator.raised_warnings[0].check_data_id
+        assert "AK" in validator.raised_warnings[0].expression
+
+    def test_invalid_geo_id_national(self):
+        validator = Validator(self.params)
+        df = pd.DataFrame(["us", "zz"], columns=["geo_id"])
+        validator.check_bad_geo_id_value(df, "name", "national")
+
+        assert len(validator.raised_errors) == 1
+        assert "check_geo_id_value" in validator.raised_errors[0].check_data_id
```
nmdefries, line 309:
```suggestion
        assert "check_bad_geo_id_value" in validator.raised_errors[0].check_data_id
```

Update due to check string/method name change.

nmdefries (commented):
A few old function name problems in tests. Suggested changes have been tested already and work fine, so once those are incorporated, we should be good to merge! 🎉

PR #437, opened 2020-11-05
Patch smoothing utility to handle short signals (for backfill)

dshemetov:
Mostly documentation fixes, confrm to pydocstyle, add missing documentation for the argument boundary_method.

But also I removed the len(signal) < window_length constraint I added at the last minute; I forgot I wrote boundary methods to handle that case a few months ago.

### DIFF
--- a/_delphi_utils_python/tests/test_smooth.py
+++ b/_delphi_utils_python/tests/test_smooth.py
@@ -167,11 +167,10 @@ class TestSmoothers:
         with pytest.raises(ValueError):
             imputed_signal = smoother.savgol_impute(signal)
 
-        # test window_length > len(signal)
+        # test window_length > len(signal) and boundary_method="identity"
         signal = np.arange(20)
         smoother = Smoother(smoother_name="savgol", boundary_method="identity", window_length=30)
-        with pytest.raises(ValueError):
-            smoothed_signal = smoother.smooth(signal)
+        smoothed_signal = smoother.smooth(signal)
 
         # test the boundary methods
         signal = np.arange(20)
--- a/_delphi_utils_python/tests/test_smooth.py
+++ b/_delphi_utils_python/tests/test_smooth.py
@@ -171,6 +171,7 @@ class TestSmoothers:
         signal = np.arange(20)
         smoother = Smoother(smoother_name="savgol", boundary_method="identity", window_length=30)
         smoothed_signal = smoother.smooth(signal)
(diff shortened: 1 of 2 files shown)


dshemetov:
Should probably be merged before @chinandrew's refactors go live, so that backfill is handled properly.

_delphi_utils_python/tests/test_smooth.py
```diff
@@ -167,11 +167,10 @@ def test_impute(self):
         with pytest.raises(ValueError):
             imputed_signal = smoother.savgol_impute(signal)
 
-        # test window_length > len(signal)
+        # test window_length > len(signal) and boundary_method="identity"
         signal = np.arange(20)
         smoother = Smoother(smoother_name="savgol", boundary_method="identity", window_length=30)
-        with pytest.raises(ValueError):
-            smoothed_signal = smoother.smooth(signal)
+        smoothed_signal = smoother.smooth(signal)
```
chinandrew, line 173: should there be some assertion on the output of this? I know it used to be a value error and now it doesn't, so checking it returned right would further verify behavior

chinandrew (commented):
one comment otherwise lgtm

dshemetov:
Cool. You know, I'm running into issues where JHU tests are giving the smoother an array full of NANs. That makes me want to change the nan handling to be a bit more permissive (a full array of nans should pass through, an array left-padded with nans should start imputing only once the first value is seen, etc.), so I think I'll handle that in this PR too.

chinandrew:
> Cool. You know, I'm running into issues where JHU tests are giving the smoother an array full of NANs. That makes me want to change the nan handling to be a bit more permissive (a full array of nans should pass through, an array left-padded with nans should start imputing only once the first value is seen, etc.), so I think I'll handle that in this PR too.

If it's independent of this change, I'd recommend making it another PR just to keep things clean.

dshemetov:
Noted. I'll do that.

_delphi_utils_python/tests/test_smooth.py
krivard: +1, @dshemetov ?

dshemetov: Good catch, added an assertion 👍

chinandrew (approved):
lgtm, only test failing is the one that #476 fixes.

krivard (approved):
lgtm

PR #439, opened 2020-11-05
Fix production params file as identified in #392

krivard:
#392 identified a change to the params file that I didn't realize needed to be committed to the ansible tree. This PR adds that change, as well as updating the `wip` signals so that we don't accidentally end up with the prototype 7dav signals in production.


korlaxxalrok (approved):
LGTM 👍

PR #443, opened 2020-11-05
Add Makefiles with convenience functions to each indicator

chinandrew:
Summary of changes:
Adds identical makefiles (except the utils, which only installs the utils) to each directory with a few convenience commands

Commands:
- `make install`: pip installs the utils package and the indicator in _editable_ mode in the venv, which means any changes you make to the utils package or your indicator will automatically propagate without having to reinstall
- `make test`: runs pytest on the entire test directory
- `make lint`: runs pylint on the code directory (through janky regex to accommodate dynamic folder names)
- `make venv`: creates a virtual environment called `env`. `install` calls this so you probably don't need to use it.
- `make clean`: deletes the virtual environment

NOTE: When you run the commands, they will all be run in your virtual environment but your shell itself won't be activated (i.e. it won't show `(env)` in front) because make commands run in their own shell that exits once the command completes. However, all the installations, tests, etc, are done in the venv, so you can activate it if needed (e.g. if you want to run individual tests) and will see everything installed.


_delphi_utils_python/Makefile
```diff
@@ -0,0 +1,22 @@
+.PHONY = venv, lint, test, clean
+
+dir = $(shell find ./delphi_* -name __init__.py |  sed 's/\/__init__.py//g;s/\.\///g')
+
+venv:
+	python3 -m venv env
```
krivard, line 6:
```suggestion
	python3.8 -m venv env
```

Necessary because anyone using CMU-standard Ubuntu has 3.6 installed, and thus:
```
$ python3 --version
Python 3.6.9
```

_delphi_utils_python/Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+lint:
+	. env/bin/activate; \
+	pylint $(dir)
+
+test:
+	. env/bin/activate ;\
+	(cd tests && ../env/bin/pytest --cov=$(dir) --cov-report=term-missing)
+
+clean:
```
krivard, line 21: Should `clean` also dump local `params.json` files?

_delphi_utils_python/Makefile
```diff
@@ -0,0 +1,22 @@
+.PHONY = venv, lint, test, clean
+
+dir = $(shell find ./delphi_* -name __init__.py |  sed 's/\/__init__.py//g;s/\.\///g')
```
krivard, line 3: This is 100% picky bash style nonsense, but slight preference for grep over sed, especially if the sed is going to require escapes:

```suggestion
dir = $(shell find ./delphi_* -name __init__.py | grep -o 'delphi_[_[:alnum:]]*')
```

_delphi_utils_python/Makefile
```diff
@@ -0,0 +1,22 @@
...
+lint:
+	. env/bin/activate; \
+	pylint $(dir)
+
+test:
+	. env/bin/activate ;\
+	(cd tests && ../env/bin/pytest --cov=$(dir) --cov-report=term-missing)
+
+clean:
+	rm -r env
```
krivard, line 22: (add EOF newline)

krivard (changes requested):
One important fix, one query, two optional nits

krivard:
Also fwiw this is practically a line-for-line match for the Makefile i've been selfishly using locally, except your `$(dir)` is better than mine, so 👍

_delphi_utils_python/Makefile
chinandrew: that's a good idea, though they are gitignored already. is there a possibility someone keeps them around for whatever reason?

_delphi_utils_python/Makefile
chinandrew: :+1:

_delphi_utils_python/Makefile
krivard: Scenario is this:

1. User runs `make clean install` to reset their local environment in response to major changes to indicator X
2. Tests fail and/or indicator can't run. `params.json.template` was updated, but local copy of `params.json` persisted after `clean`.

chinandrew: makes sense, added.

krivard (approved):
lgtm

PR #444, opened 2020-11-06
Remove dependency on cache files from GHT tests

### LINKED ISSUE #213: Remove dependency on cache files from GHT tests
As suggested by #211 -- unit tests should not depend on the live cache files. Tests that check caching behavior should use a static dataset stored in the `tests` directory.

jingjtang:
Fixes #213 

- Remove dependency on cache files from the unit tests
- Remove dependency on internet/ght-key from the unit tests
- Delete cache files 

The unit tests coverage is good for all of the scrips except for `pull_api.py` since a large proportion of the code in that script requires ght-key to pull data from the Google Health Trends API. Currently, didn't find a good way to test those functions.


google_health/tests/test_pull_api.py
```diff
@@ -1,129 +1,14 @@
...
 
 from delphi_google_health.pull_api import (
-    GoogleHealthTrends,
-    get_counts_states,
-    get_counts_dma,
-    _get_counts_geoid,
     _api_data_to_df,
     _load_cached_file,
     _write_cached_file,
 )
```
sgsmob, line 11: Rather than deleting these tests, I would prefer to:

1. Leave the classes and test names here as-is.
2. Comment out the parts of the test body that cause them to fail.  In the worst case, the non-commented part of the test body can just be `assert True`.
3. File an issue to fix these tests.
4. Leave a TODO comment on each of these tests with the number of the issue you just created.

Since we would like to test the pulling behavior, it would be good for future implementers of this feature to see what the previous intended behavior was.

sgsmob (changes requested):

google_health/tests/test_pull_api.py
```diff
@@ -18,110 +12,122 @@
...
-        ]
-
-        assert len(res["lines"][0]["points"]) == 2
-        assert set([x for x in res["lines"][0]["points"][0].keys()]) == {"date", "value"}
-        assert res["lines"][0]["points"][0]["date"] == "May 05 2020"
-        assert res["lines"][0]["points"][1]["date"] == "May 06 2020"
-
+     def test_class_state(self):
+         
+         # TODO
```
sgsmob, line 18: please add the Github issue here and to the other TODOs.

sgsmob (changes requested):

google_health/tests/test_pull_api.py
sgsmob: Sorry I wasn't clearer on this.  You should create a github issue called something like "Get the ght tests to work without authentication" and explain a bit more in the description.  Then, in the commented out tests you should leave a comment that says "TODO(#XXX):  reenable these tests" where XXX is the number of the issue you created.

sgsmob (approved):

PR #445, opened 2020-11-06
Use archive differ to construct USAFacts issues instead of 'latest'

krivard:
This PR corrects the production USAFacts settings for use with the archive differ.

USAFacts production params were left as "latest" which only runs 1 day of 7dav signals and 7 days of raw signals. With the archive differ in place, the correct setting is "20200201".


chinandrew (approved):
:shipit: :shipit: :shipit:

PR #447, opened 2020-11-06
Merge deploy-changehc to main

korlaxxalrok:
Update main with this new deployment branch.

NOTE: In addition to the standard additions for a deployment branch, we've also added:

- Jenkins staging deployments

After this branch is integrated with main (and then propagated to the other deployment branches), we'll start deploying an indicator package as part of the Jenkins pipeline that runs for the PR submission/update against the corresponding deployment branch. The Jenkins PR phase now looks like `build -> test -> package -> deploy to staging`. The subsequent merge deploys to production.


korlaxxalrok:
@krivard It may or may not be useful to merge this yet because of our possible upcoming rework of this repo's workflow, but here it is.

krivard (approved):
xlnt

krivard:
> It may or may not be useful to merge this yet because of our possible upcoming rework of this repo's workflow,

Good thought, but without knowing how long it will take to implement workflow changes, I'd rather keep the CHC indicator moving forward

Issue #448, opened 2020-11-06
Fix matrix inversion bug on smoother unit tests


chinandrew:
numerical instability leads to tests failing on certain platforms.

PR #449, opened 2020-11-06
Add CI for python

chinandrew:
Blocked by #443, which is currently merged into this branch since we use the make commands. The only actual change in this pr should be the workflow yaml

Summary of changes:
- Add workflows for all production indicators
- Google health is excluded since it is deprecated and also makes network calls in tests, though shouldn't be too much work if we want to spend the time to add it.
- Delphi utils is commented out until #448
- linting is commented out while I work on PR's to get them all to pass

There's a handful of small optimizations that could be made to the workflow, but I think things will slightly change anyway as we tune linting and see how various commands work.


krivard (approved):
lgtm

korlaxxalrok (approved):
Looks good 👍

PR #456, opened 2020-11-06
Update nchs_mortality pipeline with new signal names

jingjtang:
- Signal names were updated as required by Roni


krivard:
I think you collided with a linter update

jingjtang:
@krivard Ok, updated. Will make another PR for linter after this one is merged.

PR #431, opened 2020-11-04
Refactor CHC to use smoothing utils

chinandrew:
Closes #430 

Summary of Changes:
- Update Change to use new smoothing utils
- Update test to smooth on test data

PR #463, opened 2020-11-09
Fix linting on CHC

chinandrew:
Summary of changes:
- Fix linting issues in CHC. Mainly long lines and formatting log strings

Note that there are still linting issues after this PR that will be closed by #431 and #460

### DIFF
--- a/changehc/delphi_changehc/update_sensor.py
+++ b/changehc/delphi_changehc/update_sensor.py
@@ -6,17 +6,18 @@ Created: 2020-10-14
 # standard packages
 import logging
 from multiprocessing import Pool, cpu_count
-from delphi_utils import GeoMapper, S3ArchiveDiffer, read_params, add_prefix
 
 # third party
 import numpy as np
 import pandas as pd
+from delphi_utils import GeoMapper, read_params, add_prefix
+
 # first party
 from .config import Config, Constants
+from .constants import SIGNALS, SMOOTHED, SMOOTHED_ADJ, NA
 from .load_data import load_combined_data
 from .sensor import CHCSensor
 from .weekday import Weekday
-from .constants import SIGNALS, SMOOTHED, SMOOTHED_ADJ, NA
 
 
 def write_to_csv(output_dict, write_se, out_name, output_path="."):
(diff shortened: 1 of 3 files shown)


changehc/delphi_changehc/update_sensor.py
```diff
@@ -235,30 +242,4 @@ def update_sensor(self,
...
-        print(fails)
-
-        # Filter existing exports to exclude those that failed to archive
-        succ_common_diffs = {f: diff for f, diff in common_diffs.items() if f not in fails}
-        arch_diff.filter_exports(succ_common_diffs)
-
-        # Report failures: someone should probably look at them
-        for exported_file in fails:
-            print(f"Failed to archive '{exported_file}'")
-        '''
```
chinandrew, line 264: not sure if these were still needed

rumackaaron: I'm not sure how exactly we use S3 and archiving. @krivard?

krivard: Each day when the indicator runs, it produces some output. Much of that output may match what's already in the API. To conserve disk space and memory, we want to exclude those matching rows from the new issue to be uploaded to the API. The easiest way to do this is to compare the daily output with a set of CSV files matching the current (most recent issue for each day of data) contents of the API. These files are kept in a cache in a versioned S3 bucket. After the day's output is produced, the archive differ updates the cached files, and edits the files in receiving to only include lines with updated or new data.

krivard: Probably best not to remove those lines until we have a fix to automation ready to run the archive differ as a subsequent process.

chinandrew: Just to be clear these are currently commented out with the triple single quotes, is it worth leaving them in as reference?

krivard: oh right! Change is new, so I suppose it makes sense that it's not already running the archive differ. 

In that case let's begin as we mean to go on, and drop them.

rumackaaron (approved):

PR #465, opened 2020-11-09
Fix linting on delphi utils

chinandrew:
Summary of changes:
- Fix linting across utils. Math-y variable names were ignored.

One outstanding linting error not included in this PR, will be fixed by #460


_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -796,7 +796,7 @@ def convert_zip_to_hrr(self, data, zip_col="zip", hrr_col="hrr"):
         return data
 
     def convert_zip_to_msa(
-        self, data, zip_col="zip", msa_col="msa", date_col="date", count_cols=None
+        self, data, zip_col="zip", date_col="date", count_cols=None
```
krivard, line 799: What's going on here?

chinandrew: that variable is unused in that function.

krivard: Do we know if any indicators call it with that keyword?

chinandrew: Didn't find any uses when I searched, only saw use of the `zip_to_msa` function below it, which calls `convert_zip_to_msa`.

krivard (approved):
lgtm

PR #466, opened 2020-11-09
Fix errors in Sir-Complains-a-lot

krivard:
This PR includes two fixes that were necessary to make Sir-CAL transmit to Slack successfully:

* Actually fail when Slack exceptions are raised
* Break up complaints stream into 50-block sections

### DIFF
--- a/sir_complainsalot/delphi_sir_complainsalot/run.py
+++ b/sir_complainsalot/delphi_sir_complainsalot/run.py
@@ -33,7 +33,7 @@ def run_module():
         sys.exit(1)
 
 def split_complaints(complaints, n=49):
-    """Yield successive n-sized chunks from lst."""
+    """Yield successive n-sized chunks from complaints list."""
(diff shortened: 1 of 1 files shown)


sir_complainsalot/delphi_sir_complainsalot/run.py
```diff
@@ -32,24 +32,30 @@ def run_module():
 
         sys.exit(1)
 
-def report_complaints(complaints, params):
+def split_complaints(complaints, n=49):
+    """Yield successive n-sized chunks from lst."""
```
benjaminysmith, line 36: Typo lst->list?

benjaminysmith (approved):

sir_complainsalot/delphi_sir_complainsalot/run.py
krivard, line 36:
```suggestion
    """Yield successive n-sized chunks from complaints list."""
```

PR #468, opened 2020-11-09
Fix for automated file downloads

rumackaaron:


korlaxxalrok (approved):
LGTM 👍

This also seems to look ok in staging (but validation still needs to happen).

PR #476, opened 2020-11-10
Fix smoother to remove a non-invertible design matrix bug

### LINKED ISSUE #448: Fix matrix inversion bug on smoother unit tests

dshemetov:
The error arose in the case of imputation with fewer datapoints than polynomial degree. Fixes #448. I also modified the imputation to change the polynomial fit degree at that boundary edge case and to use quadratic fitting by default (this is because imputing with polynomial_fit_degree=0 does a really poor job, so shouldn't ever be used).

### DIFF
--- a/_delphi_utils_python/delphi_utils/smooth.py
+++ b/_delphi_utils_python/delphi_utils/smooth.py
@@ -142,7 +142,9 @@ class Smoother:  # pylint: disable=too-many-instance-attributes
         if smoother_name == "savgol":
             # The polynomial fitting is done on a past window of size window_length
             # including the current day value.
-            self.coeffs = self.savgol_coeffs(-self.window_length + 1, 0)
+            self.coeffs = self.savgol_coeffs(
+                -self.window_length + 1, 0, self.poly_fit_degree
+            )
         else:
             self.coeffs = None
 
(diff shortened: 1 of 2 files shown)


_delphi_utils_python/delphi_utils/smooth.py
```diff
@@ -262,28 +264,32 @@ def left_gauss_linear_smoother(self, signal):
...
         predicted_value: float
             The anticipated value that comes after the end of the signal based on a polynomial fit.
         """
-        coeffs = self.savgol_coeffs(-len(signal) + 1, 0)
+        coeffs = self.savgol_coeffs(-len(signal) + 1 + nr, nr, poly_fit_degree)
         predicted_value = signal @ coeffs
         return predicted_value
 
-    def savgol_coeffs(self, nl, nr):
+    def savgol_coeffs(self, nl, nr, poly_fit_degree):
```
chinandrew, line 292: add docstring for the new param.

_delphi_utils_python/delphi_utils/smooth.py
```diff
@@ -416,17 +415,24 @@ def savgol_impute(self, signal):
...
+                        signal_imputed[:ix], 2, -1
+                    )
+            # Away from the boundary, use savgol fitting on a fixed window
             else:
-                coeffs = self.savgol_coeffs(-self.window_length, -1)
-                signal_imputed[ix] = (
-                    signal_imputed[ix - self.window_length : ix] @ coeffs
+                signal_imputed[ix] = self.savgol_predict(
+                    signal_imputed[ix - self.window_length : ix],
+                    2,
```
chinandrew, line 435: How big of a difference is it using a quadratic polynomial vs the previous window_length degree polynomial (assuming im reading this right)?

dshemetov: Hm, now I think I should make a different PR and start a discussion there for this. This change is unrelated to the matrix bug.

_delphi_utils_python/delphi_utils/smooth.py
dshemetov: Done.

_delphi_utils_python/delphi_utils/smooth.py
dshemetov: Took this change out.

chinandrew: :+1:. the rest of the PR looks good. One thing that I like to do when fixing a bug is write out the test case that would cause the bug and code until it passes. Then just leave it in as a test case and you have some confidence that it's resolved.

dshemetov: Nice idea, I'll try that.

chinandrew: sounds good, will approve once a test is added

dshemetov: Added a test!

dshemetov: And a few other bad input tests for the sake of code coverage (couldn't help myself :P).

chinandrew (approved):
:shipit: :shipit: :shipit:

PR #477, opened 2020-11-10
Update smoother nan handling

dshemetov:
It seemed that there were many cases in JHU tests that required the smoother to be able to act on an array full of nans, so instead of throwing an error, I improved things to be more graceful.

Basically handles the following cases:
* the entire array is nans, in which case the array is returned
* the array begins left-padded with nans, in which case the nans are truncated, the remaining signal is smoothed, and the nans are appended back before being returned
* fixes a couple edge cases where the polynomial degree size or the window length is larger than the number of data points

Its base is #476, so should be merged after.

### DIFF
--- a/_delphi_utils_python/delphi_utils/smooth.py
+++ b/_delphi_utils_python/delphi_utils/smooth.py
@@ -138,6 +138,8 @@ class Smoother:  # pylint: disable=too-many-instance-attributes
             raise ValueError("Invalid impute_method given.")
         if self.boundary_method not in valid_boundary_methods:
             raise ValueError("Invalid boundary_method given.")
+        if self.window_length <= 1:
+            raise ValueError("Window length is too short.")
 
         if smoother_name == "savgol":
             # The polynomial fitting is done on a past window of size window_length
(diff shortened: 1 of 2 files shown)


_delphi_utils_python/delphi_utils/smooth.py
```diff
@@ -122,6 +122,8 @@ def __init__(
             raise ValueError("Invalid smoother name given.")
         if self.impute_method not in valid_impute_methods:
             raise ValueError("Invalid impute method given.")
+        if self.window_length <= 1:
+            raise ValueError("Window length is too short.")
```
chinandrew, line 126: add test for this error

dshemetov: Done.

_delphi_utils_python/delphi_utils/smooth.py
```diff
@@ -279,33 +299,39 @@ def left_gauss_linear_smoother(self, signal):
             signal_smoothed[signal_smoothed <= self.minval] = self.minval
         return signal_smoothed
 
-    def savgol_predict(self, signal):
+    def savgol_predict(self, signal, poly_fit_degree, nr):
         """Predict a single value using the savgol method.
-
+
```
chinandrew, line 304:
```suggestion

```

chinandrew (approved):
theres a whitespace hanging around, otherwise :shipit: :shipit: :shipit:

dshemetov:
fixed! do you know if there's a way to merge commits without having to reset and force push every time?

chinandrew:
> fixed! do you know if there's a way to merge commits without having to reset and force push every time?

If the commits were previously pushed to the remote branch, I think you'll have to force push after squashing.

dshemetov:
I hope github doesn't send emails every time I do that?

chinandrew:
> I hope github doesn't send emails every time I do that?

I get github emails for all new commits on every PR, so my signal to noise ratio is pretty bad already. I just use a filter for the ones that I'm reviewing or am mentioned in and it works alright.

dshemetov:
Oh no. I'm sorry for everyone's inboxes. Chains of dependent PRs: not even once.

PR #484, opened 2020-11-10
Add an test for error case in utils

chinandrew:
Noticed a simple gap in coverage that I figured I'd fix

Summary of Changes:
- Add a test that an invalid input raises a value error


krivard (approved):
lgtm

PR #488, opened 2020-11-10
Add string to int cleaning for USAFacts

### LINKED ISSUE #461: Make CSV processing more robust for USAFacts (or more generally?)
USAFacts added a value  in the count column which was `"1,020"` which lead to downstream issues since the column wasn't integers. We could add a preprocessing step (maybe a series of converters for pandas?) that we can apply to clean up these fields. Stripping commas seems to be the main suggestion for now. 

Even though this was a USAFacts issue, I think this could make sense to generalize into the utils package.

chinandrew:
### Description
Add a cleaning step to usafacts to prevent against the bug that happened last week where one of the count values was `"1,020"` instead of `1020`.

### Changelog
- Add a line that looks at any str's that were parsed, removes commas, and casts them to ints.
- Update test to check this behavior and also add a few more cases/values of interest (e.g. cruise ship fips). Previously, the test only checked that the column headers were correct. To write this new test, I ran the test with the new data on `main` and saved the output, and then used that output as the test case when I introduced this new change

### Fixes 
- Fixes #461


chinandrew:
ok seems like i broke some other tests, will fix

krivard (approved):
lgtm

(also <3 these checks, nice work)

Issue #452, opened 2020-11-06
Fix linting to get CI to pass


chinandrew:
The following directories need to be updated to pass pylint and enabled in CI:

- [x] utils - #460 + #465
- [x] changhc - #431 + #460 + #463
- [ ] claims hosp - #433 + #462
- [x] googlesymptoms #455
- [x] nchs mortality #454 
- [x] quidel - #563
- [x] usafacts - #429

some of these will be fixed by the smoothing refactor, since those modules are throwing errors

chinandrew:
pydocstyle open directories:
- [ ]  utils
- [ ] goog symptoms
- [ ] jhu
- [ ] nchs mortality
- [ ] quidel
- [ ] quidel covidtest

PR #489, opened 2020-11-10
Add linting to CI for subset of indicators

chinandrew:
### Description
Add linting for tests which currently pass linting. Decided that waiting for get all of them fixed in #452 could take a few days and may lead to other lint breaking changes getting merged in the meantime, so might as well enable the ones that work.

### Changelog
- Enable linting with a conditional statement to only run on the indicators which currently pass.
- Fixed a whitespace on google symptoms to get it to pass

### Fixes 
- n/a

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -35,8 +35,9 @@ jobs:
       run: |
         make install
     - name: Lint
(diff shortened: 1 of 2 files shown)


krivard (approved):
lgtm

PR #493, opened 2020-11-11
enable CHC linting onCI

chinandrew:
### Description
Enable linting CI step to occur for CHC now that it passes.

### Changelog
- remove CHC from the lint exclusion list

### Fixes 
- n/a


krivard (approved):
lgtm

Issue #497, opened 2020-11-11
Migrate all indicators to a common export function


sgsmob:
We have a lot of replication of export code across indicators.  It doesn't appear that the functionality is actually that different across indicators, so we should build out `delphi_utils.create_export_csv()` and migrate all the indicators to use it.

Indicators that still need to be migrated:

- [x] `google_health`
- [ ] `nchs_mortality`
- [x] `quidel`
- [x] `quidel_covidtest`
- [x] `safegraph` (#574)
- [x] `safegraph_patterns`

chinandrew:
safegraph_patterns looks like it used the original create_export_csv implementation, so checking it off

chinandrew:
`nchs_mortality` uses epiweeks instead of days, so punting for now

dshemetov:
Been diving into the `changehc` code recently. Seems like it could move to using this utility, at first glance. Would mostly need to pack the existing row data into a dataframe at the end of `update_sensor` and replace `write_to_csv` with a call to the export util. Thoughts @rumackaaron?

PR #501, opened 2020-11-11
Expand functionality of `delphi_utils.create_export_csv()` for use in more indicators

sgsmob:
### Description
This PR adds more features (see below) to `delphi_utils.create_export_csv()` such that it can replace the custom `export_csv()` functions used in multiple indicators (see #497).

### Changelog
- Make the `start_date` argument optional, in which case the oldest datapoints are allowed to be written.
- Allow an optional `end_date` argument that prevents newer data points from being written to file.
- Make the `metric` argument optional.

### Addresses
- #497


_delphi_utils_python/delphi_utils/export.py
```diff
@@ -2,16 +2,19 @@
...
 def create_export_csv(
     df: pd.DataFrame,
-    start_date: datetime,
     export_dir: str,
-    metric: str,
     geo_res: str,
     sensor: str,
+    metric: Optional[str] = None,
+    start_date: Optional[datetime] = None,
+    end_date: Optional[datetime] = None
```
krivard, line 17: just confirming for future forensics: i've done a search, and none of the indicators that currently call this function use positional arguments to specify `start_date` or `metric`, so the change in signature here isn't going to break anything

krivard (approved):
xlnt

PR #504, opened 2020-11-12
Pass dict params to archiver

### LINKED ISSUE #503: Archive differ run-as-module fails with AttributeError
**Actual Behavior:**

I am working on prototyping a multi-step indicator pipeline using the combo cases and deaths indicator. When I run the archive differ as a standalone module, it fails with an AttributeError, traceback below. It appears that the archive module main is improperly attempting to access dict keys as object attributes, since `cache_dir` is indeed present in the params read from disk:

```
$ env/bin/python -i -m delphi_utils.archive --archive_type s3 --indicator_prefix combo_cases_deaths
/usr/lib/python3.8/runpy.py:127: RuntimeWarning: 'delphi_utils.archive' found in sys.modules after import of package 'delphi_utils', but prior to execution of 'delphi_utils.archive'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/krivard/projects/covid/auto/covidcast-indicators/_delphi_utils_python/delphi_utils/archive.py", line 590, in <module>
    params.cache_dir,
AttributeError: 'dict' object has no attribute 'cache_dir'
>>> params
{'static_file_dir': './static', 'export_dir': './receiving', 'cache_dir': './cache', 'export_start_date': [2020, 4, 1], 'date_range': 'all', 'source': 'indicator-combination', 'wip_signal': [], 'aws_credentials': {'aws_access_key_id': '[elided]', 'aws_secret_access_key': '[elided]'}, 'bucket_name': 'delphi-covidcast-indicator-output-dev'}
```

**Expected behavior**

Either make `read_params()` return an Object, or make `archive.__main__` access params using dict-key notation.

sgsmob:
### Description
The archive module failed because we were accessing params as object members rather than as keyed values.  Changed to be the latter.

### Changelog
- `params.foo` has been fixed to `params["foo"]`

### Fixes 
- Fixes #503


_delphi_utils_python/delphi_utils/archive.py
```diff
@@ -588,8 +588,8 @@ def archive_exports(self, exported_files: Files) -> Tuple[Files, Files]:
     args = parser.parse_args()
     params = read_params()
     run_module(args.archive_type,
-               params.cache_dir,
-               params.export_dir,
+               params["cache_dir"],
+               params["export_dir"],
                aws_credentials=params.aws_credentials,
```
krivard, line 593: and this one

_delphi_utils_python/delphi_utils/archive.py
```diff
@@ -588,8 +588,8 @@ def archive_exports(self, exported_files: Files) -> Tuple[Files, Files]:
...
                branch_name=args.branch_name,
                bucket_name=params.bucket_name,
```
krivard, line 595: and this one too

krivard (changes requested):
Kinda wish there was a way to unit test these so we were less dependent on eyeball checks

sgsmob:
> Kinda wish there was a way to unit test these so we were less dependent on eyeball checks

I can add some sort of integration test here (probably?)

krivard (approved):

PR #509, opened 2020-11-12
Add a makefile for Sir Complains-a-lot

krivard:
### Description
Sir CAL is close enough to an indicator that a makefile for it is useful

### Changelog
- add Makefile


chinandrew (approved):
`make test` will fail because theres no tests, but should be fine to include

PR #514, opened 2020-11-13
Add licensing/DUA checks to signal release process

capnrefsmmat:
So we always check the DUA before releasing and apply any necessary restrictions. A process fix like this ensures we can't make this class of mistakes easily.

### Description
Update the issue template for new features.


krivard (approved):
xlnt

PR #515, opened 2020-11-13
Link directly to the data source template

capnrefsmmat:
To encourage its use by everyone adding a data source.

### Description

The template covers everything important (and I'll soon add licensing to it), so we should make sure people creating a new signal notice it and use it.


krivard (approved):
👍

PR #516, opened 2020-11-13
Migrate quidel to use delphi_utils.create_export_csv

sgsmob:
### Description
Migrate `quidel` to use the common export function rather than an indicator-specific one.

### Changelog
- `delphi_quidel.run()` calls `delphi_utils.create_export_csv()` instead of `delphi_quidel.export()`
- Delete `delphi_quidel.export()`

### Fixes 
- Partially addresses #497


chinandrew:
same question i left on #517, rest looks good

sgsmob:
> same question i left on #517, rest looks good

Should be fixed in the same way

chinandrew (approved):
:shipit::shipit::shipit:

PR #517, opened 2020-11-13
Migrate quidel_covidtest to delphi_utils.create_export_csv

sgsmob:
### Description
Migrate `quidel_covidtest` to use the common export function rather than an indicator-specific one.

### Changelog
- `delphi_quidel_covidtest.run()` calls `delphi_utils.create_export_csv()` instead of `delphi_quidel_covidtest.export()`
- Delete `delphi_quidel_covidtest.export()`

### Fixes 
- Partially addresses #497


quidel_covidtest/delphi_quidel_covidtest/export.py
```diff
@@ -1,39 +0,0 @@
...
-    df = df.copy()
-    df = df[np.logical_and(df["timestamp"] >= start_date,
-                           df["timestamp"] <= end_date)]
-
-    for date in df["timestamp"].unique():
-        t = pd.to_datetime(str(date))
-        date_short = t.strftime('%Y%m%d')
-        export_fn = f"{date_short}_{geo_name}_{sensor}.csv"
-        result_df = df[df["timestamp"] == date][["geo_id", "val", "se", "sample_size"]]
-        result_df = result_df[result_df["sample_size"].notnull()]
```
chinandrew, line 36: This line doesn't appear to be in the util, is it ok to remove?

sgsmob: hm not sure actually.  I was hoping you could refactor the null culling to be outside the call to `export_csv`, but that fails because it doesn't write (blank) files for dates with all null sample sizes.  Does this make a difference in practice?  When do we expect to have null values there?

sgsmob: @krivard may have the better context.

krivard: Sample sizes are null when any of the following are true:
* the signal is not based on sampling (as in cases/deaths)
* sampling is done by our data source, and is not exposed to us (as in ght/gs)
* sampling would reveal the market share of our data source, and as a courtesy to them we do not publish it (dv, ha, possibly quidel?)

chinandrew: Is not having blank files for certain dates a problem? Worst case we could write a blank file creator, since `to_csv()` will still work on an empty df.

sgsmob: Fixed.

chinandrew (approved):
:shipit::shipit::shipit:

PR #526, opened 2020-11-16
Fix issues causing `google_health` tests to fail

sgsmob:
### Description
`google_health`'s tests were failing on a clean install.  This PR hopefully fixes them to work.

### Changelog
- Only run archiving if we are not in test mode.  From what I can tell, the archiving deletes files in the receiving folder and thus doesn't let us compare them.  This is acceptable to do because:
    - Archive behavior was not being tested at all
    - Archiving will be called from its own module rather than an indicator so it will be deleted from this indicator anyway soon.
- Evaluates the existence of `wip_signal` properly.  Previous behavior interpreted any string value for the `wip_signal` param as not having a WIP signal; new behavior mimics the live behavior more closely and interprets non-null strings to count as having a WIP signal.

### DIFF
--- a/ansible/templates/google_health-params-prod.json.j2
+++ b/ansible/templates/google_health-params-prod.json.j2
@@ -11,7 +11,7 @@
     "aws_access_key_id": "{{ delphi_aws_access_key_id }}",
     "aws_secret_access_key": "{{ delphi_aws_secret_access_key }}"
   },
-  "bucket_name": "delphi-covidcast-indicator-output"
+  "bucket_name": "delphi-covidcast-indicator-output",
   "test": false,
   "test_data_dir": ""
 }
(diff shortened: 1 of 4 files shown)


krivard:
I notice that the following production params files do not include the "test" or "test_data_dir" keys; are those fixes in-scope for this PR? Otherwise tests will succeed in CI but the job will fail in staging/prod.

* ansible/templates/google_health-params-prod.json.j2
* ansible/templates/google_health-params-test.json.j2

sgsmob:
> I notice that the following production params files do not include the "test" or "test_data_dir" keys; are those fixes in-scope for this PR? Otherwise tests will succeed in CI but the job will fail in staging/prod.
> 
> * ansible/templates/google_health-params-prod.json.j2
> * ansible/templates/google_health-params-test.json.j2

Added.  Let me know if I didn't do that correctly.

ansible/templates/google_health-params-prod.json.j2
```diff
@@ -12,4 +12,6 @@
     "aws_secret_access_key": "{{ delphi_aws_secret_access_key }}"
   },
   "bucket_name": "delphi-covidcast-indicator-output"
```
krivard, line 14:
```suggestion
  "bucket_name": "delphi-covidcast-indicator-output",
```

krivard (approved):
glorious, thanks!

Issue #528, opened 2020-11-16
Fix the timestamp problem for NCHS


jingjtang:
**Actual Behavior:**

The current pipeline assume the col "start_week" to be the week-starting date. 

**Expected behavior**

But it is labeled as week-ending date. So this is one week shift in our reports.

**Context**

Fix including
- [x] code
- [x] DETAILS.md
- [x] The API documentation

jingjtang:
Fixed in #529

PR #529, opened 2020-11-16
Fix nchs timestamp

jingjtang:
### Description
bug fix，the start week used by NCHS is week-ending date not week starting date. So the previous one week shift is incorrect.

### Changelog
- Remove one week shift in `export.py`
- Update export related unit tests in `tests/test_export.py`
- Update `DETAILS.md` with epiweeks related definition and explanation. 

### Fixes 
- Fixes issue #528


krivard (approved):
lgtm

PR #531, opened 2020-11-16
Migrate google_health to use `delphi_utils.create_export_csv()`

sgsmob:
### Description
Migrate `google_health` to use the common export function rather than an indicator-specific one.

### Changelog
- `delphi_google_health.run()` calls `delphi_utils.create_export_csv()`
- Delete `delphi_google_health.export()`

### Fixes 
- Partially addresses #497


krivard (approved):
nice work

PR #533, opened 2020-11-17
USAFacts Prop Fix

### LINKED ISSUE #536: Fix doulbe/half prop problem in USAFacts this time
**Actual Behavior:**

Same bug as #325, in USAFacts this time

**Expected behavior**

State prop signals should incorporate 1x the state population (which requires ignoring the population of the megafips during county->state aggregation)

dshemetov:
### Description
Same half-correct prop signal value that JHU was having #328 

### Changelog
- Filtered out megafips when aggregating up to state, same fix as above

@sgsmob can you foresee any problems with this sort of fix here? Passed my sanity checks, but I'm not familiar with this indicator.

### Fixes
Fixes #536

### DIFF
--- a/usafacts/tests/test_geo.py
+++ b/usafacts/tests/test_geo.py
@@ -76,11 +76,11 @@ class TestGeoMap:
         """Tests that values are correctly aggregated at the state level."""
         df = pd.DataFrame(
             {
-                "fips": ["04001", "04003", "04009", "25023"],
-                "timestamp": ["2020-02-15", "2020-02-15", "2020-02-15", "2020-02-15"],
-                "new_counts": [10, 15, 2, 13],
-                "cumulative_counts": [100, 20, 45, 60],
-                "population": [100, 2100, 300, 25],
+                "fips": ["04001", "04003", "04009", "25023", "25000"],
+                "timestamp": ["2020-02-15", "2020-02-15", "2020-02-15", "2020-02-15", "2020-02-15"],
+                "new_counts": [10, 15, 2, 13, 0],
+                "cumulative_counts": [100, 20, 45, 60, 0],
+                "population": [100, 2100, 300, 25, 25],
             }
         )
 
--- a/usafacts/delphi_usafacts/geo.py
+++ b/usafacts/delphi_usafacts/geo.py
@@ -107,6 +107,13 @@ def geo_map(df: pd.DataFrame, geo_res: str, map_df: pd.DataFrame, sensor: str):
         df = df.append(unassigned_counties)
         geo_mapper = GeoMapper()
         df = geo_mapper.add_geocode(df, "fips", "state_id", new_col="geo_id")
+
+        # Zero out the state FIPS population to avoid double counting.
+        df = df.set_index("fips")
+        state_fips_codes = {str(x).zfill(2) + "000" for x in range(1, 73)}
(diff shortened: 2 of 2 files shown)


dshemetov:
Seems to work when comparing local (receiving files) and remote (API) (this is CA for example)
image

Edgecases?
* megafips counties should be fine, if USAfacts uses it, as the population zero only happens at the state aggregation level
* msa and hrr are fine, they don't use the geomapper

dshemetov:
Also a longterm thought: the way fips to state conversion works GeoMapper seems to be dangerous for prop signals. Maybe it would be best to update the function to filter out megafips by default.

capnrefsmmat:
Let me suggest that we also need a unit test here. This bug has already popped up twice in different indicators; if there's a way to test `geo_map` by passing in a fake data frame with a few different counties and ensuring the props are correct, we should definitely do it.

krivard:
> I don't think there are cases where we want to keep the megafips data when converting to state?

Can you be more specific? Because I'm pretty sure we only want to jettison megafips population -- numerator counts etc should be retained.

krivard:
> Let me suggest that we also need a unit test here.

Agree this PR should include a unit test to prevent regression in future. @dshemetov?

> This bug has already popped up twice in different indicators

...unfortunately I don't think it's possible to test the current behavior in a way that's robust to adding new indicators susceptible to the same problem. We'd have to shift prop computations into the utility to do that -- which might be a good idea regardless -- but is too big a task for a hotfix like this. I'll put it on the schedule for December at least, and see if there's anything we can rearrange to get a head start on it this month.

krivard:
Test is in; verified fails in main and succeeds in this PR.

sgsmob:
> > This bug has already popped up twice in different indicators
> 
> ...unfortunately I don't think it's possible to test the current behavior in a way that's robust to adding new indicators susceptible to the same problem. We'd have to shift prop computations into the utility to do that -- which might be a good idea regardless -- but is too big a task for a hotfix like this. I'll put it on the schedule for December at least, and see if there's anything we can rearrange to get a head start on it this month.

I don't understand why you can't test this.

usafacts/tests/test_geo.py
```diff
@@ -76,11 +76,11 @@ def test_state(self):
...
             {
-                "fips": ["04001", "04003", "04009", "25023"],
-                "timestamp": ["2020-02-15", "2020-02-15", "2020-02-15", "2020-02-15"],
-                "new_counts": [10, 15, 2, 13],
-                "cumulative_counts": [100, 20, 45, 60],
-                "population": [100, 2100, 300, 25],
+                "fips": ["04001", "04003", "04009", "25023", "25000"],
+                "timestamp": ["2020-02-15", "2020-02-15", "2020-02-15", "2020-02-15", "2020-02-15"],
+                "new_counts": [10, 15, 2, 13, 0],
+                "cumulative_counts": [100, 20, 45, 60, 0],
```
sgsmob, line 82: Are these always assumed to be 0?

sgsmob (approved):
Overall this looks pretty good to me.  I don't have a ton of USAFacts background to know if this is going to cause more issues, but if it worked in JHU, that seems pretty good.

usafacts/tests/test_geo.py
krivard: No, but it was the minimum change to the tests that failed in main and succeeded with the fix.

Issue #367, opened 2020-10-26
Release NCHS Mortality


krivard:
- [Link to issue](https://github.com/cmu-delphi/covidcast-indicators/issues/119)
- [Link to PR](https://github.com/cmu-delphi/covidcast-indicators/pull/233)
- Proposed release version: 1.11

This is a new indicator that's completed its first phase of development and is ready to consider for public release. This indicator tracks the number of covid, pneumonia, and all deaths, as well as the percentage of expected deaths, as published by the NCHS. Signal is per week, but the estimate is updated daily.

**Edit: we will not pursue map release at this time**

- [x] Statistical review (usually [correlations](https://github.com/cmu-delphi/covidcast/tree/main/docs/R-notebooks))
- [x] Signal / source name review (usually [Roni](https://docs.google.com/document/d/10hGd4Evce4lJ4VkWaQEKFQxvmw2P4xyYGtIAWF52Sf8/edit?usp=sharing))
- [x] Add to Delphi Automation
- [x] API support for weekly data

- [x] Visual review
- [ ] ~[Signal description pop-up text](https://docs.google.com/document/d/1kDqRg8EaI4WQXMaUUbbCGPlsUqEql8kgXCNt6AvMA9I/edit?usp=sharing) drafting and review~
- [ ] ~[Map release notes](https://docs.google.com/document/d/1BpxGgIma_Lkd2kxtwEo2DBdHQ3zk6dHRz-leUIRlOIA/edit?usp=sharing)~

- [x] API [documentation](https://github.com/cmu-delphi/delphi-epidata/tree/main/docs/api) and/or [changelog](https://github.com/cmu-delphi/delphi-epidata/blob/main/docs/api/covidcast_changelog.md)
- [ ] API mailing list notification

benjaminysmith:
Thanks Katie -- would you like me to shepherd the sub-tasks? Do you have an example of the process I can copy?

krivard:
I would! Some examples are in the links, but this is still a prototype of the process we'd like to build, so not everything has something to go off of.

The tasks are roughly in order -- eg if the statistical review doesn't go off, we don't want to move forward until that's been fixed.

[...]

PR #535, opened 2020-11-17
Fix nchs missing value

jingjtang:
### Description
bug fix. Do not assign 0 to missing values. The nchs mortality weekly table reports incidence. The existence of a missing value means the data does not meet the privacy threshold.

### Changelog
- pull.py and run.py
- tests/test_run.py, tests/test_pull.py (add tests for missing values)

### Fixes 
- Fixes problem mentioned [here](https://github.com/cmu-delphi/covidcast-indicators/issues/367#issuecomment-728288489)

### DIFF
--- a/nchs_mortality/delphi_nchs_mortality/pull.py
+++ b/nchs_mortality/delphi_nchs_mortality/pull.py
@@ -3,6 +3,7 @@
 import numpy as np
 import pandas as pd
 from sodapy import Socrata
+from .constants import METRICS
 
 def pull_nchs_mortality_data(token: str, map_df: pd.DataFrame, test_mode: str):
     """Pull the latest NCHS Mortality data, and conforms it into a dataset.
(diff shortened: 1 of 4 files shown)


nchs_mortality/delphi_nchs_mortality/pull.py
```diff
@@ -67,29 +66,6 @@ def pull_nchs_mortality_data(token: str, map_df: pd.DataFrame, test_mode: str):
     df = df[df["state"] != "United States"]
     df.loc[df["state"] == "New York City", "state"] = "New York"
 
-    state_list = df["state"].unique()
-    date_list = df["timestamp"].unique()
-    index_df = pd.MultiIndex.from_product(
-        [state_list, date_list], names=['state', 'timestamp']
-    )
-    df = df.groupby(
-            ["state", "timestamp"]).sum().reindex(index_df).reset_index()
```
eujing, line 76: It looks like this reindex step was used to fill in missing values by state and time and removing it makes sense given the discussion.
I am not super familiar with the actual NCHS data, but were there cases of duplicated (state, timestamp) where the sum aggregation should still be retained?

eujing (commented):
Overall looks good, except 1 possible issue to clarify.
But otherwise tests pass for me too!

eujing (approved):
Looks good to me!

eujing (approved):
After further discussion, looks good to me!

PR #560, opened 2020-11-19
Enable Pydocstyle Linting

chinandrew:
### Description
Enable pydocstyle to lint docstrings. Only done for those that currently pass, tracking the ones that need fixing in #452

### Changelog
- Add pydocstyle to all setup.py files so it gets installed
- Add pydocstyle line to makefile for those that pass so it runs on `make lint`

### Fixes 
- n/a


krivard (approved):
lgtm

PR #565, opened 2020-11-20
Add null removal feature to exporting

sgsmob:
### Description
Add the ability to only export data points whose sample size is non-null.

As part of migrating indicators to a common export function in #497, I discovered some of the indicators remove entries whose sample sizes are null (see discussion on #516 and #517).  Rather than hack it in in the indicator code, I have made it a native option of `delphi_utils.create_export_csv()`

### Changelog
- `delphi_utils.create_export_csv()` takes a `remove_null_samples` option (default `False`) that, when `True`:
    - Does not write CSV rows for data points whose sample size is null
    - Does write a file for a date even if all data points on that day have null entries (i.e., the file will be empty)


krivard (approved):
xlnt

PR #568, opened 2020-11-20
Sensorization

rumackaaron:
Performs sensorization for DV signal.

- For each time_value t:
  - Create a single linear mapping (f) from a global case rate to global DV, fit globally using the previous [j, k] days
  - For each location i:
    - Create linear mapping (g_i) from DV to cases, fit using just location i and the previous [j, k] days
    - At time t and for location i, report f(g_i(DV(t))).  In other words, pass DV(t) → g_i → f


rumackaaron:
Sorry, I meant to merge this into the `dv-package` branch. Is that still possible or do I need to create a new PR?

rumackaaron:
Tagging @krivard

krivard:
Is this still under consideration, or are we backing off of sensorization entirely?

rumackaaron:
We are trying to tweak the method of both the sensorization map and the inverse map, but I don't think we're backing off entirely.

krivard:
okay -- maybe convert to draft while you're tweaking, then it'll be easier to know when someone should review it

doctor_visits/delphi_doctor_visits/config.py
```diff
@@ -21,21 +21,24 @@ class Config:
...
     FLU1_COL = ["Flu1"]
     COUNT_COLS = CLI_COLS + FLU1_COL + ["Denominator"]
-    DATE_COL = "ServiceDate"
-    GEO_COL = "PatCountyFIPS"
-    AGE_COL = "PatAgeGroup"
-    HRR_COLS = ["Pat HRR Name", "Pat HRR ID"]
+    DATE_COL = "ServiceDate" #"servicedate"
+    GEO_COL = "PatCountyFIPS" #"patCountyFIPS"
+    AGE_COL = "PatAgeGroup" #"patAgeGroup"
+    HRR_COLS = ["Pat HRR Name", "Pat HRR ID"]#["patHRRname", "patHRRid"]
```
chinandrew, line 27: what are these comments for?

doctor_visits/delphi_doctor_visits/update_sensor.py
```diff
@@ -140,7 +147,7 @@ def update_sensor(
     params = Weekday.get_params(data) if weekday else None
 
     # handle explicitly if we need to use Jeffreys estimate for binomial proportions
-    jeffreys = True if se else False
+    jeffreys = se
```
chinandrew, line 150: if this var needed?

rumackaaron:
From Ryan:

> Update: my current view is that the best way to sensorize at time t is to use all available history for the sensor regression model.  Then use this fitted model to define a new sensor at time t, and also redefine the sensor at all times 1,...,t-1.  This will give us the best of both worlds: improved comparisons across geos, and stable comparisons across time.  It can be accomplished using the "as of" parameter in the COVIDcast API, as I explain in my comment here.  See also Addison's DAP.
> 
> Suggestion: Aaron first implements this "offline" to make sure that it truly works as expected. Then Aaron + engineering work to make this possible when stored a sensor in the API, through the "as of" parameter.  Doing this efficiently, without using up a ton of memory going forward, requires us to do some dynamic but simple calculations on the API side.

To save memory, Ryan suggests we store the indicator values for the non-sensorized signal (indexed by time, location, and issue) and the location-specific and time-invariant coefficients (indexed by location and issue) on the server, instead of storing the non-sensorized signal as well as the sensorized signal.

This is not currently supported, as I output the sensorized signal just like any other signal. Thoughts on whether this is worth implementing on the API side and on the indicator side?

doctor_visits/delphi_doctor_visits/config.py
rumackaaron: At some point, HSP changed the column names of the files they were sending us. I was testing the code on a recent drop (with lowercase column names). The code originally had the uppercase names, so I changed them back to pass the tests and to merge the PR.

doctor_visits/delphi_doctor_visits/update_sensor.py
rumackaaron: Yes, it's used in `sensor.py`

chinandrew: oh i meant if it's going to be  jeffreys = se`, why not just use the `se` variable and not create/assign it to the `jeffreys` varrable?

doctor_visits/delphi_doctor_visits/update_sensor.py
```diff
@@ -225,8 +232,78 @@ def update_sensor(
...
+                                     geo_type = geo.lower(),
+                                     start_day = pd.to_datetime(sensor_dates[0]),
+                                     end_day = pd.to_datetime(sensor_dates[-1]))
+        target_df = target_df[["geo_value","time_value","value"]]
+
+        geo_folder = "../_delphi_utils_python/delphi_utils/data"
+        fips_pop = pd.read_csv("%s/fips_pop.csv" % (geo_folder),\
+            dtype={"fips":str,"pop":int})
+        if geo.lower() == "state":
+            fips_state = pd.read_csv("%s/fips_state_table.csv" % (geo_folder),\
```
chinandrew, line 260: any reason geomapper util isnt used here?

doctor_visits/delphi_doctor_visits/update_sensor.py
rumackaaron: Yeah that's fair. I'll change it

mariajahja:
> This is not currently supported, as I output the sensorized signal just like any other signal. Thoughts on whether this is worth implementing on the API side and on the indicator side?

Just to make sure I understand -- the sensorized signal will be _in addition to_ the existing indicator, or will it replace it? I understand it may be less confusing on the map to only show the sensorized version, but we'd still like the API to provide the original indicator.

krivard:
> Thoughts on whether this is worth implementing on the API side and on the indicator side?

This approach sounds similar to/compatible with https://github.com/cmu-delphi/delphi-epidata/issues/239. In general I'm in favor of such an effort, but it's a much bigger lift than we can complete by the end of the month. Probably doable as a Q1 goal though. Who else should weigh in?

doctor_visits/delphi_doctor_visits/update_sensor.py
rumackaaron: Fixed

_delphi_utils_python/.pylintrc
```diff
@@ -1,8 +1,22 @@
-[DESIGN]
```
chinandrew, line 1: whats going on with the util files changing?

rumackaaron: In order to use GeoMapper, I had to "merge" the main `delphi_utils` folder into the current `sensorization` branch. This PR merges the `sensorization` branch into the `dv-package` branch, which is using a very old version of `delphi_utils`

krivard: > I had to "merge" the main delphi_utils folder

uh does this mean if we ever merge the dv-package branch into main, it'll generate an enormous pile of conflicts? ...wouldn't it be better to just `git merge main`?

rumackaaron: I didn't think it would since I used `git checkout main _delphi_utils_python` but my knowledge of git is very limited. I can do a `git merge main` instead

krivard: Thanks for clarifying! I wasn't familiar with that syntax, so just to be sure, I have double-checked by doing this:
```
git checkout sensorization
git checkout -b test_merge
git merge main
```
and there are indeed no conflicts. you're all set!

doctor_visits/delphi_doctor_visits/update_sensor.py
```diff
@@ -225,12 +262,92 @@ def update_sensor(
...
+            geo_weights = geomapper.replace_geocode(
+                fips_pop,"fips","hrr",from_col="fips",new_col="geo",date_col=None
+            )
+        elif geo.lower() == "msa":
+            geo_weights = geomapper.replace_geocode(
+                fips_pop,"fips","msa",from_col="fips",new_col="geo",date_col=None
+            )
+        elif geo.lower() == "county":
+            geo_weights = fips_pop.rename(columns={"fips":"geo"})
+        geo_weights = geo_weights.rename(columns={"population":"weight"})
```
chinandrew, line 304: this feels like something that could eventually live in the geomapper util to get populations for everything, but for now would recommend moving this into its own function that returns the weight DF.

doctor_visits/delphi_doctor_visits/sensorize.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+
+        w_mean = (df[data_col] * df[wt_col]).sum() / df[wt_col].sum()
+        w_var = (np.power(df[data_col]-w_mean,2) * df[wt_col]).sum()
+        w_var = w_var / df[wt_col].sum()
+        return (w_mean, np.sqrt(w_var))
+
+
+    @staticmethod
+    def sensorize(
```
chinandrew, line 100: There's a lot going on here -- I'd split it up into a few different methods that get called depending on the various conditions

PR #570, opened 2020-11-20
Get utils to pass pydocstyle

chinandrew:
### Description
Fix docs on utils so pydocstyle passes.

### Changelog
- Fix linting errors
- Enable pydocstyle to run in makefile so it gets run on CI

### Fixes 
- n/a

### DIFF
--- a/_delphi_utils_python/delphi_utils/archive.py
+++ b/_delphi_utils_python/delphi_utils/archive.py
@@ -100,7 +100,7 @@ def run_module(archive_type: str,
                cache_dir: str,
                export_dir: str,
                **kwargs):
-    """Build and runs an ArchiveDiffer.
+    """Build and run an ArchiveDiffer.
 
     Parameters
     ----------
--- a/_delphi_utils_python/delphi_utils/archive.py
+++ b/_delphi_utils_python/delphi_utils/archive.py
@@ -336,7 +336,7 @@ class S3ArchiveDiffer(ArchiveDiffer):
         self.indicator_prefix = indicator_prefix
 
     def update_cache(self):
-        """For making sure cache_dir is updated with all latest files from the S3 bucket."""
+        """Make sure cache_dir is updated with all latest files from the S3 bucket."""
         # List all indicator-related objects from S3
         archive_objects = self.bucket.objects.filter(
             Prefix=self.indicator_prefix).all()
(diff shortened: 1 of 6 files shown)


_delphi_utils_python/delphi_utils/archive.py
```diff
@@ -99,7 +100,7 @@ def run_module(archive_type: str,
                cache_dir: str,
                export_dir: str,
                **kwargs):
-    """Builds and runs an ArchiveDiffer.
+    """Build and runs an ArchiveDiffer.
```
krivard, line 103:
```suggestion
    """Build and run an ArchiveDiffer.
```

_delphi_utils_python/delphi_utils/archive.py
```diff
@@ -269,7 +273,7 @@ def filter_exports(self, common_diffs: FileDiffMap):
                 replace(diff_file, exported_file)
 
     def run(self):
-        """Runs the differ and archives the changed and new files."""
+        """Run the differ and archives the changed and new files."""
```
krivard, line 276:
```suggestion
        """Run the differ and archive the changed and new files."""
```

_delphi_utils_python/delphi_utils/archive.py
```diff
@@ -330,9 +336,7 @@ def __init__(
         self.indicator_prefix = indicator_prefix
 
     def update_cache(self):
-        """
-        For making sure cache_dir is updated with all latest files from the S3 bucket.
-        """
+        """For making sure cache_dir is updated with all latest files from the S3 bucket."""
```
krivard, line 339:
```suggestion
        """Make sure cache_dir is updated with all latest files from the S3 bucket."""
```

_delphi_utils_python/delphi_utils/archive.py
```diff
@@ -482,8 +490,9 @@ def archiving_branch(self):
 
     def update_cache(self):
         """
+        Check if cache_dir is clean: has everything nice committed if override_dirty=False.
```
krivard, line 493:
```suggestion
        Check if cache_dir is clean: has everything nicely committed if override_dirty=False.
```

krivard (changes requested):
compound verb phrases like these probably aren't caught by the linter but we should be consistent

chinandrew:
> compound verb phrases like these probably aren't caught by the linter but we should be consistent

Good catch, thanks

krivard (approved):
lgtm

PR #574, opened 2020-11-23
Refactor safegraph to use export util

chinandrew:
### Description
Migrate safegraph to use the common export function rather than an indicator-specific one. Tested by storing the test output from original code and diff'ing against the new code to make sure nothing changed.

### Changelog
- `process_window()` now calls `delphi_utils.export.create_export_csv()` instead of `to_csv()`
- intermediate `timestamp` column is added so the util function can be used.

### Fixes 
- Partially addresses #497

### DIFF
--- a/safegraph/delphi_safegraph/process.py
+++ b/safegraph/delphi_safegraph/process.py
@@ -6,7 +6,7 @@ import numpy as np
 import pandas as pd
 from delphi_utils.signal import add_prefix
 from delphi_utils.export import create_export_csv
-from delphi_utils import GeoMapper
+from delphi_utils.geomap import GeoMapper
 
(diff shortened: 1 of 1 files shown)


safegraph/delphi_safegraph/process.py
```diff
@@ -5,7 +5,7 @@
 import numpy as np
 import pandas as pd
 from delphi_utils.signal import add_prefix
-
+from delphi_utils.export import create_export_csv
 from delphi_utils import GeoMapper
```
sgsmob, line 9: This should either be

- `from delphi_utils.geo import GeoMapper` or
- all the `delphi_utils` imports on one line

sgsmob (approved):
Just the imports, and this is good to merge.

safegraph/delphi_safegraph/process.py
chinandrew: Updated

PR #575, opened 2020-11-23
Update run script to copy a specific range of files to the receiving directory

korlaxxalrok:
### Description
Adds shell script variables and one-liner to copy a specific range of generated
files to the receiving directory.

There is a current need to copy generated files that start on the Monday 8 weeks
in the past up until today (or the most recently generated). The run script has
been updated with a one-liner to handle this. It is not a beaut but seems to
work.

### Changelog
- Update run-safegraph_patterns.sh


jingjtang (approved):
lgtm

PR #578, opened 2020-11-24
Make validation class pass the linter

sgsmob:
### Description
Refactor validation module to be lint compliant and otherwise more readable.  This PR fails to address one lint issue with the number of member variables of `Validator`; I will address it in a follow-on PR (see #579).

### Changelog
- Refactor `validate()` into a short, high-level function
- Load all data in a separate routine rather than one at a time
- No longer check input parameters for type or value since the checks would always pass
- Add tests for functions outside of `validate.py`


nmdefries (approved):

sgsmob:
@nmdefries can you unapprove this?  I'm going to add some lint compliance fixes and haven't pushed them through.

validator/delphi_validator/errors.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, check_data_id, expression, message):
             pass a check, provide the date
             - message: str explaining why an error was raised
         """
+        super().__init__(message)
```
nmdefries, line 34: The super init changes the output of `Validator.exit()` so that only the error message is printed. This isn't a concern here since you have the incoming validator report PR; just wanted you to be aware that the goal is to include the  `check_data_id` (unique error-data combo identifier), expression, and message in error output.

nmdefries (commented):
Everything else looks good. Thanks for the improved organization and structure!

validator/delphi_validator/validate.py
```diff
@@ -953,22 +853,31 @@ def validate(self, export_dir):
             self.check_max_allowed_max_date(max_date, geo_type, signal_type)
 
             # Get relevant reference data from API dictionary.
-            geo_sig_api_df = all_api_df[(geo_type, signal_type)]
+            api_df_or_error = all_api_df[(geo_type, signal_type)]
 
-            if geo_sig_api_df is None:
+            if isinstance(api_df_or_error, APIDataFetchError):
+                self.increment_total_checks()
```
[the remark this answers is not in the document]
sgsmob: Good call, fixed.

validator/delphi_validator/errors.py
sgsmob: I thought I had fixed this by moving the call to `super().__init__()` to the top of the function, but you are correct that it squashes the full error message.  I have changed this to take all the arguments which produces the same output as before.

nmdefries (approved):

PR #586, opened 2020-12-01
Release nchs_mortality

korlaxxalrok:
### Description
Feature.

Adding the necessities for releasing the nchs_mortality indicator.

### Changelog
- Add new vault variable
- Add new vault variable pointer
- Add Ansible params template
- Use a prodution ingestion dir path


jingjtang (approved):
lgtm. But we may need to wait for another Monday to get the output files. This pipeline needs to run everyday and we  also check diffs everyday. But new data will only be ingested every Monday.

korlaxxalrok:
@jingjtang Ok, I'll merge so we can work on those specifics after this gets to staging.

PR #587, opened 2020-12-01
CLI signal for CHNG

rumackaaron:
Added code to download files and create indicator for covid-like illness from data provided by Change HC.


changehc/delphi_changehc/run.py
```diff
@@ -14,11 +14,12 @@
...
 
 # first party
-from .download_ftp_files import download
+from .download_ftp_files import download_covid, download_cli
+from .load_data import load_combined_data, load_cli_data
 from .update_sensor import CHCSensorUpdator
 
 
-def run_module():
+def run_module(): # pylint: disable=too-many-branches,too-many-statements
```
krivard, line 22: Not keen on this disable -- is there a reason not to just split this up into subroutines? wouldn't that be easier to test anyhow?

changehc/params.json.template
```diff
@@ -4,6 +4,10 @@
   "cache_dir": "./cache",
   "input_denom_file": null,
   "input_covid_file": null,
+  "input_flu_file": null,
+  "input_mixed_file": null,
+  "input_flu_like_file": null,
+  "input_covid_like_file": null,
```
krivard, line 10: This is borderline too many; prefer something like
```
"input": {
   "denom": null,
   "covid": null,
[...]
}
```

changehc/delphi_changehc/run.py
rumackaaron: Done

changehc/params.json.template
rumackaaron: Done

krivard (approved):
xlnt

PR #589, opened 2020-12-02
Introduce a "validation report" into the validation suite

sgsmob:
### Description
Introduce a "validation report" object to the validation suite.  This report handles the aggregation, printing, and interpretation of the errors raised by the various validation checks.

This not only separates the logic of aggregating the errors from the code that performs the checks but also should allow us to split out the validation checks into separate classes, using the report as an interface between them.

This includes changes introduced in #578 so it should not be reviewed until after that is merged.

### Changelog
- Add `delphi_validation.report.ValidationReport` to track the results of validation
- Migrate existing validation code to use `ValidationReport`


validator/delphi_validator/report.py
```diff
@@ -0,0 +1,99 @@
...
+        num_suppressed: int
+            Number of errors suppressed
+        total_checks: int
+            Number of validation checks performed
+        raised_errors: List[Exception]
+            Errors raised from validation failures
+        raised_warnings: List[Exception]
+            Warnings raised from validation execution
+        suppressed_errors: List[Exception]
+            Errors raised from validation failures not found in `self.errors_to_suppress`
```
chinandrew, line 28: Should this be unsuppressed errors? Since it's errors that are not in the ignore list.

validator/delphi_validator/report.py
```diff
@@ -0,0 +1,99 @@
...
+        error: Exception
+            Error raised in validation
+
+        Returns
+        -------
+        None
+        """
+        self.raised_warnings.append(warning)
+
+    def __str__(self):
```
chinandrew, line 79: Add a test that this formats to what we expect

validator/delphi_validator/report.py
sgsmob: Yes, done.

validator/delphi_validator/report.py
sgsmob: Added tests for the whole report class

chinandrew (approved):
:shipit::shipit::shipit:

PR #595, opened 2020-12-02
Timezone issue described in #593

### LINKED ISSUE #593: CHNG files are dropped to server at 4 Eastern
The timestamps on the server are in UTC, at 8pm or 9pm depending on daylight savings time. When Python looks at the timestamps, it automatically converts them to the server's timezone (4pm). The code currently assumes that files are dropped at 8pm server time, need to modify this to 4pm.

rumackaaron:
### Description
See #593. When we download the CHNG files, the timestamps are automatically converted into the code's timezone, in our case 4pm. We shouldn't run the indicator between 4-6pm Eastern just to be safe.

### Changelog
- run.py

### Fixes 
- Fixes #593

### DIFF
--- a/changehc/delphi_changehc/run.py
+++ b/changehc/delphi_changehc/run.py
@@ -85,8 +85,8 @@ def run_module():
     make_asserts(params)
 
     if params["drop_date"] is None:
-        # files are dropped about 8pm the day after the issue date
-        dropdate_dt = (datetime.now() - timedelta(days=1,hours=20))
+        # files are dropped about 4pm the day after the issue date
+        dropdate_dt = (datetime.now() - timedelta(days=1,hours=16))
         dropdate_dt = dropdate_dt.replace(hour=0,minute=0,second=0,microsecond=0)
     else:
         dropdate_dt = datetime.strptime(params["drop_date"], "%Y-%m-%d")


changehc/delphi_changehc/run.py
```diff
@@ -85,8 +85,8 @@ def run_module():
     make_asserts(params)
 
     if params["drop_date"] is None:
-        # files are dropped about 8pm the day after the issue date
-        dropdate_dt = (datetime.now() - timedelta(days=1,hours=20))
+        # files are dropped about 4pm the day after the issue date
+        dropdate_dt = (datetime.now() - timedelta(days=1,hours=16))
         dropdate_dt = dropdate_dt.replace(hour=0,minute=0,second=0,microsecond=0)
```
krivard, line 90: Can you say more about what this does? If they post more than an hour earlier or later, what will happen?

rumackaaron: If we run this script at 4:01 and they had posted the files already at 3, then the script will execute properly.

If they consistently post the files late, then I'm pretty sure the code will throw some sort of "File not found" error. It uses `dropdate_dt` to determine the filenames, so it will expect there to be files from yesterday, but the code can only download files from two days ago.

If CHNG posts the files at 5pm one day and at 4pm the next, and we run the code between 4pm and 5pm, the code will throw an assertion error because it found too many files dropped within the last 24 hours.

This code followed the outline of the HSP download scheme, but maybe a simpler solution is to compute the expected filename beforehand and check if it's been uploaded to their server yet. This way the errors would be more intelligible. For now, if we run the script after 8, everything should work properly.

krivard: > compute the expected filename beforehand

Yep, this sounds ideal going forward. I'll put this PR through as-is and make a[ tracking issue](https://github.com/cmu-delphi/covidcast-indicators/issues/600) for the refactor.

PR #597, opened 2020-12-03
Update params template

rumackaaron:
Fixes `KeyError: 'types'`


krivard:
Would you also fix the copy in `ansible/templates/changehc-params-prod.json.j2` please? That's the version that gets deployed in Automation.

rumackaaron:
Yes, I think I committed that as you typed that comment.

krivard (approved):
🎉

PR #598, opened 2020-12-03
Add diff bucket to params template

korlaxxalrok:
### Description
Fix.

Add missing diff bucket params.

### Changelog
- Updated nchs_mortality params template with diff bucket secrets.


jingjtang (approved):
lgtm

PR #604, opened 2020-12-04
Fix file permissions

korlaxxalrok:
### Description
Fix.

Give the `deploy` user control over the match list file.

### Changelog
- Modify the Ansible playbook to chown with the `deploy` user.


korlaxxalrok:
@krivard Another iteration. Need to adjust some file perms on the proxy.

krivard (approved):
legit

Issue #199, opened 2020-08-10
Consider supporting a national geo level for all indicators


krivard:
This would be useful for generating nationwide time series plots, and also for completing the nesting doll of scales for viz.

Easiest way to do this is probably just to add it in the geo utility for all the python indicators, and roll out national level as we convert indicators to use the package. The fb-package branch already has an implementation of it for R.

dshemetov:
State to HHS region is available in the new geocode refactor.

For National, what are the expected input geocode levels? Do you expect
there to be non-US input codes?

RoniRos:
HHS Regions: awesome!

National: So far we have been focusing on US locations only.  I expect this will continue to be the case for the next ~6 months or so.   At some point, we may well want to expand internationally.  So we should not put any effort into creating codes for other countries, but at the same time we shouldn't do anything that will make it harder to expand internationally later.

krivard:
National should use the standard two-character country codes (https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)

see also https://github.com/cmu-delphi/delphi-epidata/pull/207

dshemetov:
I am not familiar enough with other countries' county-level geocodes to say how the country code inference procedure should work in general. 

In the meantime, in #217 I have implemented a very simple aggregation to the _"us" national level only_ as a stopgap, which works by summing records with a FIPS code.

Extending to international locations may require larger changes, such as keeping the national ISO code in a separate column from the finer geocode (like [JHU does](https://raw.githubusercontent.com/CSSEGISandData/COVID-19/master/csse_covid_19_data/UID_ISO_FIPS_LookUp_Table.csv), for example).

RoniRos:
Many countries have other types of divisions, e.g. provinces, cantons, etc.  I wouldn't worry about it now.

@dshemetov When you discuss the simple aggregation by "summing records" above, do you mean aggregation of sample elements?  Or weighted averaging of signal values?  The important thing is to properly account for all FIPS codes in the country, including those for which an estimate wasn't produced, a sample was too small (or even empty), etc.   I assume this was already done in aggregating counties to states.  Why not create the national signal directly from the states' signal?  Or better yet, from the HHS Regions signals?  These are clean hierarchies: every county belongs to exactly one state or territory, every state and territory belong to exactly one HHS Region, and the 10 HHS Regions comprise the national territory.

dshemetov:
@RoniRos Sounds good, simple for now.

[...]

PR #609, opened 2020-12-08
Add HHS mapping files

chinandrew:
### Description
First part of adding HHS support to geomapper. This just adds the mapping files, does not add to the actual geomapper util

### Changelog
- Add functions for deriving mapping files
- Add mapping csv files

### Fixes 
- Addresses some items discussed in #199


_delphi_utils_python/data_proc/geomap/geo_data_proc.py
```diff
@@ -531,6 +533,56 @@ def derive_zip_to_state_code():
...
+        create_state_hhs_crosswalk()
+    if not isfile(join(OUTPUT_DIR, FIPS_POPULATION_OUT_FILENAME)):
+        create_fips_population_table()
+
+    fips_pop = pd.read_csv(
+        join(OUTPUT_DIR, FIPS_POPULATION_OUT_FILENAME), dtype={"fips": str, "pop": int}
+    )
+    state_hhs = pd.read_csv(
+        join(OUTPUT_DIR, STATE_HHS_OUT_FILENAME),
+        dtype={"state_code": str, "state_id": str, "state_name": str},
```
nmdefries, line 550: Looks like the state -> HHS file should have column names `state_code` and `hhs_region_number`. Same problem on line 576.

_delphi_utils_python/data_proc/geomap/geo_data_proc.py
```diff
@@ -531,6 +533,56 @@ def derive_zip_to_state_code():
...
+        dtype={"zip": str, "pop": int, "state_code": str}
+    )
+    state_hhs = pd.read_csv(
+        join(OUTPUT_DIR, STATE_HHS_OUT_FILENAME),
+        dtype={"state_code": str, "state_id": str, "state_name": str},
+    )
+
+    (
+        zip_state.merge(state_hhs, on="state_code", how="left")
+        .drop(columns=["state_code", "state_id", "state_name"])
```
nmdefries, line 581: Update to reflect changed column names on line 576. Should be the same as line 556.

nmdefries (changes requested):
Have you considered leftpadding the HHS region numbers with zero so that they're all the same length? It can make string processing and sorting more convenient.

_delphi_utils_python/data_proc/geomap/geo_data_proc.py
chinandrew: > Have you considered leftpadding the HHS region numbers with zero so that they're all the same length? It can make string processing and sorting more convenient.

I went with the existing implementation (state -> hhs) that had them without the zero padding, which also follows the existing HRR convention of not left padding. I think it's to distinguish geos that are defined with zero pad (e.g. a zip code "06001" is defined as 5 digits and "6001" wouldn't be right, same for a county FIPS like  "01009") from those without (e.g. [HHS Regions](https://www.hhs.gov/about/agencies/iea/regional-offices/index.html) are just "Region 1."  I don't feel super strongly either way, though if we left padded HHS we should also probably do for HRR and update any downstream documentation accordingly.

chinandrew:
> Have you considered leftpadding the HHS region numbers with zero so that they're all the same length? It can make string processing and sorting more convenient.

I went with the existing implementation (state -> hhs) that had them without the zero padding, which also follows the existing HRR convention of not left padding. I think it's to distinguish geos that are defined with zero pad (e.g. a zip code "06001" is defined as 5 digits and "6001" wouldn't be right, same for a county FIPS like  "01009") from those without (e.g. [HHS Regions](https://www.hhs.gov/about/agencies/iea/regional-offices/index.html) are just "Region 1."  I don't feel super strongly either way, though if we left padded HHS we should also probably do for HRR and update any downstream documentation accordingly.

_delphi_utils_python/data_proc/geomap/geo_data_proc.py
chinandrew: In this case, the zip_state file is the one introducing these columns. The other function uses the fips population file (was just copying an different fips -> * example) which only has the population column to drop

nmdefries: Got it.

nmdefries (approved):

PR #612, opened 2020-12-09
Add structured logging to SirCAL 

benjaminysmith:
### Description
Add structured logging to SirCAL 

### Changelog
- Include missing export in delphi utils
- Replace print statements with logging calls in SirCAL

### Fixes 
First indicator for #606


sir_complainsalot/delphi_sir_complainsalot/run.py
```diff
@@ -5,6 +5,8 @@
 when the module is run with `python -m delphi_sir_complainsalot`.
 """
 
+import logging
+import structlog
```
chinandrew, line 9: why is this import needed?

chinandrew (approved):
small question on import, otherwise :shipit:shipit:shipit

PR #621, opened 2020-12-09
Create symptom surveys contingency table tool

### LINKED ISSUE #129: Build a contingency table framework for reporting more detailed symptom survey aggregates
I think this is a task for August or later -- once we have the revised survey deployed and have extracted interesting univariate aggregates from it for the map.

Many researchers will be interested in using our survey data for purposes that do not require the full individual response data. To make the data useful to the widest possible audience, we could release data files containing aggregate contingency tables; the files would be suitable for public release and would let researchers study associations between multiple questions.

I'm imagining something like this:

1. A script reads the individual response data and codes column names and responses with more descriptive names.
2. Sub-scripts, one per output file, take this cleaned data and essentially do `group_by(stuff) %>% summarize(stuff)`. For example, you could imagine a script doing `group_by(epiweek, state, age, race) %>% summarize(mean(tested_positive), mean(cli), n(), ...)`. Each sub-script can focus on a different combination of variables of interest.
3. Each of the sub-scripts generates an output table that's written to CSV, with rows with inadequate sample size omitted.
4. We serve these CSVs in some canonical public location and also keep an archive of historical ones.

I'm not sure where we can serve the CSVs most easily -- perhaps a dedicated Git repo? or is there a common data repository service that does this kind of thing? -- but I think, in principle, this is fairly easy to do.

Also, is there a web-based data explorer tool that works well with this kind of data, so the general public could examine it?

nmdefries:
### Description
Adds functions and pipeline to the delphiFacebook package to create contingency tables from the symptom surveys data. This will allow us to share the symptom surveys data at a finer level of aggregation than is available through COVIDcast without users needing to access individual response data.

Allows user to set desired aggregations, composed of a set of grouping variables, a response column of interest, time period of interest (month or epiweek), and an aggregation function of interest (mean, percentage, or count depending on question type). Performs the desired aggregations on cleaned data, censoring rows with insufficient sample size and saving the output to CSV, one for each different set of grouping variables and time period.

Closes #129 

### Changelog
- Adds contingency table-specific functions and some general utility functions
- Modifies a handful of existing functions to make them more generic
- Adds script (outside of package) to run contingency table pipeline
- Creates new contingency-specific settings in existing params file


facebook/delphiFacebook/R/contingency_tables.R
```diff
@@ -0,0 +1,1068 @@
...
+  
+  input_data <- load_responses_all(params)
+  input_data <- filter_responses(input_data, params)
+  msg_df("response input data", input_data)
+  
+  input_data <- merge_responses(input_data, archive)
+  data_agg <- create_data_for_aggregatation(input_data)
+  
+  data_agg <- filter_data_for_aggregatation(data_agg, params, lead_days = 12)
+  data_agg <- join_weights(data_agg, params, weights = "step1")
```
capnrefsmmat, line 114: Note to self: I'm checking with Facebook on whether it makes more sense to use the full weights here instead of step1, since we're using responses from all over the survey to make the contingency tables

facebook/delphiFacebook/R/contingency_tables.R
```diff
@@ -0,0 +1,1068 @@
...
+  }
+  
+  if ( "cids" %in% params$output )
+  {
+    write_cid(data_agg, "part_a", params)
+  }
+  if ( "archive" %in% params$output )
+  {
+    update_archive(input_data, archive, params)
+  }
```
capnrefsmmat, line 137: If the contingency table generator only runs periodically, I'd assume that the daily `run.R` will handle all CID and archive generation, and this script doesn't need to do separate work on the archive or CID lists. @krivard can confirm, but I think the CID generation is pretty tightly integrated into her shell pipeline for producing weights and aggregates and won't need to also use this script

facebook/delphiFacebook/R/contingency_tables.R
```diff
@@ -0,0 +1,1068 @@
...
+    )
+  }
+  
+  return(input_data)
+}
+
+
+#' Checks user-set aggregations for basic validity
+#'
+#' @param params Named list of configuration parameters.
```
capnrefsmmat, line 309: also document `aggs`?

facebook/delphiFacebook/R/contingency_tables.R
```diff
@@ -0,0 +1,1068 @@
...
+  # to individual files
+  for (group_ind in seq_along(agg_groups$group_by)) {
+    
+    agg_group = agg_groups$group_by[group_ind][[1]]
+    geo_level = agg_groups$geo_level[group_ind]
+    geo_crosswalk = cw_list[[geo_level]]
+    
+    these_aggs = aggregations[mapply(aggregations$group_by,
+                                     FUN=function(x) {setequal(x, agg_group)
+                                     }) & aggregations$geo_level == geo_level, ]
```
capnrefsmmat, line 459: Comment on these lines? Hard to see what they're doing

facebook/delphiFacebook/R/contingency_tables.R
```diff
@@ -0,0 +1,1068 @@
...
+end_of_prev_full_week <- function(date) {
+  if (ceiling_date(date, "week") == date) {
+    return(date)
+  }
+  
+  return(floor_date(date, "week") - days(1))
+}
+
+
+#### TODO: should be epiweeks eventually. Already exists a package to calculate?
```
capnrefsmmat, line 1007: The MMWRweek package works but is clunky; lubridate [supports epiweeks](https://lubridate.tidyverse.org/reference/week.html) too

facebook/delphiFacebook/R/contingency_tables.R
```diff
@@ -0,0 +1,1068 @@
...
+#' @param date Date that output will be calculated relative to
+#' 
+#' @return Date
+#' 
+#' @importFrom lubridate floor_date
+#' 
+#' @export
+start_of_prev_full_month <- function(date) {
+  return(floor_date(date, "month") - months(1))
+}
```
capnrefsmmat, line 935: Maybe these date util functions can go into a `dates.R` or something? Just to keep the files manageable in size

facebook/delphiFacebook/R/contingency_tables.R
```diff
@@ -0,0 +1,1068 @@
...
+    "mc_state" = "A3b",
+    "mc_mask_often" = "C14",
+    "mc_anxiety" = "C8_1",
+    "mc_depression" = "C8_2",
+    "mc_isolation" = "C8_3",
+    "mc_worried_family_ill" = "C9",
+    "mc_financial_worry" = "C15",
+    "mc_gender" = "D1",
+    "mc_age" = "D2",
+    "mc_race" = "D7",
```
capnrefsmmat, line 239: D7 actually allows the respondent to select multiple races. However, @dshemetov has found that in most cases, multirace combinations are rare and would not meet the privacy threshold. How would you like to handle that here? They could be coded into a single "multirace" group, for example, instead of including all of the powerset of possible race combinations

facebook/delphiFacebook/R/contingency_tables.R
```diff
@@ -0,0 +1,1068 @@
...
+    "mc_occupational_group_other" = "Q80",
+    "mc_cough_mucus" = "B4",
+    "mc_tested_current_illness" = "B5",
+    "mc_hospital" = "B6",
+    "mc_social_avoidance" = "C7",
+    "mc_financial_threat" = "Q36",
+    "mc_cmnty_mask_prevalence" = "C16",
+    "mc_flu_shot_jun2020" = "C17",
+    "mc_children_grade" = "E1",
+    "mc_children_school" = "E2",
```
capnrefsmmat, line 266: I don't think E1 and E2 are coded this way. See 1970207822. There's one response per row in the matrix of choices, and each of those is multiple choice, so you'd separately have `E1_1`, `E1_2`, etc.

facebook/delphiFacebook/R/contingency_tables.R
```diff
@@ -0,0 +1,1068 @@
...
+    # Either number ("n"; can be averaged although may need processing) or text ("t")
+    "n_hh_num_sick" = "hh_number_sick", # A2
+    "n_hh_num_children" = "A5_1",
+    "n_hh_num_adults" = "A5_2",
+    "n_hh_num_seniors" = "A5_3",
+    # "t_zipcode" = "A3", # zip5
+    "n_cmnty_num_sick" = "A4",
+    "t_symptoms_other" = "B2_14_TEXT",
+    "t_unusual_symptoms_other" = "B2c_14_TEXT",
+    "t_gender_other" = "D1_4_TEXT",
```
capnrefsmmat, line 181: We likely won't produce any aggregates that use text responses, since users may enter identifiable information in these fields

capnrefsmmat (commented):
Some initial review comments. This is a pretty big batch of code and I haven't examined it all in detail yet. Overall I like how it looks, and I'll have to experiment with it some.

Question: How hard would it be to add some end-to-end tests? What makes me nervous is that there's a lot of data-processing logic here, and if we have some subtle bug that makes us miscalculate aggregates, we won't notice. Can we make a toy response file with various combinations of categories and verify that the resulting aggregate tables are correct?

facebook/delphiFacebook/R/contingency_tables.R
capnrefsmmat: Facebook agrees that switching to the full weights makes sense

facebook/delphiFacebook/R/contingency_tables.R
krivard: Correct. The daily pipeline run uses the following output settings in `params.json`:

```
  "output": [
    "cids",
    "individual",
    "covidalert",
    "archive",
    "community"
  ],
```

facebook/delphiFacebook/R/contingency_tables.R
krivard: (driveby; ignore if nonsensical) could we treat D7 the way we do zip codes that cross county lines; ie distribute the response across all respondent races weighted by 1/length(D7)?

facebook/delphiFacebook/R/contingency_tables.R
nmdefries: I provided descriptive names for all existing columns for completeness, regardless of intent to use columns in aggregations. Would we prefer to drop or just not rename those we don't expect to be used?

facebook/delphiFacebook/R/contingency_tables.R
nmdefries: > D7 actually allows the respondent to select multiple races

Thanks for the correction!

There's also the possibility of moving multirace responses to the "other" category and listing the response codes as the "other" text. I'll look into how frequently this type of response occurs – it's possible they're rare enough that even a "multirace" category would always be censored due to low sample size.

capnrefsmmat: You  can poke @chinandrew or @dshemetov about the counts -- they already  did some analysis of the frequency of different options while considering differential privacy.

chinandrew: From old slacks between me/alex/dmitry, these were the results looking at a file in early Nov:
```
250k responses
~65k nans 
~180k responded with only 1 race
~5k responded with >1 race
```

facebook/delphiFacebook/R/utils.R
```diff
@@ -148,7 +149,7 @@ megacounty <- function(
   df_megacounties <- mutate(df_megacounties,
                             geo_id = make_megacounty_fips(.data$geo_id))
 
-  df_megacounties <- group_by(df_megacounties, .data$day, .data$geo_id)
+  df_megacounties <- group_by_at(df_megacounties, groupby_vars)
```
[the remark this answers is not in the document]
nmdefries: > Are you using it just for the ability to take a character vector of column names, instead of needing symbols?

Yes. I'll look into alternatives. I believe I'm using another `*_at` function elsewhere also.

facebook/delphiFacebook/R/contingency_tables.R
nmdefries: Yes! I initially grouped the functions for convenience and haven't thought too much about splitting them up sensibly. Since a lot of them are derivatives of existing FB package functions, I could use the same file structure but prefix with `contingency_tables_*`, so `contingency_tables_aggregate.R`, etc. Do you have a particular preference here?

capnrefsmmat: I'd prefix with `contingency` rather than `contingency_tables` (just easier to deal with), but that sounds good. I think these date functions aren't contingency-specific, though, even though the tables are their only current use case.

facebook/delphiFacebook/R/utils.R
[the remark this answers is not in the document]
nmdefries: I ended up using `group_by(across(<character vector>))` so existing calls to `megacounty()` don't need to be modified.

It looks like `write_data_api()` is also using a superseded `*_at` function [here](https://github.com/cmu-delphi/covidcast-indicators/blob/77274a1ccc1c7fcd5950313242d40ee7ebae23ce/facebook/delphiFacebook/R/to_api.R#L28). I don't see a straightforward way to change this to `mutate`, which doesn't seem to support functional modification of a set of columns.

facebook/delphiFacebook/R/contingency_run.R
```diff
@@ -0,0 +1,121 @@
...
+#' 
+#' @export
+run_contingency_tables <- function(params, aggregations)
+{
+  if (!is.null(params$end_date) & !is.null(params$n_periods)) {
+    ## Produce historical CSVs
+    
+    if (params$aggregate_range == "week") {
+      period_step <- days(7)
+    } else if (params$aggregate_range == "week") {
```
capnrefsmmat, line 31: shouldn't one of these be `month`? and do you need to add `months` to the `@importFrom` above?

facebook/delphiFacebook/R/contingency_run.R
```diff
@@ -0,0 +1,121 @@
+#' Wrapper that runs `run_contingency_tables_one_period` over several time periods
+#'
+#' Allows pipeline to be create a series of CSVs for a range of dates.
+#'
+#' See the README.md file in the source directory for more information about how to run
+#' this function.
```
capnrefsmmat, line 6: I don't see any changes to `README.md` in this PR, so is this comment true?

facebook/delphiFacebook/R/contingency_run.R
```diff
@@ -0,0 +1,121 @@
...
+    run_contingency_tables_one_period(params, aggregations)
+  }
+  
+
+}
+
+#' Run the contingency table production pipeline for a single time period
+#'
+#' See the README.md file in the source directory for more information about how to run
+#' this function.
```
capnrefsmmat, line 56: same here -- is this true?

facebook/delphiFacebook/R/contingency_run.R
```diff
@@ -0,0 +1,121 @@
...
+  aggregations <- verify_aggs(aggregations)
+  
+  cw_list <- produce_crosswalk_list(params$static_dir)
+  archive <- load_archive(params)
+  msg_df("archive data loaded", archive$input_data)
+  
+  input_data <- load_responses_all(params)
+  input_data <- filter_responses(input_data, params)
+  input_data <- bodge_v4_translation(input_data)
+  msg_df("response input data", input_data)
```
capnrefsmmat, line 86: I wonder if we should move this ingestion sequence into a function so that if we add additional steps, that can be done in one place

(and maybe we should factor out `bodge_v4_translation` to happen in one of the other functions? it's only in `create_complete_responses` because none of the aggregates need it, but if the contingency tables will need V4, may as well put it in `load_responses_all` or something)

facebook/delphiFacebook/R/contingency_run.R
```diff
@@ -0,0 +1,121 @@
...
+      params$parallel <- FALSE
+    } else {
+      options(mc.cores = cores)
+    }
+  }
+  
+  data_agg <- make_human_readable(data_agg)
+  
+  if (nrow(aggregations) > 0) {
+    aggregate_aggs(data_agg, aggregations, cw_list, params)
```
capnrefsmmat, line 118: I feel like the `if (nrow(aggregations) > 0)` check could be moved way up, since most of the work in this function is not needed if there are no aggregations

facebook/delphiFacebook/R/contingency_utils.R
```diff
@@ -0,0 +1,153 @@
...
+#' @return Character vector of filenames
+#' 
+#' @importFrom lubridate as_date days
+#' 
+#' @export
+get_filenames_in_range <- function(start_date, end_date, params) {
+  start_date <- as_date(start_date) - days(params$backfill_days)
+  end_date <- as_date(end_date)
+  date_pattern <- "^[0-9]{4}-[0-9]{2}-[0-9]{2}.*[.]csv$"
+  youtube_pattern <- ".*YouTube[.]csv$"
```
capnrefsmmat, line 64: The YouTube surveys are not used in any of the current pipelines; I don't think we need to support them here

facebook/delphiFacebook/R/contingency_write.R
```diff
@@ -0,0 +1,47 @@
...
+#' 
+#' @export
+write_contingency_tables <- function(data, params, geo_level, groupby_vars)
+{
+  if (!is.null(data) && nrow(data) != 0) {
+    data <- arrange(data, across(groupby_vars))
+    
+    # Format reported columns.
+    data <- mutate_at(data, vars(-c(groupby_vars)), 
+                      function(x) formatC(x,digits=7,format="f",drop0trailing=TRUE))
```
capnrefsmmat, line 26: spacing after commas in argument lists

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
...
+get_range_prev_full_period <- function(date = Sys.Date(), weekly_or_monthly_flag) {
+  if (weekly_or_monthly_flag == "month") {
+    # Get start and end of previous full month.
+    date_period_range = get_range_prev_full_month(date)
+  } else if (weekly_or_monthly_flag == "week") {
+    # Get start and end of previous full epiweek.
+    date_period_range = get_range_prev_full_week(date)
+  } else {
+    stop("invalid time period selection; must indicate either 'month' or 'week'")
+  }
```
capnrefsmmat, line 125:
```suggestion
get_range_prev_full_period <- function(date = Sys.Date(), weekly_or_monthly_flag = c("month", "week")) {
  weekly_or_monthly_flag <- match.arg(weekly_or_monthly_flag)

  if (weekly_or_monthly_flag == "month") {
    # Get start and end of previous full month.
    date_period_range = get_range_prev_full_month(date)
  } else if (weekly_or_monthly_flag == "week") {
    # Get start and end of previous full epiweek.
    date_period_range = get_range_prev_full_week(date)
  }
```

`match.arg` is the idiomatic way to handle enumerated arguments like this, and automatically raises an error when the arg doesn't match the listed options

facebook/delphiFacebook/R/responses.R
```diff
@@ -85,7 +86,8 @@ load_response_one <- function(input_filename, params) {
                            Q77 = col_integer(),
                            Q78 = col_integer(),
                            Q79 = col_integer(),
-                           Q80 = col_integer()),
+                           Q80 = col_integer(), 
+                           V1 = col_character()),
```
capnrefsmmat, line 90: isn't V1 an integer?

facebook/delphiFacebook/tests/testthat/test-contingency-calculate.R
```diff
@@ -0,0 +1,16 @@
...
+  
+  expected_output <- list(val = 70.3,
+                          sample_size = sample_size,
+                          se = NA_real_,
+                          effective_sample_size = sample_size,
+                          represented = 70.3)
+  
+  out <- compute_multiple_choice(response, weight, sample_size, 70.3)
+  expect_identical(out, expected_output)
+})
```
capnrefsmmat, line 16: end file with newline

capnrefsmmat (commented):
Looks good so far. This is just a high-level scan; there's so much here that it's hard to do an in-depth review, and I'm leaning heavily on your tests being reasonable. Main comments are just on style, documentation, and requirements

facebook/delphiFacebook/R/contingency_run.R
capnrefsmmat: Actually I'm going to move `bodge_v4_translation` into `load_responses_one`, above where all the variables are coded, because I need to code V4 to put it into the API. So you'll just need to handle this change when merging/rebasing on `fb-package` once I'm done

facebook/delphiFacebook/R/binary.R
```diff
@@ -105,6 +105,7 @@ compute_binary_response <- function(response, weight, sample_size)
 #' @param df Data frame
 #' @return Updated data frame.
 #' @importFrom dplyr mutate
+#' @export
```
krivard, line 108: if we move table configuration into the library, we won't need to export these anymore

facebook/delphiFacebook/R/contingency_aggregate.R
```diff
@@ -0,0 +1,361 @@
+## Functions for performing the aggregations in an efficient way.
+
+#' Produce aggregates for all desired aggregations.
```
krivard, line 3:
```suggestion
#' Produce all desired aggregations.
```

facebook/delphiFacebook/R/contingency_aggregate.R
```diff
@@ -0,0 +1,361 @@
...
+#' @param params Named list of configuration parameters.
+#'
+#' @return none
+#'
+#' @import data.table
+#' @importFrom dplyr full_join %>%
+#' @importFrom purrr reduce
+#'
+#' @export
+aggregate_aggs <- function(df, aggregations, cw_list, params) {
```
krivard, line 31: any reason not to name this something a little less reduplicative? eg
```suggestion
produce_aggregates <- function(df, aggregations, cw_list, params) {
```

facebook/delphiFacebook/R/contingency_aggregate.R
```diff
@@ -0,0 +1,361 @@
...
+
+#' Post-process aggregations and data to make more generic
+#' 
+#' @param df Data frame of individual response data.
+#' @param aggregations Data frame with columns `name`, `var_weight`, `metric`,
+#'   `group_by`, `compute_fn`, `post_fn`. Each row represents one aggregate
+#'   to report. `name` is the aggregate's base column name; `var_weight` is the 
+#'   column to use for its weights; `metric` is the column of `df` containing the
+#'   response value. `group_by` is a list of variables used to perform the 
+#'   aggregations over. `compute_fn` is the function that computes
```
krivard, line 94:
```suggestion
#'   response value. `group_by` is a list of variables to aggregate 
#'   over. `compute_fn` is the function that computes
```

facebook/delphiFacebook/R/contingency_aggregate.R
```diff
@@ -0,0 +1,361 @@
...
+    } else if (startsWith(col_var, "ms_")) { # Multiselect
+      output <- code_multiselect(df, aggregations, col_var)
+      df <- output[[1]]
+      aggregations <- output[[2]]
+      
+    } else if (startsWith(col_var, "n_")) { # Numeric free response
+      output <- code_numeric_freeresponse(df, aggregations, col_var)
+      df <- output[[1]]
+      aggregations <- output[[2]]
+    }
```
krivard, line 178: should there be an "else throw an error for bad column name" case here?

facebook/delphiFacebook/R/contingency_aggregate.R
```diff
@@ -0,0 +1,361 @@
...
+      aggregations <- aggregations[aggregations$metric != col_var & 
+                                     !mapply(aggregations$group_by, 
+                                             FUN=function(x) {col_var %in% x}), ]
+      next
+    }
+    
+    if (startsWith(col_var, "b_")) { # Binary
+      output <- code_binary(df, aggregations, col_var)
+      df <- output[[1]]
+      aggregations <- output[[2]]
```
krivard, line 167: this code is the same for all three branches; factor it out and perform it after the if() block
```suggestion
```

facebook/delphiFacebook/R/contingency_aggregate.R
```diff
@@ -0,0 +1,361 @@
...
+    if (nrow(agg_df) > 0)
+    {
+      s_mix_coef <- params$s_mix_coef
+      mixing <- mix_weights(agg_df[[var_weight]] * agg_df$weight_in_location,
+                            s_mix_coef, params$s_weight)
+      
+      sample_size <- sum(agg_df$weight_in_location)
+      total_represented <- sum(agg_df[[var_weight]] * agg_df$weight_in_location)
+      
+      ## TODO Fix this when same bit is fixed in `aggregate.R`, ~line 181
```
krivard, line 345: Is there a git issue for tracking this fix yet? If so, cite the issue number here. If not, create one and then cite it in both places.

facebook/delphiFacebook/R/contingency_calculate.R
```diff
@@ -0,0 +1,80 @@
...
+compute_binary_and_multiselect <- function(response, weight, sample_size, total_represented)
+{
+  response_pct <- compute_binary_response(response, weight, sample_size)
+  response_pct$sample_size <- sample_size
+  response_pct$represented <- total_represented
+  
+  return(response_pct)
+}
+
+#' Returns multiple choice response estimates. Val is the number of people
```
krivard, line 51:
```suggestion
#' Return multiple choice response estimates. Val is the number of people
```

facebook/delphiFacebook/R/contingency_utils.R
```diff
@@ -0,0 +1,153 @@
...
+    return(c(file_end_date, file_start_date))
+  })), origin="1970-01-01")
+  
+  date_range <- list(min(dates), max(dates))
+  
+  return(date_range)
+}
+
+
+#' Checks user-set aggregations for basic validity
```
krivard, line 109:
```suggestion
#' Check user-set aggregations for basic validity.
```

facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -0,0 +1,446 @@
+## Functions handling renaming, reformatting, or recoding response columns.
+
+#' Rename question codes to informative descriptions
```
krivard, line 3:
```suggestion
#' Rename question codes to informative descriptions.
```

facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -0,0 +1,446 @@
...
+#' @export
+make_human_readable <- function(input_data) {
+  input_data <- remap_responses(input_data)
+  input_data <- rename_responses(input_data)
+  input_data$t_zipcode <- input_data$zip5 # Keep existing parsed zipcode column
+  
+  return(input_data)
+}
+
+#' Rename all columns to make more interpretable
```
krivard, line 29:
```suggestion
#' Rename all columns to make more interpretable.
```

facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -0,0 +1,446 @@
...
+    "n_hh_prop_ili" = "hh_p_ili" # Based on symptoms in A1, and hh sick and total counts
+  )
+  
+  map_new_old_names <- map_new_old_names[!(names(map_new_old_names) %in% names(df))]
+  df <- rename(df, map_new_old_names[map_new_old_names %in% names(df)])
+  
+  return(df)
+}
+
+#' Remap binary columns, race, and others to make more interpretable
```
krivard, line 176:
```suggestion
#' Remap binary columns, race, and others to make more interpretable.
```

facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -0,0 +1,446 @@
...
+                         map_old_new_responses[[col_var]][["map"]], 
+                         map_old_new_responses[[col_var]][["default"]],
+                         map_old_new_responses[[col_var]][["type"]]
+    )
+  }
+  
+  return(df)
+}
+
+#' Convert a response codes in a single response to specified values. Returns
```
krivard, line 290: This sentence doesn't brainparse, and I'm not sure what was intended

facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -0,0 +1,446 @@
...
+#' @return list of data frame of individual response data and user-set data 
+#' frame of desired aggregations
+#'
+#' @export
+code_binary <- function(df, aggregations, col_var) {
+  df <- remap_response(df, col_var, c("1"=1, "2"=0, "3"=NA))
+  return(list(df, aggregations))
+}
+
+#' Convert a single multi-select response column to a set of boolean columns
```
krivard, line 354:
```suggestion
#' Convert a single multi-select response column to a set of boolean columns.
```

facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -0,0 +1,446 @@
...
+    function(code) {
+      paste(col_var, 
+            stri_replace_all(code, "_", fixed=" "), 
+            sep="_")
+      }
+    ))
+  #### TODO: eval(parse()) here is not the best approach, but I can't find another 
+  # way to get col_var (a string) to be used as a var that references a column
+  # rather than as an actual string. This approach causes a shallow copy to be 
+  # made (warning is raised).
```
krivard, line 397: You probably want one of these:
```
!!sym(col_var)
!!as.symbol(col_var)
!!as.name(col_var)
```

facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -0,0 +1,446 @@
...
+      new_row$id <- paste(old_row$id, response_code, sep="_")
+      new_row$metric <- new_binary_cols[col_ind]
+      aggregations <- add_row(aggregations, new_row)
+    }
+  }
+  
+  return(list(df, aggregations[aggregations$metric != col_var, ]))
+}
+
+#' Convert a single free response column to numeric
```
krivard, line 425:
```suggestion
#' Convert a single free response column to numeric.
```

facebook/delphiFacebook/R/contingency_write.R
```diff
@@ -0,0 +1,47 @@
+#' Write csv file for sharing with researchers
```
krivard, line 1:
```suggestion
#' Write csv file for sharing with researchers.
```

facebook/delphiFacebook/R/count.R
```diff
@@ -59,6 +59,7 @@ compute_count_response <- function(response, weight, sample_size)
 #' @param df Data frame
 #' @return Updated data frame.
 #' @importFrom dplyr mutate
+#' @export
```
krivard, line 62: Don't need this if we move the table definitions inside the library as recommended above

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
+#' Get the date of the first day of the previous month
```
krivard, line 1:
```suggestion
#' Get the date of the first day of the previous month.
```

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
...
+#' @return Date
+#' 
+#' @importFrom lubridate floor_date
+#' 
+#' @export
+start_of_prev_full_month <- function(date) {
+  return(floor_date(date, "month") - months(1))
+}
+
+#' Get the date of the last day of the previous month
```
krivard, line 14:
```suggestion
#' Get the date of the last day of the previous month.
```

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
...
+  first_day_of_next_month <- ceiling_date(date, "month")
+  
+  if (first_day_of_next_month != date) {
+    first_day_of_next_month <- floor_date(date, "month")
+  }
+  
+  return(first_day_of_next_month - days(1))
+}
+
+#' Get the date range specifying the previous month
```
krivard, line 33:
```suggestion
#' Get the date range specifying the previous month.
```

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
...
+  if (eom == date) {
+    som <- start_of_prev_full_month(date + months(1))
+  } else {
+    som <- start_of_prev_full_month(date)
+  }
+  
+  return(list(som, eom))
+}
+
+#' Get the date of the first day of the previous epiweek
```
krivard, line 52:
```suggestion
#' Get the date of the first day of the previous epiweek.
```

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
...
+#' @return Date
+#' 
+#' @importFrom lubridate floor_date weeks
+#' 
+#' @export
+start_of_prev_full_week <- function(date) {
+  return(floor_epiweek(date) - weeks(1))
+}
+
+#' Get the date of the last day of the previous epiweek
```
krivard, line 65:
```suggestion
#' Get the date of the last day of the previous epiweek.
```

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
...
+  first_day_of_next_week <- ceiling_epiweek(date)
+  
+  if (first_day_of_next_week != date) {
+    first_day_of_next_week <- floor_epiweek(date)
+  }
+  
+  return(first_day_of_next_week - days(1))
+}
+
+#' Get the date range specifying the previous week
```
krivard, line 84:
```suggestion
#' Get the date range specifying the previous week.
```

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
...
+  if (eow == date) {
+    sow <- start_of_prev_full_week(date + weeks(1))
+  } else {
+    sow <- start_of_prev_full_week(date)
+  }
+  
+  return(list(sow, eow))
+}
+
+#' Get the date range specifying the previous full time period
```
krivard, line 105:
```suggestion
#' Get the date range specifying the previous full time period.
```

facebook/delphiFacebook/R/dates.R
krivard: +1

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
...
+    date_period_range = get_range_prev_full_month(date)
+  } else if (weekly_or_monthly_flag == "week") {
+    # Get start and end of previous full epiweek.
+    date_period_range = get_range_prev_full_week(date)
+  } else {
+    stop("invalid time period selection; must indicate either 'month' or 'week'")
+  }
+  
+  date_period_range[[1]] =  ymd_hms(
+    sprintf("%s 00:00:00", as_date(date_period_range[[1]])), tz = "America/Los_Angeles"
```
krivard, line 128: "America/Los_Angeles" shows up in a bunch of places in this codebase; is it time to make it a constant?

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
...
+  )
+  date_period_range[[2]] =  ymd_hms(
+    sprintf("%s 23:59:59", as_date(date_period_range[[2]])), tz = "America/Los_Angeles"
+  )
+  
+  return(date_period_range)
+}
+
+#' Get date of the first day of the epiweek `x` falls in
```
krivard, line 137:
```suggestion
#' Get date of the first day of the epiweek `x` falls in.
```

facebook/delphiFacebook/R/dates.R
```diff
@@ -0,0 +1,164 @@
...
+#' 
+#' @export
+floor_epiweek <- function(date) { 
+  for (prior_days in 1:7) {
+    if (epiweek(date) != epiweek(date - days(prior_days))) { break }
+  }
+  return(date - days(prior_days - 1))
+}
+
+#' Get date of the first day of the epiweek following the epiweek `x` falls in
```
krivard, line 153:
```suggestion
#' Get date of the first day of the epiweek following the epiweek `x` falls in.
```

facebook/delphiFacebook/R/utils.R
```diff
@@ -5,15 +5,16 @@
 #'
 #' @param path    path to the parameters file; if not present, will try to copy the file
 #'                "params.json.template"
+#' @param template_path    path to the template parameters file
```
krivard, line 8: Prefer one params file to rule them all please

facebook/delphiFacebook/R/utils.R
```diff
@@ -148,7 +152,7 @@ megacounty <- function(
   df_megacounties <- mutate(df_megacounties,
                             geo_id = make_megacounty_fips(.data$geo_id))
 
-  df_megacounties <- group_by(df_megacounties, .data$day, .data$geo_id)
+  df_megacounties <- group_by(df_megacounties, across(groupby_vars))
```
krivard, line 155: oooh very nice!

krivard (changes requested):
Excellent progress, some things to fix

facebook/delphiFacebook/R/contingency_run.R
capnrefsmmat: I pushed the change moving `bodge_v4_translation`, so you should be able to get rid of this now

facebook/delphiFacebook/tests/testthat/test-contingency-run.R
```diff
@@ -0,0 +1,362 @@
...
+  local_mock("delphiFacebook::join_weights" = mock_join_weights)
+  local_mock("delphiFacebook::mix_weights" = mock_mix_weights)
+  local_mock("delphiFacebook::load_archive" = mock_load_archive)
+  run_contingency_tables(params, base_aggs[1,])
+
+  # Expected files
+  expect_equal(!!dir(params$export_dir), c("20200501_nation_gender_anxiety.csv"))
+
+  # Expected file contents
+  raw_data <- read.csv("./input/simple_synthetic.csv")
```
capnrefsmmat, line 243: `test_path` necessary?

facebook/delphiFacebook/tests/testthat/test-contingency-run.R
```diff
@@ -0,0 +1,362 @@
...
+  "pct_comorbidities", "ms_comorbidities", c("mc_gender"), compute_binary_and_multiselect, I,
+)
+
+# Suppress loading of archive to keep output predictable.
+mock_load_archive <- function(...) {
+  return(list(input_data = NULL, seen_tokens = NULL))
+}
+
+get_params <- function(output_dir) {
+  params <- read_params("params-contingency-full.json")
```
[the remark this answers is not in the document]
capnrefsmmat: although I guess that would try to relativize `output_dir`; not sure if that would break stuff

capnrefsmmat (commented):
I took a look through the tests. I think they're pretty thorough, particularly `test-contingency-run.R`, and my comments are mainly about documentation and logistical details.

I guess one thing I don't see tested is the behavior for grouping data by time -- like if I do weekly aggregates, ensuring that each week contains the correct subset of data. You do some tests of your date functions, which is great, but avoiding accidental overlaps in the weekly/monthly data would be good (like if the intervals were [start, end] instead of [start, end), so adjacent intervals include one day of overlap).

facebook/delphiFacebook/R/contingency_aggregate.R
```diff
@@ -0,0 +1,361 @@
...
+          c(aggregations$group_by[agg_ind][[1]], aggregations$metric[agg_ind])
+      }
+    }
+  }
+  
+  # Convert most columns being used in aggregations to the appropriate format.
+  # Multiple choice and multi-select used for grouping are left as-is.
+  #### TODO: multi-select responses should be converted into set of binary cols 
+  #### if being used for grouping; low priority since using a multi-select as a
+  #### grouping variable should be rare.
```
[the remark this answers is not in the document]
nmdefries: What behavior do we want when using multi-selects for grouping? Currently, it works but reports all unique combinations of responses. This is interpretable but may be difficult to work with and censors unusual combinations of responses. For example:

```
geo_id, age, comorbidities 
us,25-34,"Asthma,Autoimmune disorder"
us,25-34,"Asthma,Compromised immune system"
us,25-34,Autoimmune disorder
us,25-34,"Autoimmune disorder,Compromised immune system"
us,25-34,"Autoimmune disorder,Compromised immune system,Type 1 diabetes"
...
```

Some options on handling multi-selects:

* Respondents selecting more than one comorbidity could be recoded as a single group, e.g. "more than 1 comorbidity". This approach doesn't necessarily generalize well to other multi-select questions (e.g. B10b "reasons tested"), but C1/comorbidites is likely to be the most important mult-select item for grouping; the others might never be used.
* Respondents selecting more than one option could be distributed all responses weighted by 1/length(response), like how zip codes are handled
* Recode column as a series of binaries, "selected response 1: true/false". However, this still censors unusual combinations of responses

facebook/delphiFacebook/R/contingency_aggregate.R
```diff
@@ -0,0 +1,361 @@
...
+  assert( length(unique(aggregations$group_by)) == 1 )
+  
+  if ( length(unique(aggregations$name)) < nrow(aggregations) ) {
+    stop("all aggregations using the same set of groupby variables must have unique names")
+  }
+  
+  ## dplyr complains about joining a data.table, saying it is likely to be
+  ## inefficient; profiling shows the cost to be negligible, so shut it up
+  # Geo group column is always named "geo_id"
+  df <- suppressWarnings(inner_join(df, crosswalk_data, by = "zip5"))
```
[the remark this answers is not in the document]
nmdefries: Ah, I think this was a "what is this doing" comment to myself. I meant that whatever geo level we're using (state, nation, county, etc) the geo name column is always called "geo_id", not "state" or "nation", etc.

facebook/delphiFacebook/R/contingency_aggregate.R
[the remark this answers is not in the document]
capnrefsmmat: I think maybe we can defer this decision until someone has a specific analysis use case and knows which of these options would fit their needs

facebook/delphiFacebook/R/contingency_aggregate.R
nmdefries: We shouldn't need an error here; there's an if block just before this that skips this section if the column doesn't exist.

facebook/delphiFacebook/R/contingency_calculate.R
```diff
@@ -0,0 +1,80 @@
...
+#' @param sample_size The sample size to use, which may be a non-integer (as
+#'   responses from ZIPs that span geographical boundaries are weighted
+#'   proportionately, and survey weights may also be applied)
+#' @param total_represented Number of people represented in sample, which may
+#'   be a non-integer
+#'      
+#' @return a list of named percentages and other descriptive statistics
+#'
+#' @export
+compute_binary_and_multiselect <- function(response, weight, sample_size, total_represented)
```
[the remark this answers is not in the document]
nmdefries: Multi-select items (where a respondent can select more than one response, e.g. comorbidities) are converted into a series of binary columns, one for each response code. So the calculation for a multi-select item is the same as the binary calculation but applied to multiple columns.

Since these `compute_*` functions are used in constructing the aggregations table, I wanted them to have names that would make it clear when to use them. I could make separate `binary` and `multiselect` versions if you think that would be clearer.

facebook/delphiFacebook/tests/testthat/test-contingency-run.R
```diff
@@ -0,0 +1,362 @@
...
+  expect_setequal(!!dir(params$export_dir), c("20200501_nation_gender.csv"))
+
+  # Expected file contents
+  expected_output <- as.data.frame(tribble(
+    ~geo_id, ~mc_gender, 
+    ~val_pct_comorbidities_None_listed, ~sample_size_pct_comorbidities_None_listed, ~represented_pct_comorbidities_None_listed,
+    ~val_pct_comorbidities_Type_1_diabetes, ~sample_size_pct_comorbidities_Type_1_diabetes, ~represented_pct_comorbidities_Type_1_diabetes,
+    "us", "Female", 
+    100, 2000L, 100 * 2000,
+    1/2000 * 100, 2000L, 100 * 2000
```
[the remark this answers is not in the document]
nmdefries: Whoops, 1 person has both "No comorbidities" and diabetes. Everyone else has no comorbidities.

facebook/delphiFacebook/tests/testthat/test-contingency-run.R
```diff
@@ -0,0 +1,362 @@
...
+  run_contingency_tables(params, base_aggs[4,])
+
+  # Expected files
+  expect_setequal(!!dir(params$export_dir), c("20200501_nation_gender.csv"))
+
+  # Expected file contents
+  expected_output <- as.data.frame(tribble(
+    ~geo_id, ~mc_gender, 
+    ~val_pct_comorbidities_None_listed, ~sample_size_pct_comorbidities_None_listed, ~represented_pct_comorbidities_None_listed,
+    ~val_pct_comorbidities_Type_1_diabetes, ~sample_size_pct_comorbidities_Type_1_diabetes, ~represented_pct_comorbidities_Type_1_diabetes,
```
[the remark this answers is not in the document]
nmdefries: Yeah, I only used a very small subset of comorbidities (2 different response codes) for ease of specifying expected output.

facebook/delphiFacebook/tests/testthat/test-contingency-run.R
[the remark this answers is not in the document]
nmdefries: Switched this to high blood pressure + diabetes so it's more sensical.

facebook/delphiFacebook/R/contingency_variables.R
nmdefries: I can't get the `!!` operator to be recognized in this context. I can't quite find a clear explanation why, but it seems like it's something to do with `!!` only being recognized within `dplyr` functions, whereas this is in a regex function (but within a `dplyr` function...). I'm going to shelve this for now and revisit as an incremental update.

krivard (approved):
No substantive changes required. Nice work!

PR #629, opened 2020-12-11
Add try/except to GHT

### LINKED ISSUE #631: Make GHT indicator robust to the whims of the GHT API
The GHT API has flip-flopped twice this week regarding whether it requires `timelineResolution` values to be in all-caps or lowercase. Expecting that they will change their minds again, the indicator should be made to try the other approach if the first one crashes.

chinandrew:
### Description
GHT validation seems to switch between uppercase and lowercase, so adding a try/except to attempt both. See [this slack thread.](https://delphi-org.slack.com/archives/C0130CSQRN3/p1607536334139600)

### Changelog
- Add try except that defaults to lowercase 'day' and then tries 'DAY' if it catches a type error

### Fixes 
- Fixes #631

### DIFF
--- a/google_health/delphi_google_health/pull_api.py
+++ b/google_health/delphi_google_health/pull_api.py
@@ -61,7 +61,7 @@ class GoogleHealthTrends:
         params = {
             "time_startDate": start_date,
             "time_endDate": end_date,
-            "timelineResolution": "DAY",
+            "timelineResolution": "day",
             "terms": TERMS,
         }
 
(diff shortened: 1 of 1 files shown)


chinandrew:
actually this needs some more work, since its not the type error thats being raised, the type error is causing a tenacity.RetryError. will close and reopen

chinandrew:
Actually I think this was the right thing to do, reopening.

krivard (approved):
lgtm

Issue #627, opened 2020-12-10
Add data source for HHS hospitalizations using per-facility Epidata dataset


krivard:
We are expediting creation of a new COVIDcast source+signals based on the following dataset:
https://healthdata.gov/dataset/covid-19-reported-patient-impact-and-hospital-capacity-facility

The raw data is expected to be served from its own Epidata endpoint shortly.

**Data details**
 
This data encodes hospital admissions on a per-facility basis, making it straightforward to aggregate to all COVIDcast geo levels.

Data is aggregated on a weekly basis, **Friday-Thursday** (note this is different from the standard epiweek definition, Sunday-Saturday)

The signals we want to compute for COVIDcast are:
- [ ] "ground truth" hospital admissions (`previous_day_admission_adult_covid_confirmed_7_day_sum` + `previous_day_admission_pediatric_covid_confirmed_7_day_sum`)
- [ ] ??? TBD

chinandrew:
do we have a good sense of how many signals we'll want and what sorts of operations we'll want to be doing to generate them (e.g. summing two columns)? I know there's probably some decent uncertainty, I'm mainly trying to think of the right ways to design the pipeline if we'll end up with signals requiring operations on varying column(s).

krivard:
Until we can hear from Roni, probably best to use an agile approach so we can get out the minimum (ground truth) ASAP. Then as we hear more about what is needed, we can build onto that and refactor if necessary.

PR #632, opened 2020-12-11
HHS Facility Level Indicator

chinandrew:
### Description
Draft version of the new HHS indicator. While waiting for the API to go online, I'm testing by using pd.read_csv on the downloaded CSV (I guess you could just pass in the healthdata.gov URL and it could work in prod) instead of querying the epidata API. 

Main design thing to note is how the indicator is configured as a tuple of `(name, columns, function, offset)`. For the one indicator currently specified and the CSV from yesterday which has collection weeks (20201127, 20201120, etc...), it will generate the following files

20201126_county_total_admissions_7d.csv
20201126_hrr_total_admissions_7d.csv
20201126_msa_total_admissions_7d.csv
20201126_state_total_admissions_7d.csv
20201119_county_total_admissions_7d.csv
20201119_hrr_total_admissions_7d.csv
20201119_msa_total_admissions_7d.csv
20201119_state_total_admissions_7d.csv
....etc...

### Changelog
- Add an initial indicator specification 
- Implement pipeline without the data retrieval.
- Unit tests

### Fixes 
- Addresses #627

### DIFF
--- a/hhs_facilities/delphi_hhs_facilities/constants.py
+++ b/hhs_facilities/delphi_hhs_facilities/constants.py
@@ -3,7 +3,7 @@ from .generate_signals import sum_cols
 
 NAN_VALUE = -999999.0
 
-ADMISSION_TOTAL = "total_admissions_7dav"
+ADMISSION_TOTAL = "total_admissions_7d"
 
 SIGNALS = [
     # (name, columns to use, operation, date offset)
--- a/hhs_facilities/delphi_hhs_facilities/constants.py
+++ b/hhs_facilities/delphi_hhs_facilities/constants.py
@@ -1,14 +1,6 @@
-"""Registry for signal names and geographies to process."""
+"""Registry for geographies to process."""
 
-TOTAL_BEDS_7DAV = "total_beds_7_day_avg"
-
-SIGNALS = [
-    TOTAL_BEDS_7DAV
-]
 
 GEO_RESOLUTIONS = {
-    "county",
-    "state",
-    "msa",
-    "hrr"
+
 }
--- a/hhs_facilities/delphi_hhs_facilities/constants.py
+++ b/hhs_facilities/delphi_hhs_facilities/constants.py
@@ -1,6 +1,6 @@
 """Registry for geographies to process."""
 
-ADMISSION_TOTAL = "total_admission"
+ADMISSION_TOTAL = "total_admissions_7dav"
 
 SIGNALS = [
     # (name, columns to use, operation, date offset)
--- a/hhs_facilities/delphi_hhs_facilities/constants.py
+++ b/hhs_facilities/delphi_hhs_facilities/constants.py
@@ -1,6 +1,19 @@
 """Registry for geographies to process."""
 
+ADMISSION_TOTAL = "total_admission"
 
-GEO_RESOLUTIONS = {
+SIGNALS = [
+    # (name, columns to use, operation, date offset)
+    (ADMISSION_TOTAL,
+     ["previous_day_admission_adult_covid_confirmed_7_day_sum",
+      "previous_day_admission_pediatric_covid_confirmed_7_day_sum"],
+     sum,
+     -1),
+]
 
-}
+GEO_RESOLUTIONS = [
+    "county",
+    "msa",
+    "state",
+    "hrr"
+]
--- a/hhs_facilities/delphi_hhs_facilities/constants.py
+++ b/hhs_facilities/delphi_hhs_facilities/constants.py
@@ -6,9 +6,9 @@ SIGNALS = [
     TOTAL_BEDS_7DAV
 ]
 
-GEO_RESOLUTIONS = [
+GEO_RESOLUTIONS = {
     "county",
     "state",
     "msa",
     "hrr"
-]
+}
--- a/hhs_facilities/delphi_hhs_facilities/constants.py
+++ b/hhs_facilities/delphi_hhs_facilities/constants.py
@@ -1,0 +1,14 @@
+"""Registry for signal names and geographies to process."""
+
+TOTAL_BEDS_7DAV = "total_beds_7_day_avg"
+
+SIGNALS = [
+    TOTAL_BEDS_7DAV
+]
+
+GEO_RESOLUTIONS = [
+    "county",
+    "state",
+    "msa",
+    "hrr"
+]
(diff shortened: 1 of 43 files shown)


hhs_facilities/delphi_hhs_facilities/constants.py
```diff
@@ -1,14 +1,22 @@
-"""Registry for signal names and geographies to process."""
+"""Registry for signals and geographies to process."""
+from .generate_signals import sum_cols
 
-TOTAL_BEDS_7DAV = "total_beds_7_day_avg"
+NAN_VALUE = -999999.0
+
+ADMISSION_TOTAL = "total_admissions_7dav"
```
krivard, line 6: total nit but since you're not dividing by 7 to get an average,
```suggestion
ADMISSION_TOTAL = "total_admissions_7d"
```

hhs_facilities/delphi_hhs_facilities/generate_signals.py
```diff
@@ -0,0 +1,56 @@
...
+    ----------
+    df: pd.DataFrame
+        Input DataFrame containing columns specified in `input_cols`.
+    input_cols: list of strings
+        List of column names to pass to `signal_func`.
+    signal_func: function
+        Function which takes in a list of Series and produces a signal Series.
+    date_offset: integer
+        Number of days to add to the timestamp. This is used because some of the columns are
+        "previous_day_<metric>" and require us adding -1 days to accurate depict the right timespan.
```
krivard, line 28:
```suggestion
        "previous_day_<metric>" and require us adding -1 days to represent the right timespan.
```

hhs_facilities/delphi_hhs_facilities/generate_signals.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def generate_signal(df: pd.DataFrame,
+                    input_cols: list,
+                    signal_func: Callable,
+                    date_offset: int) -> pd.DataFrame:
+    """
+    Generate a signal DataFrame derived from an input DataFrame.
+
+    Applies the provided function the the columns specified, and then aggregates by geo and time.
```
krivard, line 16:
```suggestion
    Applies the provided function on the columns specified, and then aggregates by geo and time.
```

hhs_facilities/delphi_hhs_facilities/generate_signals.py
```diff
@@ -0,0 +1,56 @@
...
+    Parameters
+    ----------
+    cols: list of Series
+        List of Series to sum.
+
+    Returns
+    -------
+    Series of summed inputs.
+    """
+    return pd.concat(cols, axis=1).sum(axis=1, min_count=1)
```
krivard, line 56: if all values are nan, is the row removed from the output?

hhs_facilities/delphi_hhs_facilities/geo.py
```diff
@@ -0,0 +1,37 @@
...
+
+import pandas as pd
+from delphi_utils.geomap import GeoMapper
+
+
+def convert_geo(df: pd.DataFrame, geo: str, gmpr: GeoMapper) -> pd.DataFrame:
+    """
+    Map a DataFrame to desired regions.
+
+    The HHS facility level data contains a zip, state, and fips columns. For state and fips, we
```
krivard, line 11:
```suggestion
    The HHS facility level data contains columns for zip, state, and fips. For state and fips, we
```

hhs_facilities/delphi_hhs_facilities/geo.py
```diff
@@ -0,0 +1,37 @@
...
+    use them as given. For all other geos, we map from zip (the smallest of the regions) to the
+    desired geo.
```
krivard, line 13: > For all other geos, we map from zip

Some zips cross county and state lines, so the geomapper will split those entries proportionally by population. That's probably not what we want, unless we have reason not to trust the FIPS published by HHS.

Are there facilities that provide a zip but no fips?

krivard (changes requested):
One biggie, couple nits

hhs_facilities/delphi_hhs_facilities/geo.py
chinandrew: > Are there facilities that provide a zip but no fips?
Good question, here's what I get:
```
In [5]: sum(pd.isna(df.zip))                                                                                                                                         
Out[5]: 9

In [6]: sum(pd.isna(df.fips_code))                                                                                                                                   
Out[6]: 239
```

I wonder if we should do a more custom mapping, e.g. for HRR do zip -> HRR and fips -> state since those nest properly. Not sure if we want to map the missing fips from the zips

hhs_facilities/delphi_hhs_facilities/generate_signals.py
chinandrew: Not from this function, since it needs to be the same dimension as the input df to become a new column. I can add a dropna() statement so we don't get this behavior
```
In [18]: df                        
Out[18]: 
   val geo
0  NaN   a
1  NaN   a
2  3.0   b
3  4.0   b

In [19]: df.groupby('geo', as_index=False).sum()                          
  geo  val
0   a  0.0
1   b  7.0
```

and instead get 
```
In [20]: df.dropna(subset=['val']).groupby('geo', as_index=False).sum()                    
Out[20]: 
  geo  val
0   b  7.0
```

hhs_facilities/delphi_hhs_facilities/constants.py
chinandrew: thanks, apparently my mind has gotten stuck on thinking everything thats 7 days is an average.

chinandrew:
still one open question on geo mapping, but going to mark as ready for review now that tests are added

hhs_facilities/delphi_hhs_facilities/geo.py
krivard: > Not sure if we want to map the missing fips from the zips

It's a matter of what the most reliable guess is. If the FIPS is included by the source, that's the best estimate of the actual FIPS. If the FIPS is missing, then the geo utility's estimate(s) from ZIP5 are the best guess.

krivard: So probably do a preprocessing pass to fill in missing FIPS and then proceed with your approach

chinandrew: ok that sounds good to me. I've implemented the zip->HRR and fips->else logic in this PR, and will follow up with a separate PR to fill in missing FIPS

krivard (approved):
xlnt

PR #637, opened 2020-12-14
Impute FIPS if zip is present for HHS Faciliies

chinandrew:
### Description
Merge after #632. 

Some rows in the HHS data have a zip but no fips, so this adds a function that imputes the missing fips from the zip. See [this thread](https://github.com/cmu-delphi/covidcast-indicators/pull/632#discussion_r541273291) for discussion. Ended up being a little messier than I thought.

### Changelog
-  Add function and unit tests to impute fips

### Fixes 
- n/a

### DIFF
--- a/hhs_facilities/delphi_hhs_facilities/geo.py
+++ b/hhs_facilities/delphi_hhs_facilities/geo.py
@@ -69,8 +69,8 @@ def fill_missing_fips(df: pd.DataFrame, gmpr: GeoMapper) -> pd.DataFrame:
     # we keep unmapped zips because they may be used in HRR mapping so we don't want to drop them
     added_fips = gmpr.add_geocode(no_fips, "zip", "fips", dropna=False)
     added_fips["fips_code"] = added_fips["fips"]
-    # set weight of unmapped zips to 1 to they don"t zero out all the values when multiplied
-    added_fips.weight.fillna(1, inplace=True)  
+    # set weight of unmapped zips to 1 to they don't zero out all the values when multiplied
+    added_fips.weight.fillna(1, inplace=True)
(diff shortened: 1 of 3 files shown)


hhs_facilities/delphi_hhs_facilities/geo.py
```diff
@@ -38,3 +39,39 @@ def convert_geo(df: pd.DataFrame, geo: str, gmpr: GeoMapper) -> pd.DataFrame:
...
+    Returns
+    -------
+    DataFrame with missing FIPS imputed with zip.
+    """
+    mask = pd.isna(df["fips_code"]) & ~pd.isna(df["zip"])
+    no_fips = df[mask]
+    fips_present = df[~mask]
+    no_data_cols = [c for c in df.columns if df[c].dtypes not in (dtype("int64"), dtype("float64"))]
+    data_cols = list(set(df.columns) - set(no_data_cols))
+    # we keep unmapped zips because they may be used in HRR mapping so we don't want to drop them
```
krivard, line 69: > we keep unmapped zips because they may be used in HRR mapping

is that actually true? how many zips have HHRs but not FIPS codes?

krivard (approved):

hhs_facilities/delphi_hhs_facilities/geo.py
chinandrew: I saw one zip that didn't get mapped to a fips: 00935 (in puerto rico). It's also not in our HRR mapping file, but I figure it _could_ be there so it would be better to keep, and otherwise it would just get dropped during HRR mapping.

The other thing this gives us is consistent column sums between a df before and after imputation, so this function is just responsible for filling in values where possible and ignoring otherwise. Then any dropping of values is handled by the later geomapping. I can update the comment accordingly

PR #639, opened 2020-12-14
Add Run tests and factor out data retrieval

chinandrew:
### Description
Couple of changes here to further prep the HHS indicator. 

### Changelog
- Add suspected + confirmed admissions signal. Names still not finalized but it's pretty straightforward to make edits.
- Create a `pull_data()` placeholder function which is still reading the csv as before. This lets me mock out the retrieval in the end to end test. Once API is ready it will get filled in here properly.
- Add tests for `run_module()`. Right now I'm comparing output to some stored expected csvs that I verified by hand. Slightly clunky if we're adding new signals and we want to add new expected files, but don't think it's any worse than manually initializing a bunch of dfs. If anyone has better ideas I'm all ears.

### Fixes 
- n/a


hhs_facilities/tests/test_run.py
```diff
@@ -0,0 +1,40 @@
...
+            "fips_code": ["25013", "25013", np.nan, np.nan],
+            "zip": ["01001", "01001", "00601", "00601"],
+            "state": ["AL", "AL", "PR", np.nan],
+            "previous_day_admission_adult_covid_confirmed_7_day_sum": [1., 2., 3., 4.],
+            "previous_day_admission_pediatric_covid_confirmed_7_day_sum": [10., 20., 30., 40.],
+            "previous_day_admission_adult_covid_suspected_7_day_sum": [0, 50, 0, 50],
+            "previous_day_admission_pediatric_covid_suspected_7_day_sum": [5., 10., 15., 20.]
+        })
+        with tempfile.TemporaryDirectory() as tmp:
+            # change tmp to './expected' to generate new expected files.
```
krivard, line 30:
```suggestion
            # when adding tests for new signals, change tmp to './expected' to generate new expected files.
```

krivard (changes requested):
clarify docs for generating expected files

krivard (approved):
xlnt

krivard:
wat; y u fail checks, google symptoms?

PR #641, opened 2020-12-15
Add hhs and nation to google symptoms

chinandrew:
### Description
Add HHS and Nation regions to google symptoms. There's some unique processing going on for MSAs and HRRs in `geo_map()` that I didn't quite get, but since HHS is just a mapping of states and nation is a collection of everything, I _think_ I dont have to worry about any of the weighted sums. @jingjtang can you verify if this is true? Also just noting that this is a relatively low pri PR so don't be afraid to punt if other things are pressing.

Another thing I noticed when trying to verify the outputs is that the `val` columns don't sum to the same thing across geographies, but maybe that's expected. e.g. here's the sums of the `val` column in 20200726_<geo>_ageusia_raw_search.csv
state = 6.82
nation = 2.22
msa = 1.38
hrr = 2.215
hhs = 2.22
county = 1.56
 
### Changelog
- Add conditional statement to capture HHS and Nation geos
- Update constants

### Fixes 
- Fixes #(issue)


jingjtang (approved):
lgtm

PR #643, opened 2020-12-15
Fix google symptoms test

chinandrew:
### Description
Cast google symptoms test series to list before pytest approx, which breaks in their latest release today.

### Changelog
- cast test Series to lists

### Fixes 
- Fixes #(issue)


krivard (approved):
🚀

PR #644, opened 2020-12-15
Add support for writing logs to file via param. 

### LINKED ISSUE #606: Add structured logging to an indicator

benjaminysmith:
### Description
Add support for writing logs to file via param. 

### Changelog
- Read params for logging file location (param "log_filename")
- Pass this to logger as optional parameter
- Pass logger directly to helper code instead of creating twice
- Convert SirCAL logging to use this approach

### Fixes 
- Fixes #606


sir_complainsalot/delphi_sir_complainsalot/run.py
```diff
@@ -20,21 +20,24 @@
 
 from .check_source import check_source
 
-logger = get_structured_logger(__name__)
+def get_logger():
+    params = read_params()
+    return get_structured_logger(__name__, filename = params.get("log_filename"))
```
chinandrew, line 25: is there a convention to init the logger outside the main function as a constant (as opposed to putting it right after the read_params line in run_module)? I'm not super familiar with the best practice. Otherwise this PR looks good

benjaminysmith: I looked through a bunch of discussions on stack overflow, and there were advocates for both versions of this (and other versions too) so my reaction was that there is not a clear preference. I chose this one because it's slightly less work to pass around the logger to the individual functions (and I am also biased toward a java-ey approach).

chinandrew (approved):
:shipit::shipit::shipit:

PR #645, opened 2020-12-16
Add a monthly rollup CSV generator

capnrefsmmat:
This takes all the daily `.csv.gz` files in a directory (such as pulled from SFTP), finds the latest recordedby for each date, and produces a single CSV.

I've chosen to set readr to use `col_character()` for everything, so we don't have problems with it being "smart" and converting something. I verified on a single data file that this results in the output CSV being a character-for-character match for the input CSV, so no funny business happened.

When this is plugged into the automation, we should generate each month since April. I suggest that we should make the CSV for each month a few days into the next one, to allow backfill to happen -- say, 5ish days into the next month.

My main worry is that R won't be too robust to failures, so I dunno what will happen if something goes wrong during the automated process and the CSV fails to be generated.

### DIFF
--- a/facebook/monthly-files.R
+++ b/facebook/monthly-files.R
@@ -22,10 +22,10 @@ suppressPackageStartupMessages({
 #'
 #' There can be multiple data files for a single day of survey responses, for
 #' example if the data is reissued when late-arriving surveys are recorded.
-#' Each file contains *all* data recorded for that date, so only the later files
-#' are needed.
+#' Each file contains *all* data recorded for that date, so only the most
+#' recently updated file for each date is needed.
 #'
-#' This function extracts the date from each file, determines which files are
+#' This function extracts the date from each file, determines which files contain
 #' reissued data, and produces a single data frame representing the most recent
 #' data available for each day. It can read gzip-compressed CSV files, such as
 #' those on the SFTP site, using `readr::read_csv`.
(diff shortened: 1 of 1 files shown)


facebook/monthly-files.R
```diff
@@ -0,0 +1,98 @@
...
+  library(readr)
+  library(purrr)
+})
+
+#' Fetch all survey data in a chosen directory.
+#'
+#' There can be multiple data files for a single day of survey responses, for
+#' example if the data is reissued when late-arriving surveys are recorded.
+#' Each file contains *all* data recorded for that date, so only the later files
+#' are needed.
```
krivard, line 26:
```suggestion
#' Each file contains *all* data recorded for that date, so only the most
#' recently updated file for each date is needed.
```

facebook/monthly-files.R
```diff
@@ -0,0 +1,98 @@
...
+#'
+#' This function extracts the date from each file, determines which files are
```
krivard, line 28:
```suggestion
#' This function extracts the date from each file, determines which files contain
```

krivard:
to confirm -- usage is
```
$ Rscript monthly-files.R YYYY NN path/to/individual/files/ >outfile.csv
```

?

krivard (approved):
Minor docs fixes but otherwise 👍

capnrefsmmat:
Yep, that's the exact usage

Issue #579, opened 2020-11-30
Alternate data hierarchy for validation parameters/members


sgsmob:
The validation code expects a long list (>15) of parameters to be set in the `params.json` file and then copied over to members of the `Validator` class.  This not only makes the `Validator` class fail linting but also is hard to reason about.  We should refactor both the structure of the `params.json` file as well as the members of `Validator` to be hierarchical rather than flat, i.e.:

```
{
  "date_options": {
    "start":  ...
    "end":  ...
  }
  "single_file_options": {
    "min_sample_size": ...  
  }
  "signal_options": {
    "lag": ...
  }
}
```

PR #646, opened 2020-12-16
Refactor validator to be lint compliant

sgsmob:
### Description
Refactor the `Validator` class to have two components: `StaticValidator` and `DynamicValidator`.

### Changelog
- Large `Validator` class split into two smaller classes
- Validation tests split into their own files too
- Time window over which the validation is defined is split out into a new data class
- Dropped the `sanity_check_X` parameters in favor of single way to enable/disable specific validation checks.
- Renamed `tests/test_checks.py` to the more consistent `tests/test_validator.py`.

### Fixes 
- Addresses part of #579 (can change the json hierarchy in a follow-on)


validator/delphi_validator/validate.py
```diff
@@ -83,11 +83,9 @@ def __init__(self, params):
         self.suppressed_errors = {(item,) if not isinstance(item, tuple) and not isinstance(
             item, list) else tuple(item) for item in params.get('suppressed_errors', [])}
 
-        self.active_report = ValidationReport(self.suppressed_errors)
```
chinandrew, line 86: whats the rational for moving this from being a class attribute? A Validator object maintaining it's own ValidatorReport doesn't seem unreasonable

validator/delphi_validator/static.py
```diff
@@ -0,0 +1,399 @@
+"""Static file checks."""
+from os.path import join
+import re
+from datetime import date, datetime, timedelta
+import pandas as pd
+from .datafetcher import FILENAME_REGEX
+from .errors import ValidationError
+from .utils import GEO_REGEX_DICT
+
+class StaticValidation:
```
chinandrew: Should these be named Static/DynamicValidator (instead of Validation)?

chinandrew:
Am I correct in assuming that the validator -> static/dynamic splits (2e03f3f and cfe6458) are essentially copy pastes of the old methods into two files? If so I won't spend much time looking into all of the methods that were existing.

validator/delphi_validator/validate.py
sgsmob: The original point of having it as its own attribute was that then you didn't have to pass it as an argument to all its function calls for individual checks.  Now that the checks are in separate objects/files, that advantage is no longer there.  The disadvantage of having it be an attribute is that it unnecessarily introduces an internally persistent state (need to be sure nothing changes between initialization and the actual validation call).  Essentially we are considering two possibilities now:

```
class X:
  def __init__(self):
    self.y = None
  def add(self):
    return self.y + 3
  def run(self):
    self.y = Y()
    return self.do_something()
```
and
```
class X:
  def add(self, y):
    return y + 3
  def run(self):
    y = Y()
    return self.do_something(y)
```

The latter seems much better to me.

validator/delphi_validator/static.py
sgsmob: Yes I think you may have seen a slightly earlier version

sgsmob: i.e. they have already been changed

sgsmob:
> Am I correct in assuming that the validator -> static/dynamic splits ([2e03f3f](https://github.com/cmu-delphi/covidcast-indicators/commit/2e03f3f63a16c621b7bf01f5aa4c66a379bd6b9c) and [cfe6458](https://github.com/cmu-delphi/covidcast-indicators/commit/cfe6458a318c91ee4500fb59b49b03006f485f68)) are essentially copy pastes of the old methods into two files? If so I won't spend much time looking into all of the methods that were existing.

Yes, this change is intended to be a no-op with respect to `python -m delphi_validator` (with the exception of the deleted parameters).

validator/delphi_validator/static.py
```diff
@@ -84,10 +83,11 @@ def check_missing_date_files(self, daily_filenames, report):
         Returns:
             - None
         """
-        number_of_dates = self.end_date - self.start_date + timedelta(days=1)
+        number_of_dates = self.time_window.end_date - self.time_window.start_date +\
```
chinandrew: Should this and the `date_seq` belong as attributes in the TimeWindow object?

validator/delphi_validator/utils.py
```diff
@@ -13,6 +13,15 @@
     'national': r'^[a-zA-Z]{2}$'
 }
 
+
+def get_today():
+    """Wrapper function for mocking out date.today().
```
chinandrew: may be able to use https://github.com/spulec/freezegun for tesiting

validator/delphi_validator/static.py
chinandrew: ah sorry about that. I'm going through individual commits and commenting as I go.

validator/delphi_validator/validate.py
chinandrew: sgtm

validator/delphi_validator/dynamic.py
```diff
@@ -1,59 +1,52 @@
...
 from .errors import ValidationError, APIDataFetchError
 from .datafetcher import get_geo_signal_combos, threaded_api_calls
 from .utils import relative_difference_by_min, TimeWindow
 
 class DynamicValidation:
     """Class for validation of static properties of individual datasets."""
-    def __init__(self, params, suppressed_errors=None):
+
+    @dataclass
+    class Parameters:
```
chinandrew, line 15: whats the benefit of this over just putting these in the init?

validator/delphi_validator/static.py
sgsmob: Changed this to derive from the `TimeWindow.span_length` member.  I can do more if you'd like.

validator/delphi_validator/dynamic.py
sgsmob: We were running into lint issues with the number of class members being too large; it appears that the linter will let us put arbitrarily many variables in a dataclass.

validator/delphi_validator/static.py
chinandrew: I think adding the date_seq to TimeWindow  too would make sense (bit of a nit), since then the TimeWindow tells you the start, end, dates included, length, etc.

validator/delphi_validator/dynamic.py
chinandrew: ah, interesting. would it be easier to ignore the linter?

chinandrew:
> > Am I correct in assuming that the validator -> static/dynamic splits ([2e03f3f](https://github.com/cmu-delphi/covidcast-indicators/commit/2e03f3f63a16c621b7bf01f5aa4c66a379bd6b9c) and [cfe6458](https://github.com/cmu-delphi/covidcast-indicators/commit/cfe6458a318c91ee4500fb59b49b03006f485f68)) are essentially copy pastes of the old methods into two files? If so I won't spend much time looking into all of the methods that were existing.
> 
> Yes, this change is intended to be a no-op with respect to `python -m delphi_validator` (with the exception of the deleted parameters).

:+1: sounds good. few minor questions but otherwise looks good

validator/delphi_validator/static.py
sgsmob: Yes, done

validator/delphi_validator/utils.py
sgsmob: Good idea, done

validator/delphi_validator/dynamic.py
sgsmob: @krivard thoughts?

krivard: Doing it in init fails linting because having that many class members is legit complicated and you should look for another way. This usage seems to fit the definition and desired function of a data class. Are there downsides I'm not aware of?

chinandrew: My only question is around whether
```
    @dataclass
    class Parameters:
        """Configuration parameters."""
        data_source: str
        time_window: TimeWindow
        generation_date: date
        max_check_lookbehind: timedelta
        smoothed_signals: Set[str]
        expected_lag: Dict[str, int]

    def __init__(self, params):
        """
        Initialize object and set parameters.
        Arguments:
            - params: dictionary of user settings; if empty, defaults will be used
        """
        self.test_mode = params.get("test_mode", False)

        self.params = self.Parameters(
            data_source = params['data_source'],
            time_window = TimeWindow.from_params(params["end_date"], params["span_length"]),
            generation_date = date.today(),
            max_check_lookbehind = timedelta(days=params.get("ref_window_size", 7)),
            smoothed_signals = set(params.get("smoothed_signals", [])),
            expected_lag = params["expected_lag"]
        )

   def foo(self):
       ...
       self.params.test_mode
```
is substantially improved over 
```
    def __init__(self, params):
        """
        Initialize object and set parameters.
        Arguments:
            - params: dictionary of user settings; if empty, defaults will be used
        """
        self.test_mode = params.get("test_mode", False)
        self.data_source = params['data_source']
        self.time_window = TimeWindow.from_params(params["end_date"], params["span_length"])
        self.generation_date = date.today()
        self.max_check_lookbehind = timedelta(days=params.get("ref_window_size", 7))
        self.smoothed_signals = set(params.get("smoothed_signals", []))
        self.expected_lag = params["expected_lag"]

   def foo(self):
       ...
       self.test_mode
```
Maybe once the class grows it becomes more useful?

sgsmob: The size growing is the main advantage but a secondary one is that it lets us have other hierarchical information at the class level.  In particular, I'm thinking about a future check enable/disable system that is handled in the `XValidator` classes rather than in the `ValidationReport`.  You would be getting this class initialization:
```
def __init__(self, params):
    self.params = self.Parameters(params)
    self.disabled_checks = self.DisabledChecks(params)
```
rather than
```
def __init__(self, params):
    self.p1 = params["1"]
    ...
    self.pn = params["n"]
    self.enable_check1 = params["check1"]
    ...
    self.enable_checkn = params["checkn"]
```

chinandrew: :+1: sgtm

chinandrew (approved):
:shipit::shipit::shipit:

PR #647, opened 2020-12-17
Add code to pull data from epidata api

chinandrew:
### Description
Draft  in preparation for the Epidata API to go live with HHS facility level data.

_NOTE: Because the latest Epidata hasn't been pushed to pypi, this will not pass linting/CI due to the required Epidata methods not being availablbe. I wrote this by using a local copy of the Epidata package to run and test it. It's still ready for review, but we are blocked by the pypi release to merge this_

### Changelog
- Add code to retrieve data from EpiData API. Errors if there are no results across all regions or if it hits a truncation message
- Add tests.
- Update list of possible nans to include None and integer -999999
- Snuck a signal name update in as well

### Fixes 
- Fixes #(issue)


hhs_facilities/delphi_hhs_facilities/delphi_epidata.py
```diff
@@ -0,0 +1,667 @@
+"""
```
dfarrow0, line 1: are you planning to remove this file when the pypi epidata package is updated? fyi i'm planning to push this update in the next hour or so.

hhs_facilities/delphi_hhs_facilities/pull.py
```diff
@@ -1,22 +1,59 @@
...
+    -------
+    List of dictionaries. Concatenation of all the response['epidata'] lists.
+    """
+    responses = []
+    for state in states:
+        lookup_response = Epidata.covid_hosp_facility_lookup(state)
+        state_hospital_ids = [i["hospital_pk"] for i in lookup_response.get("epidata", [])]
+        for i in range(0, len(state_hospital_ids), 50):
+            response = Epidata.covid_hosp_facility(state_hospital_ids[i:i+50], dates)
+            if "too many results, data truncated" in response["message"]:
```
dfarrow0, line 37: it might be safer to check that `response['result'] == 2` (or even `response['result'] != 1`) in case the message ever changes. the status code is intended to be part of the stable api surface, whereas the message is just intended for debugging.

hhs_facilities/tests/test_pull.py
```diff
@@ -0,0 +1,72 @@
...
+import numpy as np
+
+from delphi_hhs_facilities.pull import _pull_data_iteratively, pull_data
+
+
+class TestPull:
+
+    @patch("delphi_hhs_facilities.pull.Epidata.covid_hosp_facility")
+    @patch("delphi_hhs_facilities.pull.Epidata.covid_hosp_facility_lookup")
+    def test__pull_data_iteratively(self, covid_hosp_facility_lookup, covid_hosp_facility):
```
dfarrow0, line 16: little nit, but this test uses `test__name` while the one below uses `test_name`. i'm not sure what the convention is in pytest, but i recommend naming them the same way.

dfarrow0 (changes requested):

hhs_facilities/delphi_hhs_facilities/delphi_epidata.py
dfarrow0: (done: https://github.com/cmu-delphi/delphi-epidata/releases/tag/v0.0.9-pypi)

chinandrew: Yup this snuck in, I'll remove it.

hhs_facilities/delphi_hhs_facilities/pull.py
chinandrew: I checked -1 originally, but then if a state like virgin islands has no results it throws an error instead of skipping. is `2` == too many results?

dfarrow0: ah, makes sense.

yeah, `2` is only returned when data is truncated.

chinandrew: perfect, I'll have it check for 2 then

hhs_facilities/tests/test_pull.py
chinandrew: Updated

dfarrow0: doh, i see it was because the function name itself started with an underscore. either way is fine imo. looks good

dfarrow0 (approved):

Issue #648, opened 2020-12-17
Fips/Zip crosswalk has more zip codes than zip population file


chinandrew:
```
In [2]: f_z = pd.read_csv("delphi_utils/data/fips_zip_table.csv", dtype = str)                                                                                       

In [3]: zpop = pd.read_csv("delphi_utils/data/zip_pop.csv", dtype = str)                                                                                             

In [4]: len(set(f_z.zip))                                                                                                                                            
Out[4]: 32976

In [5]: len(set(zpop.zip))                                                                                                                                           
Out[5]: 32959

In [6]: set(f_z.zip) - set(zpop.zip)                                                                                                                                 
Out[6]: 
{'57756',
 '57764',
 '57770',
 '57772',
 '57794',
 '99554',
 '99563',
 '99581',
 '99585',
 '99604',
 '99620',
 '99632',
 '99650',
 '99657',
 '99658',
 '99662',
 '99666'}

In [7]: set(zpop.zip) - set(f_z.zip)                                                                                                                                 
Out[7]: set()
```

Those appear to be real zips, unsure why discrepancy exists.

krivard:
Possibilities:
* old/retired zip codes
* zip codes that receive mail but have no residents

?

krivard:
Work needed on this:
* Figure out why these lists don't match ([possibly useful source table](https://github.com/cmu-delphi/covidcast-indicators/issues/1092#issuecomment-875843327))
* Figure out whether we _can_ fix it
* Figure out whether we _should_ fix it

RafaelCatoia:
Hi @krivard, can I start working on this one?

RafaelCatoia:
@krivard , I looked up into those Zips and found some information about them.

> * old/retired zip codes?

I made search on [discontinued Post Offices](https://webpmt.usps.gov/pmt014.cfm) from 01/01/2010 to 01/09/2021 and it returned 698 discontinued Post Offices and the zip codes they covered  in that interval. None of them are  in our list. 

> * Are they zip codes that receive mail but have no residents?

The population size is usually low (One with  ~6k, one ~2k, and others with less than 1k)  and majorly composed by American Indian Or Alaskan Native. 

> * Figure out why these lists don't 
[00_Issue_0648.pdf](https://github.com/cmu-delphi/covidcast-indicators/files/7237056/00_Issue_0648.pdf)
match (possibly useful source table)

[...]

PR #649, opened 2020-12-17
Add geoutil method to return all values for a given geo

chinandrew:
### Description
Using the existing crosswalk files, this method will return all the values for a given geo type. It does not necessarily return all the geo values for the entire country only those present in the crosswalks, e.g. there are ~42k zip codes and we only map 32k, so this only returns 32k

Note that we use the greater zip number re: #648 

### Changelog
- Add geo_geo_values() method to GeoMapper class and unit tests

### Fixes 
- Fixes #(issue)


krivard (approved):
xlnt

PR #650, opened 2020-12-17
Convert hhs_hosp to use geomapper geo values

chinandrew:
### Description
Both fixes a bug where PR and other territories were excluded, and also converts to using the new geomapper method to return all the state id values.

### Changelog
- Use geomapper.get_geo_values() function to get all the states

### Fixes 
- Fixes #(issue)


krivard (approved):
tyvm

PR #651, opened 2020-12-17
Redo validation error recording system

sgsmob:
### Description
Update the `Validator` class to use a simpler and cleaner system for recording individual errors.

Rather than passing around `Exception` subclasses 

### Changelog
- Replace the `Exception` subclass `ValidationError` with a `ValidationFailure` data class
  - Moving away from the `Exception` hierarchy since we are just recording messages rather than raising errors
  - Renaming the class to avoid confusion with `Exception` subclasses like `AssertionError`, `ValueError`, etc.
- Remove logging of partial data frames with errors since they clog up the output rather than enhance it
- Remove superfluous removal of errors once they are suppressed
- Simplify param specification of errors to suppress, assuming that every one is a 2-tuple specifying the check name and data source, respectively.

Depends on #646 and thus need not be reviewed/accepted before it.


validator/delphi_validator/static.py
```diff
@@ -204,14 +209,16 @@ def find_all_unexpected_geo_ids(df_to_test, geo_regex, geo_type):
...
             if len(unexpected_geos) > 0:
-                report.add_raised_error(ValidationError(
-                    ("check_geo_id_format", nameformat),
-                    unexpected_geos, "Non-conforming geo_ids found"))
+                report.add_raised_error(
+                    ValidationFailure("check_geo_id_format",
+                                      nameformat,
+                                      f"Non-conforming geo_ids {unexpected_geos} found"))
 
         if geo_type not in GEO_REGEX_DICT:
```
JedGrabman, line 217: There's a few places in this file that use a "if not... else" pattern which I find harder to read. I would prefer to switch the order of branches.

Not introduced by you, so feel free to skip.

validator/tests/test_validator.py
```diff
@@ -1,14 +1,45 @@
...
+        """Test initialization with suppressed errors."""
+        params = {"data_source": "",
+                  "span_length": 0,
+                  "end_date": "2020-09-01",
+                  "expected_lag": {},
+                  "suppressed_errors": [["a", "b"], ["c", "d"], ["a", "b"]]
+                 }
+
+        validator = Validator(params)
+        assert not validator.suppressed_errors ^ set([("a", "b"), ("c", "d")])
```
JedGrabman, line 25: Is this just checking for equality? If so, I believe `==` will work and be more readable.

sgsmob: Yes, done

validator/delphi_validator/static.py
sgsmob: No this is an antipattern I have been trying to correct.  Done

validator/tests/test_report.py
```diff
@@ -1,37 +1,31 @@
...
 
-    ERROR_1 = ValidationError(("good", date(2020, 10, 5)), "exp 1", "msg 1")
-    ERROR_2 = ValidationError(("bad", date(2020, 11, 18)), "exp 2", "msg 2")
+    ERROR_1 = ValidationFailure("good", "data 1", "msg 1")
+    ERROR_2 = ValidationFailure("bad", "data 2", "msg 2")
 
     def test_add_raised_unsuppressed_error(self):
         """Test that an unsupressed error shows up in the unsuppressed error list."""
-        report = ValidationReport([("bad", "2020-10-05")])
+        report = ValidationReport([("bad", "data 1")])
```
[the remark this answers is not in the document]
sgsmob: Yes, done

validator/tests/test_report.py
```diff
@@ -1,37 +1,31 @@
...
-        assert len(report.errors_to_suppress) == 0
-
-        # Each error can only be surpressed once.
-        report.add_raised_error(self.ERROR_1)
-        assert report.unsuppressed_errors == [self.ERROR_1]
 
     def test_str(self):
         """Test that the string representation contains all information."""
-        report = ValidationReport([("good", "2020-10-05")])
+        report = ValidationReport(set([("good", "data 1")]))
```
JedGrabman, line 28: I think this is a set which contains a list, which will error out because a list isn't hashable. I think you either want `set(("good", "data 1"))` or `set(tuple(["good", "data 1"]))`.

validator/tests/test_static.py
```diff
@@ -197,24 +165,23 @@ def test_non_consecutive_duplicates(self):
         df = pd.DataFrame([["a", "1"], ["b", "2"], ["a", "1"]])
         validator.check_duplicate_rows(df, "file", report)
         assert len(report.raised_warnings) == 1
-        assert report.raised_warnings[0].expression == [2]
-        assert report.raised_warnings[0].check_data_id[1] == "file"
+        assert report.raised_warnings[0].check_name == "check_duplicate_rows"
 
     def test_multiple_distinct_duplicates(self):
         validator = StaticValidator(self.params)
         report = ValidationReport([])
```
JedGrabman, line 172: Update all references to ValidationReport to use sets.

JedGrabman (commented):
Once the set comments are addressed, that should be everything from me. I recommend re-running the tests though, in case I missed something.

validator/tests/test_report.py
sgsmob: This syntax constructs a set from the elements of the list and works as is.

validator/tests/test_static.py
sgsmob: Done

JedGrabman (approved):

PR #652, opened 2020-12-18
Add initial logging to JHU indicator

benjaminysmith:
### Description
Add initial (trivial) logging to JHU indicator so that we start collecting log data

### Changelog
- Adds simple logging to main JHU code.

### Fixes 
- Partial fix to #606

### DIFF
--- a/jhu/delphi_jhu/run.py
+++ b/jhu/delphi_jhu/run.py
@@ -87,7 +87,12 @@ def run_module():
         METRICS, GEO_RESOLUTIONS, SENSORS, SMOOTHERS
     ):
         print(metric, geo_res, sensor, smoother)
-        logger.info(event = "processing data", metric = metric, geo_res = geo_res, sensor = sensor, smoother = smoother)
+        logger.info(
+            event="processing data",
+            metric=metric,
+            geo_res=geo_res,
+            sensor=sensor,
+            smoother=smoother)
         df = dfs[metric]
         # Aggregate to appropriate geographic resolution
         df = geo_map(df, geo_res)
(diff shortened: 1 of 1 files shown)


jhu/delphi_jhu/run.py
```diff
@@ -86,6 +87,12 @@ def run_module():
         METRICS, GEO_RESOLUTIONS, SENSORS, SMOOTHERS
     ):
         print(metric, geo_res, sensor, smoother)
+        logger.info(
+            event="processing data",
```
chinandrew, line 91: is this descriptive enough (e.g. isn't everything in this pipeline "processing data")? I guess the logs will have file/indicator name and the metric/geo/sensor/smoother which seems sufficient, but haven't followed the logging so don't know how these all get used/parsed.

benjaminysmith: I updated to be a bit more specific here, but yeah we should have lots of information to use in the logs. We'll likely go back and edit these as we start understanding what it most helpful to collect.

jhu/delphi_jhu/run.py
```diff
@@ -86,6 +87,12 @@ def run_module():
         METRICS, GEO_RESOLUTIONS, SENSORS, SMOOTHERS
     ):
         print(metric, geo_res, sensor, smoother)
+        logger.info(
+            event="exporting to CSV",
```
chinandrew, line 91:
```suggestion
            event="generating signal and exporting to CSV",
```

chinandrew (approved):
one suggestion, otherwise :shipit: :shipit: :shipit:

PR #653, opened 2020-12-18
Fix covid_hosp date argument

chinandrew:
### Description
Fix argument for the `covid_hosp` function in the `hhs_hosp` indicator

### Changelog
- Update argument from `[date1,date2]` -> `Epidata.range(date1, date2)` which is equiavelent to`{'from':date1, 'to':date2}`

### Fixes 
- Fixes #(issue)


krivard (approved):
👍

PR #654, opened 2020-12-18
Add date intervals and loop to avoid API limits

chinandrew:
### Description
The Epidata API has a row return limit (3650), so this changes the hhs_hosp indicator so it loops through 30 day intervals (54 data points a day) to retrieve the data. Could have done 60 day intervals but 30 felt like a nicer number.

Left them all in run.py instead of a separate time file for now since they're short, though moving them out seems reasonable.

### Changelog
- Add convenience function to convert date to ints 
- Add function that generates the necessary epidata ranges for two dates
- Update unit tests

### Fixes 
- Fixes #(issue)


krivard (approved):
nice work

PR #655, opened 2020-12-18
hhs_hosp: implement confirmed+suspected; make signal names match docs

krivard:
### Description
Slide in an additional signal (confirmed+suspected) to match draft API docs.

### Changelog
- factored out signal building into its own function
- signal building is now verified in tests

### DIFF
--- a/hhs_hosp/delphi_hhs/run.py
+++ b/hhs_hosp/delphi_hhs/run.py
@@ -12,7 +12,7 @@ from delphi_utils.export import create_export_csv
 from delphi_utils.geomap import GeoMapper
 import pandas as pd
 
-from .constants import HOSPITALIZATIONS
+from .constants import SIGNALS, CONFIRMED, SUM_CONF_SUSP
 
 
 def _date_to_int(d):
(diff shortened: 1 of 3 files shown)


hhs_hosp/delphi_hhs/run.py
```diff
@@ -78,18 +78,41 @@ def run_module():
...
+            "val": \
+            all_columns.previous_day_admission_adult_covid_confirmed + \
+            all_columns.previous_day_admission_adult_covid_suspected + \
+            all_columns.previous_day_admission_pediatric_covid_confirmed + \
+            all_columns.previous_day_admission_pediatric_covid_suspected,
+            "se": None,
+            "sample_size": None
+        })
+    raise Exception(
+        "Bad programmer: signal '{sig}' in SIGNALS but not handled in make_signal"
```
chinandrew, line 117: :laughing: 

related discussion for the future: should we add some custom exceptions that all the indicators can share? Not sure how many cases there would be and don't think it's that important right now, but might be a nice-to-have.

hhs_hosp/tests/test_run.py
```diff
@@ -35,4 +36,31 @@ def test_generate_date_ranges():
...
+    pd.testing.assert_frame_equal(expected_confirmed, make_signal(data, CONFIRMED))
 
+    expected_sum = pd.DataFrame({
+        'geo_id': ['na'],
+        'timestamp': [datetime(year=2020, month=1, day=1)],
+        'val': [15],
+        'se': None,
+        'sample_size': None
+    })
+    pd.testing.assert_frame_equal(expected_sum, make_signal(data, SUM_CONF_SUSP))
```
chinandrew, line 66: add test for the error case

chinandrew (approved):
:shipit::shipit::shipit:

PR #659, opened 2020-12-22
Update hhs signal name in staging api

korlaxxalrok:
### Description
We want to be able to query for `hhs` in the staging api. Was `hhs_hosp`
previously.


chinandrew (approved):
:shipit: :shipit: :shipit:

PR #667, opened 2020-12-28
Add hhs and nation to quidel covidtest

chinandrew:
### Description
Add HHS and nation to Quidel. Some more code changes than the other hhs/nation due to the parent state cases for quidel.

### Changelog
- Generalize state case (which was the only non-parent state geo) to a list of geos, including hhs and nation, that dont use parent states and loop through them
- Add test cases for geo_map hhs/nation and update test_run to check all files
- refactor some function and constant names for the parent/nonparent state split.

### Fixes 
- Fixes #(issue)


jingjtang (approved):
lgtm

Issue #507, opened 2020-11-12
Add or correctly fetch JHU-CSSE data for remaining U.S. jurisdictions in covidcast


brookslogan:
The covidcast package (and perhaps the API?) does not recognize AS, GU, MP, or VI, but these jurisdictions (as well as the Diamond Princess) are reported in the [JHU data](https://github.com/CSSEGISandData/COVID-19/blob/master/csse_covid_19_data/csse_covid_19_daily_reports_us/11-08-2020.csv).

```
## These succeed:
covidcast::covidcast_signal(data_source="jhu-csse", signal="deaths_incidence_num", geo_type="state", geo_values="AL", start_day="2020-09-01", end_day="2020-09-15")
covidcast::covidcast_signal(data_source="jhu-csse", signal="deaths_incidence_num", geo_type="state", geo_values="PR", start_day="2020-09-01", end_day="2020-09-15")
## These fail:
covidcast::covidcast_signal(data_source="jhu-csse", signal="deaths_incidence_num", geo_type="state", geo_values="AS", start_day="2020-09-01", end_day="2020-09-15")
covidcast::covidcast_signal(data_source="jhu-csse", signal="deaths_incidence_num", geo_type="state", geo_values="GU", start_day="2020-09-01", end_day="2020-09-15")
covidcast::covidcast_signal(data_source="jhu-csse", signal="deaths_incidence_num", geo_type="state", geo_values="MP", start_day="2020-09-01", end_day="2020-09-15")
covidcast::covidcast_signal(data_source="jhu-csse", signal="deaths_incidence_num", geo_type="state", geo_values="VI", start_day="2020-09-01", end_day="2020-09-15")
## (and original source data also contains values for Diamond Princess)
```

The abbreviation to FIPS translation appears to be failing.  It might also be the case that the data is not available via API calls at all; I have not tested this.

dshemetov:
Can confirm, we currently drop those geocodes. Questions:
* do those states have county codes and does JHU report at that level?
* do we have a population source for those county codes and are those supported in the county<->zip file?
* if any of these were recoded by JHU at some point, we will need to update the backfill procedure

Diamond Princess and Grand Princess were removed because we mapped them to FIPS 88888 and 99999, which the API loader considers an error.

chinandrew:
[...]

PR #668, opened 2020-12-28
Add territories to mapping files

chinandrew:
### Description
Add  support for American Samoa, Guam, US Virgin Islands, and Mariana Islands. Hard coded the fips and population numbers in because they're in 4 separate excel files that dont contain the fips codes, just the region names. Link is commented in code and also in [this issue](https://github.com/cmu-delphi/covidcast-indicators/issues/507#issuecomment-747914143). Doesn't include any zip data yet, will need to find those if we want them. I'm not sure if anyone reports data for territories in zip format.

Not really sure how we want to approach backfilling for source that report this (e.g. JHU, which reports at the state level and not at the county level, or safegraph, which had a sample file report at the county level). Converting to draft mode after running checks until we're clear on what needs to happen after merging.

### Changelog
- Add territory populations to fips population generation code
- Update mapping files after rerunning geo_data_proc.py
- Fix a safegraph test that had virgin islands and guam data that were previously dropped

### Fixes 
- #507


_delphi_utils_python/data_proc/geomap/geo_data_proc.py
```diff
@@ -400,7 +400,14 @@ def create_fips_population_table():
     df_pr = df_pr[~df_pr["fips"].isin(census_pop["fips"])]
     census_pop_pr = pd.concat([census_pop, df_pr])
 
-    census_pop_pr.to_csv(join(OUTPUT_DIR, FIPS_POPULATION_OUT_FILENAME), index=False)
+    # Filled from https://www.census.gov/data/tables/2010/dec/2010-island-areas.html
+    territories_pop = pd.DataFrame({
+        "fips": ["60010", "60020", "60030", "60040", "60050", "66010", "78010", "78020", "78030", "69085", "69100", "69110", "69120"],
+        "pop": [23030, 1143, 0, 17, 31329, 159358, 50601, 4170, 51634, 0, 2527, 48220, 3136]
```
dshemetov, line 406: The sources for the FIPS county codes to names can be found [here](https://en.wikipedia.org/wiki/List_of_United_States_FIPS_codes_by_county#cite_note-CensusT-15) and cross-referenced [here](https://www.nrcs.usda.gov/wps/portal/nrcs/detail/national/home/?cid=nrcs143_013697) and [here](http://datadictionary.naaccr.org/default.aspx?c=11&Version=21#Virgin%20Islands%20of%20the%20United%20States). The population sources can be found at the Census [here](https://www.census.gov/data/tables/2010/dec/2010-island-areas.html). I double checked the numbers and all look good :+1:

dshemetov: While there are a couple other territories that have close aid relationships with the United States such as Marshall Islands, the Republic of Palau, and Micronesia, our data sources do not appear to report on these at the moment. Adding support for those can be done as needed in a separate PR.

dshemetov (approved):

krivard:
Pulling in some Reich lab folks to weigh in on deployment

krivard:
Backissues have been requested for JHU, which we can build with the commitlog tool. 

Safegraph has no backfill (it doesn't issue updates on past values) so we can probably just generate an issue per file, and compute the issue date as the file date + 4.

Once we deploy this we can let it run for a week and then do a database query to figure out what other indicators have data from Territories.

chinandrew:
realized i never moved this out draft form, will do that now.

krivard:
(just double-checking that everything still passes; a lot's happened in three weeks)

PR #669, opened 2020-12-29
Add HHS and Nation to combo cases

chinandrew:
### Description
This ones a bit tricky since the signal combines usafacts and JHU. What I ended up doing was having hhs and nation process as if they were fips and then geomap them to the corresponding regions at the end. One thing to note was that the state and county values dont sum up to the same thing, not sure if thats expected due to JHU/USAFacts processing):

```
In [5]: df = pd.read_csv("receiving/20201218_county_confirmed_cumulative_num.csv")                                                                                   

In [6]: df.val.sum()                                                                                                                                                 
Out[6]: 17103973.0

In [7]: df = pd.read_csv("receiving/20201218_state_confirmed_cumulative_num.csv")                                                                                    

In [8]: df.val.sum()                                                                                                                                                 
Out[8]: 17167997
```

### Changelog
- Add handling of hhs and nation conversion by adding having them retrieve the fips data and then geomap at the end. Update constants.

### Fixes 
- Fixes #(issue)

### DIFF
--- a/combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
+++ b/combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
@@ -15,8 +15,11 @@ import covidcast
 import pandas as pd
 
 from delphi_utils import read_params, add_prefix
+from delphi_utils.geomap import GeoMapper
 from .constants import METRICS, SMOOTH_TYPES, SENSORS, GEO_RESOLUTIONS
 
+
+GMPR = GeoMapper()
 
 def check_none_data_frame(data_frame, label, date_range):
     """Log and return True when a data frame is None."""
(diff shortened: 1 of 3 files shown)


combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
```diff
@@ -46,31 +49,45 @@ def maybe_append(df1, df2):
 def combine_usafacts_and_jhu(signal, geo, date_range, fetcher=covidcast.signal):
     """Add rows for PR from JHU signals to USA-FACTS signals."""
     print("Fetching usa-facts...")
-    usafacts_df = fetcher("usa-facts", signal, date_range[0], date_range[1], geo)
+    # for hhs and nation, fetch the county data so we can combined JHU and USAFacts before mapping
+    # to the desired geos.
+    geo_to_fetch = "county" if geo in ["county", "hhs", "nation"] else geo
```
krivard, line 54: "county" is not needed in that list. I have a slight impulse to extract to a constant set but it's probably more readable as is.

krivard (changes requested):
Add test for hhs or nation just to verify logic works as designed

krivard:
+tests?

chinandrew:
> +tests?

Just got around to updating them. Also ran the indicator and eyeballed the output files.

krivard (approved):
xlnt

PR #675, opened 2021-01-04
SirCAL: logger -> LOGGER

krivard:
### Description
Bug fix: `logger` should be `LOGGER`

### Changelog
- run.py


chinandrew:
uh oh why is hhs_facilities failing on a dtype check....

benjaminysmith (approved):
Thanks for fixing this.

PR #678, opened 2021-01-05
Set logging filename output for JHU indicator

benjaminysmith:
### Description
Set logging filename output for JHU indicator

### Changelog
- Adds filename to prod params

### Fixes 
- Progress on #606


ansible/templates/jhu-params-prod.json.j2
```diff
@@ -9,4 +9,5 @@
     "aws_secret_access_key": "{{ delphi_aws_secret_access_key }}"
   },
   "bucket_name": "delphi-covidcast-indicator-output"
```
korlaxxalrok, line 11: I think we need this comma.
```suggestion
  "bucket_name": "delphi-covidcast-indicator-output",
```

korlaxxalrok (approved):
See the one suggestion for the missing comma, but otherwise looks good to go. We are picking up everything in `/var/log/indicators/**` so this should work jammers 👍 :shipit:

ansible/templates/jhu-params-prod.json.j2
benjaminysmith: Good eye... done.

PR #679, opened 2021-01-06
Validation logs verbose errors to logger

sgsmob:
### Description
Results of validation are no longer all output to terminal, but split between the logger and the terminal.

### Changelog
- Detailed error messages about individual failed validations are logged at the appropriate level rather than printed to console


validator/delphi_validator/report.py
```diff
@@ -1,6 +1,9 @@
 """Validation output reports."""
 import sys
 from typing import Set, Tuple
+from delphi_utils.logger import get_structured_logger
+
+logger = get_structured_logger()
```
benjaminysmith, line 6: I recommend passing in __name__ here as well so the logger will have the name of this calling class associated.

get_structured_logger(__name__)

(ultimately we will also likely want a filename passed in to persist to disk, e.g., return get_structured_logger(__name__, filename = params.get("log_filename"))

validator/tests/test_report.py
```diff
@@ -34,5 +35,22 @@ def test_str(self):
...
+        report.increment_total_checks()
+        report.increment_total_checks()
+        report.increment_total_checks()
+        report.add_raised_warning(ImportWarning("wrong import"))
+        report.add_raised_warning(ImportWarning("right import"))
+        report.add_raised_error(self.ERROR_1)
+        report.add_raised_error(self.ERROR_2)
+
+        report.log()
+        mock_logger.critical.assert_called_once_with(
```
benjaminysmith, line 54: I don't feel too strongly, but generally don't feel its necessary to test logging statements (since it's mostly just duplication of the actual code and doesn't really verify state)

benjaminysmith (approved):

validator/tests/test_report.py
sgsmob: I feel that the logging here is important to test since it is part of the core functionality of the validator.

validator/delphi_validator/report.py
sgsmob: done, I think.

PR #684, opened 2021-01-12
Add more and better logging to JHU indicator

benjaminysmith:
### Description
Add more and better logging to JHU indicator

### Changelog
- Return list of dates written by the CSV exporter
- Log the # of CSV exports, oldest final date written, lag, and execution time for JHU indicator.

### Fixes 
- Improves on #606


jhu/delphi_jhu/run.py
```diff
@@ -4,8 +4,10 @@
 This module should contain a function called `run_module`, that is executed
 when the module is run with `python -m MODULE_NAME`.
 """
-from datetime import datetime
+from datetime import datetime as dt
 from itertools import product
+import datetime
```
sgsmob, line 9: I don't think this `import datetime` is needed.

sgsmob: So you can probably also delete the alias of `datetime.datetime` to `dt` if you would like.

sgsmob (approved):
LGTM modulo the quick comments.

jhu/delphi_jhu/run.py
benjaminysmith: I'm still using it on line 156 (datetime.datetime.now())

sgsmob: right but you have already imported `datetime.datetime` as `dt`

benjaminysmith: Oh nvm, I see what you mean. Removed.

PR #687, opened 2021-01-12
Push JHU logging to prod

benjaminysmith:
### Description
Push JHU logging to prod

### DIFF
--- a/jhu/delphi_jhu/run.py
+++ b/jhu/delphi_jhu/run.py
@@ -6,6 +6,7 @@ when the module is run with `python -m MODULE_NAME`.
 """
 from datetime import datetime
 from itertools import product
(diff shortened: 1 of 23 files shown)


jhu/delphi_jhu/run.py
```diff
@@ -137,3 +147,15 @@ def run_module():
         # Report failures: someone should probably look at them
         for exported_file in fails:
             print(f"Failed to archive '{exported_file}'")
```
krivard, line 149: Is mixing prints and logging here intentional?

benjaminysmith: It's intentional in that I didn't actively replace print statements yet... fine for now.

PR #693, opened 2021-01-13
Refactor validation parameters to hierarchical structure

### LINKED ISSUE #579: Alternate data hierarchy for validation parameters/members

sgsmob:
### Description
Validation configurations in `params.json` are now specified with a nested rather than flat structure.

### Changelog
- Validation parameters are split into "global", "static", and "dynamic"
- All non-global validation parameters have default values and are not required in `params.json` for the validation to function
- Updated documentation of module

### Fixes 
- Fixes #579


nmdefries (commented):
The code/documentation looks good, but I'm having a few issues with the tests.

The `freezegun` package isn't automatically installed in the environment, so running tests causes an `ImportError`; I was able to resolve this issue by adding `freezegun` to the `required` list in `setup.py`.

`TestDataFetcher.test_threaded_api_calls` is failing because of a type difference;

```
AssertionError: Attributes of DataFrame.iloc[:, 0] (column name="geo_id") are different

Attribute "dtype" are different
[left]:  float64
[right]: object

test_datafetcher.py:100: AssertionError
```

sgsmob:
Added the `freezegun` dependency, but I can't replicate the type difference error.

krivard:
Did you push? There doesn't seem to be an update to `setup.py`.

I'm able to replicate the type error. I'm going to try a few things to see if I can pin it down, or at least gather more information, and will update here.

sgsmob:
> Did you push? There doesn't seem to be an update to `setup.py`.
> 
> I'm able to replicate the type error. I'm going to try a few things to see if I can pin it down, or at least gather more information, and will update here.

Pushed to the wrong branch; should be updated now

sgsmob:
Replicated error on a clean virtualenv.  I think it has to do with an old version of the delphi utils.  Looking into it.

krivard:
For some reason, it's parsing the same "1044" string as a float in one instance and a (string) object in another.

Possible culprit in `datafetcher.fetch_api_reference`: does casing None to np.nan automatically en-numeric-ify a column?

```
    # Replace None with NA to make numerical manipulation easier.                                                                                                                                                                             
    # Rename and reorder columns to match those in df_to_test.                                                                                                                                                                                
    api_df = api_df.replace(
        to_replace=[None], value=np.nan
    ).rename(
        columns={'geo_value': "geo_id", 'stderr': 'se', 'value': 'val'}
    ).drop(
        ['issue', 'lag'], axis=1
    ).reindex(columns=column_names)
```

sgsmob:
Yes, that appears to be the problem.  FWIW it appears to be a problem on `main` and is thus unrelated to this PR.  I still can't replicate it on my normal environment which makes me wonder whether this is a bug introduced in pandas itself.  Will look further about how to handle it.

sgsmob:
bug is fixed (I think) by not doing the replace across geo values

nmdefries (approved):

PR #696, opened 2021-01-14
HHS hosp state add nation, hhs region geos

krivard:
### Description
We're putting off the facilitywise indicator for a bit, since it looks like forecast hub won't be doing anything below state. We may as well add nation and HHS region support before we release.

### Changelog
- Use GeoMapper to aggregate state to hhs and nation


hhs_hosp/delphi_hhs/constants.py
```diff
@@ -7,3 +7,9 @@
     CONFIRMED,
     SUM_CONF_SUSP
 ]
+
+# incoming geo: state
+GEOS = [
```
chinandrew, line 12: Since some of the other indicators use the geos variable to indicate all geos, not just additional ones, would it be more clear to rename this as ADDITIONAL_GEOS? Alternatively, could add state and deal with it in the code, or just write a longer comment about this

krivard: that's fair, will convert to all geos and handle state conditionally in code

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -319,7 +325,7 @@ def add_geocode(
             return df
         elif new_code == "nation":
             raise ValueError(
-                "Conversion to the nation level is only supported from the FIPS and ZIP codes."
+                f"Conversion to the nation level is not supported from {from_code}; try fips, zip, or state_*"
```
chinandrew, line 328: thanks, forgot to update these in my pr

hhs_hosp/delphi_hhs/run.py
```diff
@@ -79,39 +80,56 @@ def run_module():
...
-        create_export_csv(
-            make_signal(all_columns, sig),
-            params["export_dir"],
-            "state",
-            sig
-        )
+        state = geo_mapper.add_geocode(make_signal(all_columns, sig),
+                                       "state_id", "state_code",
+                                       from_col="state")
+        for geo in GEOS:
```
chinandrew, line 89: should tests be added for this logic? In this case I don't feel too strongly since they're just calling replace/add geocode and csv export which are both already tested, so going to approve regardless.

chinandrew (approved):
:shipit: :shipit: :shipit:

hhs_hosp/delphi_hhs/run.py
krivard: ...yes, especially considering how much time it took me to debug that section

Issue #689, opened 2021-01-13
Add logging to all indicators


benjaminysmith:

benjaminysmith:
Indicators done:
- USA facts
- JHU
- Safegraph
- Safegraph patterns
- Google health
- Google Symptoms
- NCHS mortality
- Combo (but not running in prod)
- Sircal

Remaining:
- FB
- Quidel
- Hospitalizations
- Dr visits

After adding basic logging, need to refactor the CSV count code and add to the indicators in next round.

benjaminysmith:
We are blocked on collecting data for:
FB, DV, HA

As these are not productionized yet.

We are blocked on adding the logging to FB as this is not checked in yet.

i.e., remaining work is to add logging code to DV and HA.

PR #700, opened 2021-01-15
Add logging to google health and symptoms.

benjaminysmith:
### Description
Add logging to google health and symptoms.

### Changelog
- Add log lines
- Update prod params for logfile location

### Fixes 
- Progress on #689


google_health/delphi_google_health/run.py
```diff
@@ -95,12 +102,20 @@ def run_module():
     for signal in signal_names:
         is_smoothed = signal.endswith(SMOOTHED)
         for geo_res, df in df_by_geo_res.items():
-            create_export_csv(format_for_export(df, is_smoothed),
+            exported_csv_dates = create_export_csv(
+                              format_for_export(df, is_smoothed),
```
sgsmob, line 106: I don't like the alignment here; shouldn't it just be one tab level under the "c" in "create_export_csv"?

google_symptoms/delphi_google_symptoms/run.py
```diff
@@ -47,10 +59,29 @@ def run_module():
...
+    max_lag_in_days = None
+    formatted_oldest_final_export_date = None
+    if oldest_final_export_date:
+        max_lag_in_days = (datetime.now() - oldest_final_export_date).days
+        formatted_oldest_final_export_date = oldest_final_export_date.strftime("%Y-%m-%d")
+    logger.info("Completed indicator run",
+        elapsed_time_in_seconds = elapsed_time_in_seconds,
+        csv_export_count = csv_export_count,
+        max_lag_in_days = max_lag_in_days,
+        oldest_final_export_date = formatted_oldest_final_export_date)
```
sgsmob, line 87: This whole block seems to be repeated frequently within and across indicators.  Is there a good way to make this a function or something else reusable?

sgsmob, line 87: Add trailing newline

google_health/delphi_google_health/run.py
benjaminysmith: Done

google_symptoms/delphi_google_symptoms/run.py
benjaminysmith: Done

benjaminysmith: Definitely. The wrinkle is that some of the indicators use quite different setups (see e.g., change), so waiting to refactor after I've worked out how to handle those (it may be better to push this all the way into the csv_export function, for example)

I'm not too concerned with copypasting the boilerplate for now since it's just logging code.

sgsmob (approved):

PR #701, opened 2021-01-18
Fix usafacts params

korlaxxalrok:
### Description
Fix:

Add a missing comma that was causing the indicator run to fail.

### Changelog
- Add comma to params file.


capnrefsmmat (approved):
I may not know much about the production setup, but that sure looks like a comma to me

PR #704, opened 2021-01-19
Fix quidel covidtest newformat

jingjtang:
### Description
update code to read quidel raw data from aws s3 bucket
All historical files are re-uploaded, and stored in csv format

### Changelog
Itemize code/test/documentation changes and files added/removed.
- Update `pull.py`, `params.json.template`
- Change test data from xlsx to csv

### Fixes 
- Fixes #(issue)

PR #706, opened 2021-01-19
Fix quidel covidtest newformat (off main)

krivard:
### Description
Change quidel to fetch data from an Amazon S3 bucket and comply with new input data format.

Based on https://github.com/cmu-delphi/covidcast-indicators/pull/704, which branched off the outdated `quidel_covidtest`. I've now deleted the old branch.

I've reviewed the code and found it reasonable, and tests pass, so should be good to go.

### Changelog

- pull now fetches from S3 using boto
- pull now processes updated data format
- tests use updated data format
- **@krivard edits: update documentation** -- @jingjtang to verify all correct


jingjtang (approved):
lgtm

PR #709, opened 2021-01-19
Fix inconsistent lag calculation in SirCAL

### LINKED ISSUE #523: Make Sir Complains-a-lot log more details about gaps
Sir Complains-a-lot includes only the signal name, geo level, and date in its complaints, to keep them short enough to be readable by a human.

However, if it finds an issue that can then not be confirmed when a human investigates, there's not much that can be done.

Sir Complains-a-lot should generate a verbose log that includes the following information:
* The endpoint URL
* A sample of the data around any gaps identified, including the issue column

benjaminysmith:
### Description
Fix inconsistent lag calculation in SirCAL

### Changelog
- Use current date instead of metadata max date to avoid problems with metadata delays
- Add more logging to lag and date retrieval
- Fix the lag calculation by subtracting 1 (e.g., no gaps should be gap = 0 days, not gap = 1 day)

### Fixes 
- Fixes #523

### DIFF
--- a/sir_complainsalot/delphi_sir_complainsalot/check_source.py
+++ b/sir_complainsalot/delphi_sir_complainsalot/check_source.py
@@ -99,6 +99,10 @@ def check_source(data_source, meta, params, grace, logger):
             geo_type=row["geo_type"]
         )
 
+        if latest_data is None:
+            logger.info("No signal data retrieved")
+            continue
+
         # convert numpy datetime values to pandas datetimes and then to
         # datetime.date, so we can work with timedeltas after
         unique_dates = [pd.to_datetime(val).date()
--- a/sir_complainsalot/delphi_sir_complainsalot/check_source.py
+++ b/sir_complainsalot/delphi_sir_complainsalot/check_source.py
@@ -3,6 +3,7 @@ from typing import List
 from delphi_utils import get_structured_logger
 
 import covidcast
+from datetime import datetime, timedelta
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


sir_complainsalot/delphi_sir_complainsalot/check_source.py
```diff
@@ -87,26 +88,36 @@ def check_source(data_source, meta, params, grace, logger):
...
-            start_day=row["max_time"] - gap_window,
-            end_day=row["max_time"],
+            start_day=datetime.now() - timedelta(days = 14),
+            end_day=datetime.now(),
             geo_type=row["geo_type"]
         )
 
         # convert numpy datetime values to pandas datetimes and then to
         # datetime.date, so we can work with timedeltas after
         unique_dates = [pd.to_datetime(val).date()
```
chinandrew, line 104: i know this isn't affected in this PR (though the unique issues is), but if this is unique values, should it be wrapped in `set()`?

benjaminysmith: I feel like that only really matters if we were to add more values to it (to make sure that it really stayed a unique set) -- since we are using unique(), the list should still fullfill the contract of the variable name. wydt?

chinandrew: oh i didnt even see the unique(), that makes sense to me. is unique() guaranteed to return them in the right order? Based on the gap_days computation line, seems like you want them monotonically increasing.

krivard: unique() is order-preserving, so if they start out ordered, they'll stay that way

chinandrew (approved):
:shipit::shipit::shipit:

sir_complainsalot/delphi_sir_complainsalot/check_source.py
chinandrew: :+1:

benjaminysmith:
Please don't merge yet -- noticed an edge case I need to fix...

benjaminysmith:
Added. @krivard please merge if it looks ok.

The additional check I added was for None coming back from API (since we are no longer using metadata for the last date this means for very out of date signals we might get no data back using the current date as the endpoint)

krivard (approved):
👍

PR #711, opened 2021-01-20
Add Quidel Ansible and params config for accessing the S3 bucket

korlaxxalrok:
### Description
Feature:

Adds the necessary Ansible vars/vault entries plus the prod params config so
that the indicator can interact with the S3 bucket.

### Changelog
- Update prod params.
- Add Ansible bucket name variable.
- Add Ansible variable pointers to vault.
- Add Ansible Vault variables.


korlaxxalrok:
@jingjtang I also removed the email settings. I made an assumption that we aren't checking the mail account anymore, but do those need to stay in the params?

jingjtang:
> @jingjtang I also removed the email settings. I made an assumption that we aren't checking the mail account anymore, but do those need to stay in the params?

The email settings are not needed anymore. They should be removed from `params` settings.

korlaxxalrok:
@jingjtang If it looks good to you I think this is ok to merge now.

jingjtang (approved):
lgtm

PR #712, opened 2021-01-20
Allow validation to accept nation/hhs geo types

sgsmob:
### Description
Fixes bugs in validation that made it unable to correctly validate data at the nation and HHS geo resolution.

### Changelog
- Change references to "national" geo resolution to the proper term, "nation"
- Make HHS a valid geo type for validation
- Fix small bug where we accidentally print the `ValidationReport.summary` method rather than the results of calling `ValidationReport.summary()` ... oops


validator/delphi_validator/utils.py
```diff
@@ -8,10 +8,11 @@
 GEO_REGEX_DICT = {
     'county': r'^\d{5}$',
     'hrr': r'^\d{1,3}$',
+    'hhs': r'^\d{1,2}$',
```
chinandrew, line 11: why is there both regex and a static file?

validator/tests/test_static.py
```diff
@@ -256,6 +265,15 @@ def test_invalid_geo_id_hrr(self):
         assert len(report.raised_errors) == 1
         assert report.raised_errors[0].check_name == "check_bad_geo_id_value"
 
+    def test_invalid_geo_id_hhs(self):
```
chinandrew, line 268: duplicate test name of line 155

validator/delphi_validator/utils.py
sgsmob: Not sure on the original design but it appears we are checking bad formatting vs. bad values but I agree it seems like overkill.

chinandrew: got it. does it matter if the column is read in as int vs string vs zero padded string (e.g. 1, '1', '01')?

validator/tests/test_static.py
sgsmob: Changed naming convention to make the test names globally unique

validator/delphi_validator/utils.py
sgsmob: geo ids are read as strings so 1 vs. '1' is not relevant.  The formatting will pass for '01' but the value will fail; we could change that to be a formatting failure (require the first digit to be non-zero).

sgsmob: I have no real preference for how this is done.  It would be nice to only have to do value comparison and ignore formatting but it appears that we are guarding against geo ids that have encoded as ints or floats rather than strings.  It is a plausible design to trigger a failure if the geo ids do not match the strings due to a numeric conversion rather than try to reconstruct to the underlying string value.  @nmdefries might have some insight into the latitude we have here to be strict/lenient

chinandrew: makes sense. The geomapper doesnt add leading 0's so i think this is fine for now.

chinandrew (approved):
:shipit: :shipit: :shipit:

validator/delphi_validator/utils.py
nmdefries: >  it appears that we are guarding against geo ids that have encoded as ints or floats

This is a nice-to-have. We want geo ids to be strings as the gold standard; ints and floats are tech debt but don't cause any actual problems besides inconsistent formatting. Value comparison is certainly more important.

PR #714, opened 2021-01-20
Add runtime logging to NCHS mortality, SirCAL, combo

benjaminysmith:
### Description
Add runtime logging to NCHS mortality, SirCAL, combo

### Changelog
- Add logging
- Update params

### Fixes 
- Progress on #689


krivard (changes requested):
Would you add the same changes to the params.json.template files as well?

benjaminysmith:
> Would you add the same changes to the params.json.template files as well?

Done.

krivard (approved):
merci

PR #716, opened 2021-01-21
Fix nchs schema

krivard:
### Description
NCHS has been crashing due to a change in column names at the source. This PR shifts to the new column names, but includes rename mappings so it can still run on old files if needed.

### Changelog

- run.py -- don't archive if bucket name isn't specified (needed for running indicator outside of production)
- pull.py -- column standardization routine, alternate consistency checks for new schema
- tests -- use column standardization routine


nchs_mortality/delphi_nchs_mortality/constants.py
```diff
@@ -20,3 +26,4 @@
 ]
 INCIDENCE_BASE = 100000
 GEO_RES = "state"
+NEWLINE = "\n"
```
sgsmob, line 29: All else equal, I am against macro-ing this.  Based on the use cases in the other file, I would prefer to just put the literal "\n" wherever `NEWLINE` is.

nchs_mortality/delphi_nchs_mortality/pull.py
```diff
@@ -3,7 +3,18 @@
...
+from .constants import METRICS, RENAME, NEWLINE
+
+def standardize_columns(df):
+    """Rename columns to comply with a standard set.
+
+    NCHS has changed column names a few times, so this will help us maintain
+    backwards-compatibility without the processing code getting all gnarly.
+    """
+    rename_pairs = [ (from_col, to_col) for (from_col, to_col) in RENAME
+                     if from_col in df.columns ]
```
sgsmob, line 15: I prefer not having spaces between the body of the list and the brackets.

nchs_mortality/delphi_nchs_mortality/pull.py
```diff
@@ -42,25 +53,44 @@ def pull_nchs_mortality_data(token: str, map_df: pd.DataFrame, test_mode: str):
...
-    except AssertionError as exc:
-        raise ValueError(
-            "end_week is not always the same as start_week, check the raw file"
-        ) from exc
+    df = standardize_columns(df)
+
+    if "end_date" in df.columns:
+        # Check missing week_ending_date == end_date
+        try:
+            assert sum(df["week_ending_date"] != df["end_date"]) == 0
```
sgsmob, line 67: If I understand what is going on here, it seems it would be better to use `all()` or `any()` rather than `sum()`.

nchs_mortality/delphi_nchs_mortality/pull.py
```diff
@@ -42,25 +53,44 @@ def pull_nchs_mortality_data(token: str, map_df: pd.DataFrame, test_mode: str):
...
+        except AssertionError as exc:
+            raise ValueError(
+                "week_ending_date is not always the same as end_date, check the raw file"
+            ) from exc
+    else:
+        # Check missing start_week == end_week
+        try:
+            assert sum(df["timestamp"] != df["end_week"]) == 0
```
sgsmob, line 75: same

nchs_mortality/tests/test_pull.py
```diff
@@ -5,7 +5,7 @@
 import pandas as pd
 from delphi_utils import read_params
 
-from delphi_nchs_mortality.pull import pull_nchs_mortality_data
+from delphi_nchs_mortality.pull import pull_nchs_mortality_data, standardize_columns
```
sgsmob, line 8: Would be nice to have an explicit test for `standardize_columns()`

sgsmob (changes requested):

nchs_mortality/delphi_nchs_mortality/constants.py
krivard: [backslash expressions aren't allowed in f-strings, alas](https://www.python.org/dev/peps/pep-0498/#escape-sequences)

sgsmob: Can you please add a comment to explain why are aliasing this here then?

nchs_mortality/tests/test_pull.py
```diff
@@ -19,18 +19,35 @@
 )
 
 class TestPullNCHS:
+    def test_standardize_columns(self):
+        df = standardize_columns(
+            pd.DataFrame({
+                "start_week":[1],
```
sgsmob, line 25: Nit:  leave a space after the colons here and in the next two lines.

sgsmob (approved):
Just one minor comment but otherwise LGTM

PR #717, opened 2021-01-21
Add logging to Quidel and Change indicators

benjaminysmith:
### Description
Add logging to Quidel and Change indicators

### Changelog
- Add logging
- Update params

### Fixes 
- Progress on #689

### DIFF
--- a/quidel/delphi_quidel/run.py
+++ b/quidel/delphi_quidel/run.py
@@ -7,7 +7,13 @@ when the module is run with `python -m MODULE_NAME`.
 from os.path import join
 
 import pandas as pd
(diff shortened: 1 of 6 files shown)


quidel/delphi_quidel/run.py
```diff
@@ -21,7 +27,9 @@
 
 def run_module():
     """Run Quidel flu test module."""
+    start_time = time.time()
```
chinandrew, line 30: this seems like something that could be a decorator, though not sure quite how to get around the logger being initialized within the function....

chinandrew (approved):
:shipit: :shipit: :shipit:

Issue #680, opened 2021-01-06
Add PCR tests signal from COVID Act Now


krivard:
To start out with we'd like to mirror and reaggregate PCR testing information. Some research into CAN's available metrics and methodology will be required to determine whether this is actually possible; a cursory look through their documentation suggests that they do not actually specify what kinds of tests are included in their metrics, or whether they count number of tests or number of people tested. Both will need to be confirmed before we start writing code.

cc @tildechris 

**Data details**
 
https://apidocs.covidactnow.org/

**Additional context**

CAN's API terms of use specifies a CC BY-NC-ND license, which would normally prevent reaggregation. However, we have received special permission from CAN to produce derivative data streams from their data so long as we refrain from commercial use. Essentially we've been given a CC BY-NC license, which fits easily within COVIDcast's established data licenses.

eujing:
Current update for the nature of their test positvity metrics:

Based on previous discussions in the #can-delphi channel, there are two possible ways for obtaining test posivity metrics from CAN, either from:
1) Their API directly as `testPosivityRatio`, or 
2) Their latest parquet file using the `pcr_test_*` test result count columns

Each obtains the metrics from a mixture of sources for greater geo coverage, and also explicity specifies a `source` and `provider` for each unit of time and geo.

 `testPositivityRatio` from their API has the following sources (in descending geo coverage):
1) CDCTesting: PCR and specimen-based
2) Valorum: Not confirmed yet
3) Others: Not confirmed yet
4) CMSTesting: PCR and specimen-based

Parquet file `pcr_test_*` columns have the following `providers` (in descending geo coverage). The first two are used for county-level metrics, CTP is used for state-level metrics:
1) CDC: PCR and specimen-based
2) State: Not confirmed yet
3) Covid Tracking Project: CTP has quite a detailed breakdown of PCR vs Antibody and Specimen vs People counts, but still need to confirm if specimen-based counts are used here.

It seems that majority of the metrics are PCR and specimen-based, but I am currently waiting for confirmation from CAN on the remaining sources. From my current understanding, Valorum and State are based on state dashboard scrapers, so it might be harder to confirm for those.

tildechris:
Our intended measure is percent of positive PCR tests vs. all PCR tests right?  So, we need to confirm we're only looking at PCR on a specimen-basis?

eujing:
Yeah, we need to confirm if all the sources are PCR on a specimen-basis

PR #720, opened 2021-01-25
COVID Act Now PCR-specimen signals

eujing:
### Description
Generates PCR-specimen test positivity rates from CAN / CDC along with the testing volume in the signal `pcr_specimen_positivity_rate`.
The raw data is pulled from CAN's latest parquet file.
The positivity rate is `val` while the testing volume is `sample_size`.
The data is originally at the county level, and is aggregated to state, msa, hrr, hhs and nation levels.

### Notes
1) So far it seems the best time to run the indicator is a bit after 8 pm ET as the CDC makes changes throughout the day up till then.
2) However, not all counties have data up to the latest date available in the data. The data for these counties seem to be padded as 0% positivity rate and 0 total tests. One good example would be California, where many counties seem to get updates less frequently.
3) CDC only provides the smoothed 7-day rolling averages of positivity rate and testing volume, not the raw values. The aggregations are currently done by multiplying the smoothed rate with the smoothed total to approximate the counts of positive tests. The counts are then summed in the aggregations, with no further smoothing applied afterwards.

### Fixes 
- For #680


covid_act_now/delphi_covid_act_now/geo.py
```diff
@@ -0,0 +1,94 @@
...
+        Columns: val, sample_size, ...
+
+    Returns
+    -------
+    pd.Series
+        Standard error of the positivity rate of PCR-specimen tests.
+    """
+    p = x.val
+    n = x.sample_size
+    return np.sqrt(p * (1 - p) / n)
```
krivard, line 48: Won't this generate `inf` when sample_size is zero?

We've been using a Jeffreys prior for sampling-based indicators with occasional 0 sample sizes. The rule is you add a pseudocount of 1/2 to each possible outcome. This bumps `n` up to 1, so you're no longer dividing by zero.

[Here's an example in the changehc indicator](https://github.com/cmu-delphi/covidcast-indicators/blob/127c97643d3ee6ad0441f039131653d092249284/changehc/delphi_changehc/sensor.py#L133)

You'd apply the jeffrey's prior to the statistic and to the standard error.

covid_act_now/tests/test_geo.py
```diff
@@ -0,0 +1,94 @@
...
+    def test_std_err(self):
+        df = pd.DataFrame({
+            "val": [0, 0, 0.5, 0.4, 0.3, 0.2, 0.1],
+            "sample_size": [0, 2, 2, 5, 10, 20, 50]
+        })
+
+        expected_se = np.sqrt(df.val * (1 - df.val) / df.sample_size)
+        se = std_err(df)
+
+        assert np.allclose(se, expected_se, equal_nan=True)
```
krivard, line 34: Add an assert here that `se` should always be a positive number, never nan, inf, or 0

covid_act_now/tests/conftest.py
```diff
@@ -0,0 +1,171 @@
...
+            remove(join("receiving", fname))
+
+    with mock_s3():
+        # Create the fake bucket we will be using
+        params = read_params()
+        aws_credentials = params["aws_credentials"]
+        s3_client = Session(**aws_credentials).client("s3")
+        s3_client.create_bucket(Bucket=params["bucket_name"])
+
+        run_module()
```
krivard, line 28: Is this also mocking the network call to the CAN API? If not, it should -- tests shouldn't hit the network.

krivard (changes requested):
Some fixes, some questions, but an excellent first draft

covid_act_now/delphi_covid_act_now/pull.py
```diff
@@ -0,0 +1,79 @@
...
+            variable_name.str.startswith('pcr_tests_')
+            """)
+        .pivot(index=["fips", "timestamp"], columns="variable_name", values="value")
+        .reset_index()
+        # pcr_tests_positive from the CDC is actually positivity rate (percentage)
+        .rename(columns={"pcr_tests_positive": "pcr_positivity_rate"})
+    )
+
+    df_tests["pcr_positivity_rate"] /= 100
+    df_tests["pcr_tests_positive"] = df_tests.pcr_positivity_rate * df_tests.pcr_tests_total
```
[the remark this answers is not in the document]
eujing: In the case of the county level signal, that is true I do just use the original positivity rate from the source.
However I found the counts to make it easier to use our sum-based geo aggregation utils to the higher geo levels.
I think I could equivalently do this with a mean of the rates weighted by the total tests in each county, but was trying to stick to the utils in its current form as much as possible.

covid_act_now/delphi_covid_act_now/geo.py
eujing: Yeah the `sample_size` 0 is quite a common issue! This has been producing NaN for the `se` in the current form I think because of the `sqrt`.

Especially for the most recent dates, many counties have a sample_size 0 with 0% positivity rate to testing data not being known yet.

I was thinking of using Jeffreys prior, but was unsure if we are fine with the positivity rate being 0.5 for such cases. Should I go ahead and use Jeffrey's prior anyway?

covid_act_now/tests/conftest.py
eujing: In the tests, I load a local parquet file `tests/test_data/small_CAN_data.parquet` instead of CAN's actual URL through the `params.json` file.
That file should have the exact same data as in the `CAN_parquet_data` fixture in `conftest.py`
I do feel this is quite duplicated, and maybe I should write `CAN_parquet_data` to a tmp directory and modify the test `params.json` before running the module?

covid_act_now/delphi_covid_act_now/geo.py
krivard: we should run it by Chris and a stats advisor to make sure, but I think so -- 0.5 with a large standard error is what you'd expect if you were guessing without any data to go off of.

krivard: alternately we could suppress reporting for 0 sample size

covid_act_now/tests/conftest.py
krivard: ah! i see the tests params template now. all good!

covid_act_now/delphi_covid_act_now/geo.py
krivard: Hey @tildechris thoughts? CAN publishes records with 0 sample size, which is not ideal from a statistical standpoint. How should we handle that?

* suppress the zeros
* mirror the zeros (which will be awkward to validate)
* use Statistics™️ on the zeros (which will make them 0.5 with a hefty error bar)

tildechris: Hi @krivard, I met with @eujing on this topic and it seems that the least surprising thing to do would be to simply drop entries which are 0 sample size.

If 0 sample size means that there was a lack of observation as opposed to knowing that 0 tests were performed, I think this is also semantically correct.  Even in the case where there is an observation which is 0 ("we didn't do any tests" vs "i don't yet know how many tests we did"), since we're ultimately trying to get to percent positivity, I think that dropping is still the correct thing to do.

krivard: Sounds good to me -- thanks!

krivard (changes requested):
Add a test for pull.extract_testing_metrics to make sure it removes entries with 0 sample size, and then I think we're all set

covid_act_now/tests/test_geo.py
```diff
@@ -0,0 +1,115 @@
...
+    def test_std_err(self):
+        df = pd.DataFrame({
+            "val": [0, 0.5, 0.4, 0.3, 0.2, 0.1],
+            "sample_size": [2, 2, 5, 10, 20, 50]
+        })
+
+        expected_se = np.sqrt(df.val * (1 - df.val) / df.sample_size)
+        se = std_err(df)
+
+        assert (se >= 0).all()
```
krivard, line 34:
```suggestion
        # 0 se is permitted in this indicator, since applying the Jeffreys prior would violate the mirror
        assert (se >= 0).all()
```

krivard (approved):
👍

Issue #628, opened 2020-12-11
Have Geomapper accept "county" for "fips" conversions


chinandrew:
To convert something to counties, the geomapper requires the argument be  `'fips'` and not `'county'`, which makes it hard to do something like

```
for geo in georesolutions:
    geomapper.add_geocode(df, fromcode, geo,....)
```
since we need `'county'` to exist in the output filenames, you end up having to do a condition statement or use a dict to map `'county'` to `'fips'`

PR #722, opened 2021-01-25
Changehc: get region count from GeoMapper instead of config.Constants

### LINKED ISSUE #721: Make changehc use the geomapper's list of counties instead of its own constant
`changehc` is currently crashing on the files from 23 January because the number of FIPS codes exceeds a hard-coded threshold which is incorrect.

```
    exec(code, run_globals)
  File "/home/indicators/runtime/changehc/delphi_changehc/__main__.py", line 11, in <module>
    run_module()  # pragma: no cover
  File "/home/indicators/runtime/changehc/delphi_changehc/run.py", line 150, in run_module
    su_inst.update_sensor(
  File "/home/indicators/runtime/changehc/delphi_changehc/update_sensor.py", line 190, in update_sensor
    data_frame = self.geo_reindex(data)
  File "/home/indicators/runtime/changehc/delphi_changehc/update_sensor.py", line 167, in geo_reindex
    assert (len(multiindex) <= (Constants.MAX_GEO[geo] * len(self.fit_dates))
AssertionError: more loc-date pairs than maximum number of geographies x number of dates
```

```
[indicators@delphi-master-prod-01 changehc]$ zcat cache/20210123_*_Count_By_County.dat.gz | cut -d '|' -f 1 | sort -u | wc -l
3228
[indicators@delphi-master-prod-01 changehc]$ grep NUM_COUNTIES delphi_changehc/config.py 
    NUM_COUNTIES = 3141 + 52
    MAX_GEO = {"county": NUM_COUNTIES,
```

There are in fact 3,242 counties and county equivalents in the United States, if you include Puerto Rico and other territories, which Change does.

Rather than update the hard-coded constant, `changehc` should retrieve this value from GeoMapper.

krivard:
### Description
Make changehc use GeoMapper to get the region count instead of using a hardcoded value.

### Changelog
- Let GeoMapper help with converting "state" and "county" to mappable region types
- Use GeoMapper for region counts
- Fix tests

### Fixes 
- Fixes #721

### DIFF
--- a/_delphi_utils_python/delphi_utils/geomap.py
+++ b/_delphi_utils_python/delphi_utils/geomap.py
@@ -528,14 +528,14 @@ class GeoMapper:  # pylint: disable=too-many-public-methods
         data = data.reset_index().groupby([date_col, mega_col]).sum()
         return data.reset_index()
 
-    def as_mapper_name(self, geo_type):
+    def as_mapper_name(self, geo_type, state="state_id"):
         """
         Return the mapper equivalent of a region type.
 
         Human-readable names like 'county' will return their mapper equivalents ('fips').
         """
         if geo_type == "state":
-            return "state_code"
+            return state
         if geo_type == "county":
             return "fips"
         return geo_type
--- a/_delphi_utils_python/delphi_utils/geomap.py
+++ b/_delphi_utils_python/delphi_utils/geomap.py
@@ -528,6 +528,17 @@ class GeoMapper:  # pylint: disable=too-many-public-methods
         data = data.reset_index().groupby([date_col, mega_col]).sum()
         return data.reset_index()
 
+    def as_mapper_name(self, geo_type):
+        """
+        Return the mapper equivalent of a region type.
+
+        Human-readable names like 'county' will return their mapper equivalents ('fips').
+        """
+        if geo_type == "state":
+            return "state_code"
+        if geo_type == "county":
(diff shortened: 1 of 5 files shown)


_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -528,6 +528,17 @@ def fips_to_megacounty(
         data = data.reset_index().groupby([date_col, mega_col]).sum()
         return data.reset_index()
 
+    def as_mapper_name(self, geo_type):
```
chinandrew, line 531: Are we sure "state" -> "state_code" is always the intended mapping?

krivard: hmmm it should probably be state_id by default; i always get that wrong.

Added an override. For this use case it doesn't matter, but for others it might.

chinandrew: got it. what do you think of making as_mapper_name internal to geomapper so the mapping functions natively accept "state" and "county" (would address #628), and geomapper users dont need to worry about using it? Something like calling it as the first line to the add/replace/get geocode values. I think the argument against is that it's a bit less explicit and you cant easily choose something like  'state' -> 'state_name'. Don't feel that strongly so will go ahead and approve this for now.

chinandrew (approved):
:shipit::shipit::shipit:

PR #731, opened 2021-01-26
hhs hosp state daily: makefile and production params

krivard:
### Description
Adds a Makefile target for running the indicator, and updates the production params to include the AWS credentials.

The `delphi_hhs` folder has been added to `delphi-covidcast-indicator-output` in S3.

### Changelog
- Makefile - new target
- params.json.template - new keys (blanked for dev environments)
- ansible template params - new keys (with templated constants for credentials)

### DIFF
--- a/ansible/templates/hhs_hosp-params-prod.json.j2
+++ b/ansible/templates/hhs_hosp-params-prod.json.j2
@@ -2,5 +2,12 @@
   "static_file_dir": "./static",
   "export_dir": "/common/covidcast/receiving/hhs",
   "cache_dir": "./cache",
-  "wip_signal": ""
+  "wip_signal": "",
+  "aws_credentials": {
+    "aws_access_key_id": "{{ delphi_aws_access_key_id }}",
+    "aws_secret_access_key": "{{ delphi_aws_secret_access_key }}"
+  },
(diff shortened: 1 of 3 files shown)


chinandrew (commented):
with this change, will automation run `make run` instead of the usual `python -m ...`?

Not familiar with the archive differ details, but is there anything that determines if it's in the pipeline code vs run after like in this case?

krivard:
> will automation run make run instead of the usual python -m ...

That's correct; the `cmd` entry for this step is 

```
sudo -u indicators -s bash -c "cd /[absolute/path/to]/hhs_hosp && make run"
```

> is there anything that determines if it's in the pipeline code vs run after like in this case?

No. We are currently in the middle of a long development process for a modular indicator runner that will coordinate running the indicator, running the validator, running the archive differ, and dropping the files off in `receiving`. Including the archive differ in the indicator code itself has created problems for folks trying to run indicators outside the production environment, so rather than add that code, plus conditional handling for running without AWS keys, and then having to remove that code later when Indicator Runner is ready, this seemed less of a hassle

chinandrew (approved):
:shipit::shipit::shipit:

PR #732, opened 2021-01-26
Remove public signal check from signal prefix

chinandrew:
### Description
Removes footgun for wip signals by having the add_prefix signal not auto-add "wip_" to non-public signals.

`grep -r '"wip_signal":' *` shows everything is `"wip_signal": "",` or `"wip_signal": [""],` in the repo, so it's hard to say if any signals would be affected by this change. I had to update some cdc covidnet changes, not sure if those are actually wip and we need to add it to the params json

### Changelog
- Remove public_signal() function
- Have add_prefix just return the input signals if wip_signals is false or ''
- Fix cdc covidnet tests.

### Fixes 
- The OKRs "Remove footgun in WIP signal rename widget"


krivard (approved):
lgtm

covidnet is not scheduled in production, so we won't get any accidental releases there.

PR #735, opened 2021-01-27
Tweak SirCAL age check for combo to reduce alerts.

benjaminysmith:
### Description
Tweak SirCAL age check for combo to reduce alerts.

### Changelog
- Combo max age 3 -> 4 days.


krivard (approved):
👍

PR #737, opened 2021-01-27
Add symptom survey item V9 ("concerned about vaccine side effects") to API

nmdefries:
### Description
Response codes "Very concerned" and "Moderately concerned" in item V9 are mapped to `1`. The resulting variable is used to produce a weighted and an unweighted binary response.

### Changelog
- Binary variable definition added to `variables::code_vaccines`
- Aggregates added to `binary::get_binary_indicators`


capnrefsmmat (approved):
Looks good to me

PR #742, opened 2021-01-28
Remove log_filename from json template

benjaminysmith:
### Description
Remove log_filename from json template


sgsmob (approved):
I assume there is a good reason for doing this so I am approving this PR.

krivard:
Yep, the /var/log/indicators paths don't belong in the dev params, only in production

PR #743, opened 2021-01-28
Add "acceptance or already vaccinated" signal

capnrefsmmat:
For use in the survey dashboard as a replacement for the acceptance signal, particularly as more people get vaccinated.


nmdefries (approved):
Looks good!

PR #746, opened 2021-01-28
Explicitly log current lag by signal

benjaminysmith:
### Description
Explicitly log current lag by signal

### Changelog
- Add explicit log calculation to sircal for lag dashboard

### Fixes 
- Progress on #689


chinandrew (approved):
:shipit::shipit::shipit:

PR #748, opened 2021-01-28
Disable logging exceptions by default in dev

### LINKED ISSUE #747: Disable structured log exception logging in dev
It's hard to read when manually debugging code

benjaminysmith:
### Description
Disable logging exceptions by default in dev

### Changelog
- Explicitly enable exceptions in logger call
- Add param in dev envs to override this (to make them prettier when developing)

### Fixes 
- Fixes #747


changehc/delphi_changehc/run.py
```diff
@@ -80,7 +80,9 @@ def run_module():
     start_time = time.time()
     params = read_params()
 
-    logger = get_structured_logger(__name__, filename = params.get("log_filename"))
+    return get_structured_logger(
```
krivard, line 83:
```suggestion
    logger = get_structured_logger(
```

claims_hosp/delphi_claims_hosp/run.py
```diff
@@ -22,7 +22,9 @@ def run_module():
     """Read from params.json and generate updated claims-based hospitalization indicator values."""
     start_time = time.time()
     params = read_params()
-    logger = get_structured_logger(__name__, filename = params.get("log_filename"))
+    logger = get_structured_logger(
+        __name__, filename=params.get("log_filename"),
+        log_exceptions=params.get("log_exceptions", True))
```
krivard, line 27: These parameter names are copy-pasted 13 times in the code -- at what point do we switch to just passing in `params` so that it only has to be written once and the names are always the same by default?

krivard (changes requested):
One definite fix, one design question

claims_hosp/delphi_claims_hosp/run.py
benjaminysmith: I went back and forth a bit on this. The benefit is that it means if we add more params we only need to change the logger code. The drawback is that it ties the logger directly to the parameters (which means we can't generalize it as easily because that locks any other consumers into using the same parameters -- potentially even across codebases if we extract into a package.

I think the way to get around this is to make a dedicated function to get the logger from params which can call the one we have now (i.e., the params -> parameter mapping is in a single location, but can be easily extracted from the logger).

Since I already have this working I would lean toward keeping as is for now and refactoring when we change it next. wdyt?

changehc/delphi_changehc/run.py
benjaminysmith: wow good catch. I was surprised this was not caught by tests but it looks like run.py is not tested for Change

claims_hosp/delphi_claims_hosp/run.py
krivard: there's always
```
def get_structured_logger(name, filename=None, log_exceptions=None, params=None):
    assert (filename is not None) or (params is not None)
```

but I see your point -- I'm okay with waiting to refactor until the next change.

krivard (approved):
👍

PR #750, opened 2021-01-28
Include some decimal responses as valid for symptom survey item A2

nmdefries:
### Description
Include as valid decimal A2 responses with no actual decimal value (e.g. `1.0` and `1.` are valid vs `1.03` is invalid).

### Changelog
- Read A2 as character
- Convert A2 to integer if it meets formatting requirements; invalid decimal responses are set to NA
- Imports a few additional functions from `dplyr`

### Fixes 
- Fixes failing `test-gold-individual` as referenced [here](https://github.com/cmu-delphi/covidcast-indicators/commit/4d3e13820b93a201e295531413e4b0738372ab6a).


nmdefries:
I'm realizing that `col_double` has limited precision (e.g. "5.0000000000000000000001" is read in as 5). I'm going to switch this to a regex approach.

facebook/delphiFacebook/R/responses.R
```diff
@@ -31,7 +31,7 @@ load_responses_all <- function(params) {
 #'
 #' @importFrom stringi stri_split stri_extract stri_replace_all stri_replace
 #' @importFrom readr read_lines cols locale col_character col_integer
-#' @importFrom dplyr arrange desc case_when
+#' @importFrom dplyr arrange desc case_when mutate if_else
```
capnrefsmmat, line 34: don't need `if_else`? `ifelse` is in base

capnrefsmmat (changes requested):
looks good, just one fix. the existing gold test cover this case, right?

facebook/delphiFacebook/R/responses.R
nmdefries: Whoops, meant to use `if_else` there. Thanks for the catch.

Re: tests, yep.

capnrefsmmat (approved):
looks good. I didn't know about `if_else`; should be useful

Issue #725, opened 2021-01-25
Set up validation for all indicators


sgsmob:
Indicators:
- [ ] `cdc_covidnet`
- [ ] `changehc`
- [ ] `claims_hosp`
- [ ] `combo_cases`
- [ ] `google_health`
- [ ] `google_symptoms`
- [ ] `hhs_facilities`
- [ ] `hhs_hosp`
- [ ] `jhu` (#752)
- [ ] `nchs_mortality`
- [ ] `quidel`
- [ ] `quidel_covidtest`
- [ ] `safegraph`
- [ ] `safegraph_patterns`
- [ ] `usafacts`

This should basically involve:
1. Set up bare-bones `params.json` that specifies:
    - `global.data_source` (the indicator name)
    - `global.span_length = 14`
    - `global.end_date = "latest"`
    - `dynamic.smoothed_signals`
    - `dynamic.expected_lag`
2. Get the indicator output by running it or downloading some cached files from somewhere
3. Run the validation from the indicator directory using `python -m delphi_validation`
4. Investigate the output and use it to tune the remaining validation parameters

qx-teo:
Validation parameters have been added (and implemented) for all but one indicator ran using automation. 

Exceptions include:
* `nchs_mortality`: Indicator provides weekly data, which requires modification of the validator (see #1117)
* cron indicators: doctor-visits, fb-survey, hospital-admissions, indicator-combination. Of all these, params.json.template files have been created, but validation isn't currently implemented in the pipeline yet.

nmdefries:
Validation is rolled out to all daily signals in dry-run mode. Acquisition blocking is turned off because validation is too sensitive for it to be useful.

PR #752, opened 2021-01-29
Add JHU validation parameters

sgsmob:
### Description
Add a `validation` section to the `jhu` `params.json.template`.

### Fixes 
- Partially addresses #725


krivard:
I think you ran into #748

rumackaaron:
Is there more documentation for this? I'm not able to run the `python -m delphi_validation` step mentioned in the issue report and I don't see a `delphi_validation` module anywhere in the code.

sgsmob:
See the `validation/` subdirectory

rumackaaron:
The code runs, but I'm not sure that the output makes sense (this may be a problem with the way I ran the indicator or a bug in `delphi_validator`). I ran the indicator on February 4 and I get 288 `NoDataWarning`s for the most recent 3 days x 16 signals x 6 geo types. But there are files and values up to February 3 in the `receiving` folder. I would have expected 96 errors (for just February 4) instead of 288.

I'm not sure how the JHU indicator is expected to run, do we expect data from the very day on which the indicator is run? If not, the configurations might need to be changed to prevent warnings.

sgsmob:
Can you paste some sample output here please?

rumackaaron:
For each of the 288 combinations of signal, date, and geo-type, I get this warning:
`/Users/arumack/Delphi/ncov2019/covidcast-indicators/jhu/env/lib/python3.6/site-packages/covidcast/covidcast.py:419: NoDataWarning: No jhu-csse deaths_incidence_prop data found on 20210204 for geography 'county'
  NoDataWarning)`
Followed by:
`22693 checks run`
`97 checks failed`
`239 checks suppressed`
`0 warnings`
`{"event": "check_missing_date_files failed for None at resolution None on *: Missing dates are observed; if these dates are already in the API they would not be updated", "logger": "delphi_validator.report", "level": "critical", "timestamp": "2021-02-04T23:18:07.196680Z"}`
Followed by 96 of the following, one for each signal x geo-type, but only for 2021-02-04:
`{"event": "check_missing_geo_sig_date_combo failed for confirmed_7dav_cumulative_num at resolution county on 2021-02-04: test data for a given checking date-geo type-signal type combination is missing. Source data may be missing for one or more dates", "logger": "delphi_validator.report", "level": "critical", "timestamp": "2021-02-04T23:18:07.196900Z"}`

sgsmob:
I am getting a similar report having run today, which in my case was due to JHU not generating any data on 2021-02-05.  I would guess that your experience was the same, which I also presume is because the upstream JHU data had not yet been generated.  Not really sure what we want to do with this since when validation is formally run, it will be after we actually run the indicator in production.

rumackaaron:
But in any event, it's a bug that there are 288 `NoDataWarning`s instead of 96, right?

And maybe the validation params should have a parameter for number of days of delay. For example, for JHU we shouldn't expect to find data for the actual day on which the indicator is run. For Change Healthcare, the delay is five days. I see that sir-cal has this functionality in the `max_age` field of the [params](https://github.com/cmu-delphi/covidcast-indicators/blob/main/sir_complainsalot/params.json.template) (JHU's `max_age` is 2 and Change's `max_age` is 6).

sgsmob:
The `NoDataWarnings` are not part of the validation package; they come from trying to download COVIDcast API data on days for which it doesn't exist but these don't translate into validation failures.

rumackaaron:
> And maybe the validation params should have a parameter for number of days of delay. For example, for JHU we shouldn't expect to find data for the actual day on which the indicator is run. For Change Healthcare, the delay is five days. I see that sir-cal has this functionality in the max_age field of the params (JHU's max_age is 2 and Change's max_age is 6).

Am I correct that there shouldn't be JHU validation errors for the most recent day? If so, is the above a good suggestion?

sgsmob:
If the upstream JHU data for the day is available at the time of the run, the validation errors on the most recent day represent something actually being wrong.  If such data is not available yet for the day, as I suspect was true during our individual runs, the errors are noise.  Currently, we have no way to disambiguate these cases in a one-off run.  However, when run in production, we should assume that the JHU data *is* present and thus the errors corresponding to the most recent day *should* be raised.  As such, I think I agree that the delay parameters are unwanted here.

rumackaaron:
Looking at the timestamps [here](https://github.com/CSSEGISandData/COVID-19/tree/master/csse_covid_19_data/csse_covid_19_daily_reports) it appears that the JHU data for day `t` is uploaded shortly after midnight on day `t+1`. If that's the case, even in production, when the JHU indicator is run on day `t`, it will only produce indicator values for days up to `t-1`.

sgsmob:
I think the validation package is now sufficiently updated that this will work in production.

rumackaaron:
Looks a lot better. When I run the validator on 3/26, I'm getting 96 `NoDataWarning`s because the Covidcast API doesn't have data for 3/25. This seems to be a problem with the API rather than with the validator.

I also get the following:
`check_test_vs_reference_avg_changed failed for deaths_7dav_incidence_prop at resolution hrr on 2021-03-20: Average differences in variables by geo_id between recent & reference data seem large --- either large increase tending toward one direction or large mean absolute difference, relative to average values of corresponding variables.`
Some of the values in `receiving/` are indeed different than in the API, I'm assuming that a file was updated on Github between the last update to the API and when I ran JHU locally. Some of the county and msa values are also different, why is the validator only warning about HRRs?

I also get a few of these:
`RuntimeWarning: Problem obtaining jhu-csse deaths_cumulative_prop data on 20210318 for geography 'nation': error: HTTPSConnectionPool(host='api.covidcast.cmu.edu', port=443): Max retries exceeded with url: /epidata/api.php?endpoint=covidcast&data_source=jhu-csse&signals=deaths_cumulative_prop&time_type=day&geo_type=nation&time_values=20210318&geo_value=%2A (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x1172a4580>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))`
Is this a problem with my internet connection or is the validator creating the wrong URL?

krivard:
> `Failed to establish [...]`

This is a transient network error, which may be related to the hijinks we're currently playing with the API replicas to ease the 2020 as_of restore.

sgsmob:
I also got the message ```check_test_vs_reference_avg_changed failed for deaths_7dav_incidence_prop at resolution hrr on 2021-03-20: Average differences in variables by geo_id between recent & reference data seem large --- either large increase tending toward one direction or large mean absolute difference, relative to average values of corresponding variables.```
but I think that since it is a single day, we should probably treat it as a legitimate failure worthy of a human second look through.

rumackaaron:
The `check_test_vs_reference_avg_changed` is a valid error and I took a look. There are indeed discrepancies between the local indicator and the API values. These discrepancies are also present at the county and msa level, but the error is only triggered at the hrr level. Does it make sense that the validator would throw an error for only hrr in this case?

sgsmob:
county and msa checks are disabled since they were failing on every day.  But that may not be a great way to distinguish legitimate versus illegitimate

sgsmob:
@rumackaaron When you say "valid error" do you just mean "validation correctly computed that the signal value has changed by a lot over the last n days" or "the value of signal changed by a lot and this is probably due to an error in the JHU indicator or the upstream JHU data, so we would not want to accept this output"?

rumackaaron:
Ok, if the desired validator behavior is to disable county and msa checks, then I'm ready to approve the review. @krivard can you confirm?

krivard:
> county and msa checks are disabled since they were failing on every day

that doesn't seem right; we may need to increase the amount of variability we accept from smaller regions rather than suppressing those checks entirely. 

We need the validator to be able to catch abnormally large updates, since USAFacts published a couple of doozies via typos last year (which we'd want to block in future), and individual counties periodically publish a pile of retrospective cases or deaths (which we'd want to flag for the [data anomalies table](https://docs.google.com/spreadsheets/d/1w3jzNizh7rBkl1P-ikXa_GW2m_tRmV97eN7naOIR6jA/edit?usp=sharing) before letting through)

rumackaaron:
> @rumackaaron When you say "valid error" do you just mean "validation correctly computed that the signal value has changed by a lot over the last n days" or "the value of signal changed by a lot and this is probably due to an error in the JHU indicator or the upstream JHU data, so we would not want to accept this output"?

I misunderstood what the error was supposed to check for. I thought that this error was supposed to indicate when the computed values in `receiving/` didn't match those pulled from the API (this is the case for the day and signal in question). So actually ignore "valid error"

sgsmob:
Seems like we need to tune these thresholds better.  See #970

PR #753, opened 2021-01-29
Update makefiles to fix linting failures not failing

chinandrew:
### Description
See https://github.com/cmu-delphi/covidcast-indicators/pull/699#issuecomment-769932078. Having both lint commands run in the same line means that if the first fails but the second succeeds, the command exits with status code 0 and the CI passes. 

One thing we might experience is some existing PRs getting merged in that altered this code but dont pass linting, and we'll just have to fix them if they come up

### Changelog
- Update all makefiles to split command into 2 lines.
- Fix linting errors that snuck by.

### DIFF
--- a/_delphi_utils_python/Makefile
+++ b/_delphi_utils_python/Makefile
@@ -11,9 +11,8 @@ install: venv
 	pip install -e .
 
 lint:
-	. env/bin/activate; \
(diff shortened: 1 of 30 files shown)


krivard (approved):
xlnt

PR #744, opened 2021-01-28
Dataclasses and model code for nowcasting

chinandrew:
Merge after #741 

### Description
This adds the data containers and modelling code for nowcasting sensorization. It's been split into 3 commits for each part described below. For context, these are used to fit/predict the sensor values for a given day, which are fused together later to generate the estimate and then stored in a db for future use (sensor fusion requires training on past sensors). That code is in #745 for refernece

### Changelog
Itemize code/test/documentation changes and files added/removed.
- Data containers and tests. There were some objects that nowcasting needs that we thought would be better encapsulated in a custom class due to our requirements, and also some collections of items that always went together (the sensor config values), so these classes were created instead over passing a bunch of tuples around. I'm sure there's a better way to implement how LocationSeries work, but they seem to suffice for now. Open to suggestions. Basically the challenge was passing a time series around that was tied to a geo type/location that would also allow accessing a properly ordered and continuous array for a given date range for use in linalg operations. (c26bfe8)
- Add AR and regression modeling code. This is fairly standard least squares stuff. AR code is febf27f and regression code is 906cd18.

### Fixes 
- Fixes #(issue)

PR #754, opened 2021-01-29
Nowcast sensor computation/retrieval/export code

chinandrew:
### Description
Merge after #744 
This PR completes the sensorization code for nowcasting and is broke into 4 main parts, detailed below. Context for this code is that the nowcasting will be configured to use a number of sensors, and this should complete the flow for retrieving past/computing new sensors.

### Changelog
- a3b05f3: code to retrieve indicator data for training sensors. Note that get_indicator_data and the two helper functions in get_epidata.py are functional but intended to be replaced once https://github.com/cmu-delphi/delphi-epidata/pull/396 goes live.
- 5156366: computation of sensor values. basically calling the models in #744 for each sensor specified
- f10f3be: retrieval of past sensor values, which we are going to be storing in the epidata database.
- d3c6e82: export data functions so they can be acquired by the database and retrieved the next day by the above commit. It's a little different than now utils.export.create_export_csv works, but I'm open to getting them compatible. The indicator values themselves (the output of the completed which will be surfaced in covidcast as a new signal) should use create_export_csv.


nowcast/delphi_nowcast/sensorization/get_epidata.py
```diff
@@ -0,0 +1,131 @@
...
+                    "endpoint": "covidcast",
+                    "data_source": sensor.source,
+                    "signals": sensor.signal,
+                    "time_type": "day",
+                    "geo_type": location.geo_type,
+                    "time_values": f"20200101-{as_of}",
+                    "geo_value": location.geo_value,
+                    "as_of": as_of
+                }
+            task = asyncio.ensure_future(get(params, session, sensor, location))
```
sgsmob, line 37: `create_task()` is preferred to `ensure_future()`
https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future

nowcast/delphi_nowcast/sensorization/get_epidata.py
```diff
@@ -0,0 +1,131 @@
...
+    as_of
+        Date that the data should be retrieved as of.
+    Returns
+    -------
+    """
+    # gets all available data up to as_of day for now, could be optimized to only get a window
+    output = {}
+    all_combos = product(sensors, locations)
+    loop = asyncio.get_event_loop()
+    future = asyncio.ensure_future(fetch_epidata(all_combos, as_of.strftime("%Y%m%d")))
```
sgsmob, line 64: `create_task()` may apply here too.

nowcast/delphi_nowcast/sensorization/get_epidata.py
```diff
@@ -0,0 +1,131 @@
...
+    responses = loop.run_until_complete(future)
+    for response, sensor, location in responses:
+        if response["result"] not in (-2, 1):
```
sgsmob, line 67: Where do the `(-2, 1)` values come from?

sgsmob: Looking below it seems that this is part of the Epidata definition.  I would add a comment here explaining what those two codes mean and what sort of other return values might trigger this if statement.

nowcast/delphi_nowcast/sensorization/get_epidata.py
```diff
@@ -0,0 +1,131 @@
...
+            raise Exception(f"Bad result from Epidata: {response['message']}")
+        data = LocationSeries(
+            geo_value=location.geo_value,
+            geo_type=location.geo_type,
+            dates=[datetime.strptime(i, "%Y%m%d").date() for i in response.get("epidata", [])
```
sgsmob, line 72: Do we anticipate the `LocationSeries` always being constructed with this idiom?  If so, it seems that it might be cleaner to combine these into a dictionary comprehension here rather than in the `LocationSeries` constructor.  

```
dates=[datetime.strptime(i, "%Y%m%d").date() for i in response["epidata"] if not isnan(i["value"])],
values=[i["value"] for i in response["epidata"] if not isnan(i["value"])],
```
vs.
```
date_to_vals={datetime.strptime(i, "%Y%m%d").date(): i["value"] for i in response["epidata"]
                   if not isnan(i["value"])}
```

sgsmob: Also what is the type of `i` that it can both be coerced into a date from a string but also have a `["value"]` attribute?

nowcast/delphi_nowcast/sensorization/sensor.py
```diff
@@ -0,0 +1,147 @@
...
+            output[ground_truth_sensor].append(
+                LocationSeries(loc.geo_value, loc.geo_type, [ground_truth_pred_date], [ar_sensor])
+            )
+        for sensor in regression_sensors:
+            sensor_pred_date = as_of_date - timedelta(sensor.lag)
+            covariates = indicator_data.get(
+                (sensor.source, sensor.signal, loc.geo_type, loc.geo_value)
+            )
+            if not covariates:
+                print(f"No data: {(sensor.source, sensor.signal, loc.geo_type, loc.geo_value)}")
```
sgsmob, line 56: Should this be a log statement?  If so, it would probably suffice in this PR to leave a TODO comment.

nowcast/tests/sensorization/test_sensor.py
```diff
@@ -0,0 +1,91 @@
...
+
+from delphi_nowcast.data_containers import LocationSeries, SensorConfig
+from delphi_nowcast.sensorization.sensor import compute_sensors, historical_sensors, _export_to_csv
+
+
+class TestComputeSensors:
+
+    @patch("delphi_nowcast.sensorization.sensor.compute_ar_sensor")
+    @patch("delphi_nowcast.sensorization.sensor.get_indicator_data")
+    def test_compute_sensors_no_covariates(self, get_indicator_data, compute_ar_sensor):
```
sgsmob, line 16: These arguments should be, e.g., `mock_get_indicator_data`.

nowcast/tests/sensorization/test_sensor.py
```diff
@@ -0,0 +1,91 @@
...
+        assert compute_sensors(
+            date(2020, 5, 5), test_sensors, test_ground_truth_sensor, test_ground_truth, False
+        ) == {
+            SensorConfig("i", "j", "k", 3): [LocationSeries("ca", "state", [date(2020, 5, 2)], [1.5])],
+        }
+
+    @patch("delphi_nowcast.sensorization.sensor.compute_regression_sensor")
+    @patch("delphi_nowcast.sensorization.sensor.compute_ar_sensor")
+    @patch("delphi_nowcast.sensorization.sensor.get_indicator_data")
+    def test_compute_sensors_covariates(self, get_indicator_data, compute_ar_sensor, compute_regression_sensor):
```
sgsmob, line 32: Same

nowcast/tests/sensorization/test_sensor.py
```diff
@@ -0,0 +1,91 @@
...
+        ) == {
+            SensorConfig("i", "j", "k", 3): [LocationSeries("ca", "state", [date(2020, 5, 1)], [1.5])],
+            SensorConfig("a", "b", "c", 1): [LocationSeries("ca", "state", [date(2020, 5, 4)], [2.5])],
+        }
+
+
+class TestHisoricalSensors:
+
+    @patch("delphi_nowcast.sensorization.sensor.get_historical_sensor_data")
+    def test_historical_sensors_some_data(self, historical):
```
sgsmob, line 52: Same

nowcast/tests/sensorization/test_sensor.py
```diff
@@ -0,0 +1,91 @@
...
+        test_sensors = [SensorConfig("i", "j", "k", 3), SensorConfig("a", "b", "c", 1), SensorConfig("x", "y", "z", 2)]
+        test_ground_truth = [LocationSeries("ca", "state")]
+        assert historical_sensors(
+            None, None, test_sensors, test_ground_truth) == {
+            SensorConfig("i", "j", "k", 3): [LocationSeries(dates=[date(2020, 1, 1)], values=[2])],
+            SensorConfig("a", "b", "c", 1): [LocationSeries(dates=[date(2020, 1, 3)], values=[4])]
+        }
+
+    @patch("delphi_nowcast.sensorization.sensor.get_historical_sensor_data")
+    def test_historical_sensors_no_data(self, historical):
```
sgsmob, line 66: Same

nowcast/tests/sensorization/test_sensor.py
```diff
@@ -0,0 +1,91 @@
...
+                                   lag=4)
+        test_value = LocationSeries("ca", "state", [date(2020, 1, 1)], [1.5])
+        with tempfile.TemporaryDirectory() as tmpdir:
+            out_files = _export_to_csv(test_value, test_sensor, date(2020, 1, 5), receiving_dir=tmpdir)
+            assert len(out_files) == 1
+            out_file = out_files[0]
+            assert os.path.isfile(out_file)
+            assert out_file.endswith("issue_20200105/src/20200101_state_sig.csv")
+            with open(out_file) as f:
+                assert f.read() == "sensor_name,geo_value,value\ntest,ca,1.5\n"
```
sgsmob, line 91: This is fine for the file size now, but it might be more future-proof and set a better precedent for other tests to read the file using `csvreader` or `pandas` and comparing against a Python object of appropriate type rather than the string version.

sgsmob (changes requested):

nowcast/delphi_nowcast/sensorization/get_epidata.py
chinandrew: > Also what is the type of `i` that it can both be coerced into a date from a string but also have a `["value"]` attribute?

ah that's a bug that I didn't catch since the endpoint wasn't live yet. should be `i["timevalue"]`. I like that dict comprehension suggestion, will add

nowcast/tests/sensorization/test_sensor.py
chinandrew: :+1: fixed

nowcast/tests/sensorization/test_sensor.py
chinandrew: fixed

nowcast/tests/sensorization/test_sensor.py
chinandrew: fixed

nowcast/tests/sensorization/test_sensor.py
chinandrew: fixed

nowcast/tests/sensorization/test_sensor.py
chinandrew: updated to use pandas

nowcast/delphi_nowcast/sensorization/get_epidata.py
chinandrew: added a comment

nowcast/delphi_nowcast/sensorization/get_epidata.py
```diff
@@ -0,0 +1,131 @@
...
+    tasks = []
+    async with ClientSession() as session:
+        for sensor, location in combos:
+            params = {
+                    "endpoint": "covidcast",
+                    "data_source": sensor.source,
+                    "signals": sensor.signal,
+                    "time_type": "day",
+                    "geo_type": location.geo_type,
+                    "time_values": f"20200101-{as_of}",
```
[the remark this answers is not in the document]
chinandrew: updated

nowcast/delphi_nowcast/sensorization/get_epidata.py
chinandrew: oh cool, thanks for the info. Updated

nowcast/delphi_nowcast/sensorization/get_epidata.py
chinandrew: I tried changing both and that didnt work, so I think it only applies to the first loop.

nowcast/delphi_nowcast/sensorization/get_epidata.py
chinandrew: updated so LocationSeries takes in a dictionary and these calls becomes dict comprehensions

nowcast/delphi_nowcast/sensorization/sensor.py
chinandrew: added

sgsmob (approved):

PR #757, opened 2021-02-01
Prepare covid_act_now for staging

korlaxxalrok:
### Description
Feature:

Add params j2 template with S3 variables.

### Changelog
- Add ansible/templates/covid_act_now-params-prod.json.j2

### Fixes 
- Fixes #(issue)


korlaxxalrok:
@eujing I think this covers the barest of necessities.

ansible/templates/covid_act_now-params-prod.json.j2
```diff
@@ -0,0 +1,12 @@
+{
+  "static_file_dir": "./static",
+  "export_dir": "/common/covidcast/receiving/covid_act_now",
+  "cache_dir": "./cache",
+  "wip_signal": true,
```
eujing, line 5: Oops this is my fault for providing an outdated params.json, but the `wip_signal` key is no longer needed.
Sorry about that!

eujing (changes requested):
Looks good to me, just the extra `wip_signal` I accidentally sent!

ansible/templates/covid_act_now-params-prod.json.j2
korlaxxalrok: Makes sense. I'll remove it.

eujing (approved):

Issue #537, opened 2020-11-17
Move population-based prop computation into the utility so it can be adequately tested


krivard:
Both USAFacts and JHU aggregate prop signals by population from county to state, a process which is susceptible to error. Specifically, the population of the megafips is equal to the state population, and megafips + all individual counties yields state population + population of each individual county = 2x the state population.

Testing this effect in individual indicators will always result in us playing catch-up when new counts-based indicators are added which include population-based prop signals aggregated from the county level upwards.

We should instead pursue a refactor which moves population-prop aggregation into the utility, so that it can be tested once there, and used everywhere in confidence.

chinandrew:
> Specifically, the population of the megafips is equal to the state population

Why is this the case? Does zeroing it out cause a div/0 error elsewhere?

krivard:
> Does zeroing it out cause a div/0 error elsewhere?

Yes -- prop signals divide by the population; the best thing a megafips prop signal can divide by is the state population

chinandrew:
Per a conversation with @krivard  and @dshemetov, we're going with an approach where replace_geocode will do the "zero out the megafips populations so we don't double count" logic  that are [currently implemented in JHU](https://github.com/cmu-delphi/covidcast-indicators/blob/961356f8f7c6d2f3d919c9967f448e7f8e08b4c7/jhu/delphi_jhu/geo.py#L40-L44) and USAfacts. This requires the population column to be added (it will throw a warning otherwise) before the geomapping. After mapping, the indicators will be able to do a direct prop=value/population computation.

We also discussed supporting more geographies in add_population_column and having that operation done after geomapping, but didn't end up doing that because we want the denominator to be the population of all the reported geographies and not the population of the full region (e.g. if 5 counties are reported in a state, use the population from all 5 counties, not the state population).

PR #759, opened 2021-02-02
Have replace_geocode ignore megafips populations

chinandrew:
### Description
See https://github.com/cmu-delphi/covidcast-indicators/issues/537#issuecomment-771299794

### Changelog
- Add a population column arg to replace_geocode that will be used to null out megafips, with a warning thrown if it is not specified
- Add logic to only null the megafips population if there are other fips reported in that state
- unit tests and update docstrings. Also updated an unrelated docstring entry that was wrong.

### Fixes 
- Partially addresses #537

### DIFF
--- a/_delphi_utils_python/delphi_utils/geomap.py
+++ b/_delphi_utils_python/delphi_utils/geomap.py
@@ -424,9 +424,9 @@ class GeoMapper:  # pylint: disable=too-many-public-methods
             df.loc[megafips_to_zero, "population"] = 0
             print(megafips_to_zero)
         if from_code == "fips" and not pop_col:
-            warnings.warn("Without specifying a population column, megaFIPS populations be double "
-                          "counted later on. If working with populations, add the FIPS column "
-                          "before replacing the geocode.")
+            warnings.warn("Without specifying a population column, megaFIPS populations may be "
+                          "double counted later on. If working with populations, add the FIPS "
+                          "column before replacing the geocode.")
         df = self.add_geocode(
             df, from_code, new_code, from_col=from_col, new_col=new_col, dropna=dropna
         ).drop(columns=from_col)
--- a/_delphi_utils_python/delphi_utils/geomap.py
+++ b/_delphi_utils_python/delphi_utils/geomap.py
@@ -401,10 +401,13 @@ class GeoMapper:  # pylint: disable=too-many-public-methods
         date_col: str or None, default "date"
             Specify which column contains the date values. Used for value aggregation.
             If None, then the aggregation is done only on geo_id.
+        pop_col: str or None
+            Specify which columns contains population values. Used for correcting for double
+            counting if a megafips is present.
         data_cols: list, default None
             A list of data column names to aggregate when doing a weighted coding. If set to
             None, then all the columns are used except for date_col and new_col.
-        dropna: bool, default False
+        dropna: bool, default True
             Determines how the merge with the crosswalk file is done. If True, the join is inner,
             and if False, the join is left. The inner join will drop records from the input database
             that have no translation in the crosswalk, while the outer join will keep those records
--- a/_delphi_utils_python/delphi_utils/geomap.py
+++ b/_delphi_utils_python/delphi_utils/geomap.py
@@ -9,6 +9,7 @@ TODO:
 """
 # pylint: disable=too-many-lines
 from os.path import join
+import warnings
 
 import pandas as pd
 import pkg_resources
@@ -371,6 +372,7 @@ class GeoMapper:  # pylint: disable=too-many-public-methods
         from_col=None,
         new_col=None,
         date_col="date",
+        pop_col=None,
         data_cols=None,
         dropna=True,
     ):
@@ -416,6 +418,15 @@ class GeoMapper:  # pylint: disable=too-many-public-methods
         from_col = from_code if from_col is None else from_col
         new_col = new_code if new_col is None else new_col
 
+        if from_code == "fips" and pop_col:
+            # Zero out the duplicate population in megafips to avoid double counting.
+            megafips_to_zero = [i in self._megafips_to_zero(df[from_col]) for i in df[from_col]]
+            df.loc[megafips_to_zero, "population"] = 0
+            print(megafips_to_zero)
+        if from_code == "fips" and not pop_col:
+            warnings.warn("Without specifying a population column, megaFIPS populations be double "
+                          "counted later on. If working with populations, add the FIPS column "
+                          "before replacing the geocode.")
         df = self.add_geocode(
             df, from_code, new_code, from_col=from_col, new_col=new_col, dropna=dropna
         ).drop(columns=from_col)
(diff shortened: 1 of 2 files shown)


_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -399,10 +401,13 @@ def replace_geocode(
         date_col: str or None, default "date"
             Specify which column contains the date values. Used for value aggregation.
             If None, then the aggregation is done only on geo_id.
+        pop_col: str or None
+            Specify which columns contains population values. Used for correcting for double
+            counting if a megafips is present.
```
krivard, line 406:
```suggestion
            Specify which columns contains population values. Used to avoid double
            counting if a megafips is present.
```

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -580,3 +595,16 @@ def get_geo_values(self, geo_type):
             crosswalk = pd.read_csv(stream, dtype=str)
             self.geo_lists[geo_type] = set(crosswalk[geo_type])
             return self.geo_lists[geo_type]
+
+    def _megafips_to_zero(self, fips_list):
+        output = set()
+        fips_set = set(fips_list)
+        state_fips_codes = {str(x).zfill(2) + "000" for x in range(1, 73)}
```
krivard, line 602: should we be using the actual list of state codes here instead of `range(1,73)`?

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -416,6 +421,15 @@ def replace_geocode(
         from_col = from_code if from_col is None else from_col
         new_col = new_code if new_col is None else new_col
 
+        if from_code == "fips" and pop_col:
+            # Zero out the duplicate population in megafips to avoid double counting.
+            megafips_to_zero = [i in self._megafips_to_zero(df[from_col]) for i in df[from_col]]
```
krivard, line 426: does Python do the smart thing here and compute `self._megafips_to_zero(df[from_col])` once, or will it run it N times?

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -580,3 +595,16 @@ def get_geo_values(self, geo_type):
...
+        output = set()
+        fips_set = set(fips_list)
+        state_fips_codes = {str(x).zfill(2) + "000" for x in range(1, 73)}
+        for state in state_fips_codes:
+            is_present = state in fips_set
+            # check if megafips is the only code reported for that state
+            is_only_fips = not any(i.startswith(state[:2]) and
+                                   not i.endswith(state[2:]) for i in fips_set)
+            if is_present and not is_only_fips:
+                output.add(state)
```
krivard, line 609: uh this appears to be O(73n)

would it make more sense to do something like
```
index = set()
for fips in fips_list:
    if fips[2:] != '000':
        index.add(fips[:2])
return {statecode + '000' for statecode in index}
```

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -580,3 +595,16 @@ def get_geo_values(self, geo_type):
             crosswalk = pd.read_csv(stream, dtype=str)
             self.geo_lists[geo_type] = set(crosswalk[geo_type])
             return self.geo_lists[geo_type]
+
+    def _megafips_to_zero(self, fips_list):
```
krivard, line 599: add docstring

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -416,6 +421,15 @@ def replace_geocode(
...
         new_col = new_code if new_col is None else new_col
 
+        if from_code == "fips" and pop_col:
+            # Zero out the duplicate population in megafips to avoid double counting.
+            megafips_to_zero = [i in self._megafips_to_zero(df[from_col]) for i in df[from_col]]
+            df.loc[megafips_to_zero, "population"] = 0
+            print(megafips_to_zero)
+        if from_code == "fips" and not pop_col:
+            warnings.warn("Without specifying a population column, megaFIPS populations may be "
+                          "double counted later on. If working with populations, add the FIPS "
```
krivard, line 431:
```suggestion
                          "double counted later on. If working with populations, add the population "
```

_delphi_utils_python/tests/test_geomap.py
```diff
@@ -310,3 +311,24 @@ def test_get_geos(self):
         assert len(gmpr.get_geo_values("fips")) == 3287
         assert len(gmpr.get_geo_values("state_id")) == 60
         assert len(gmpr.get_geo_values("zip")) == 32976
+
+    def test_megafips_zeroed(self):
+        gmpr = GeoMapper()
+        pd.testing.assert_frame_equal(
+            gmpr.replace_geocode(self.fips_data_2, "fips", "state_id", pop_col="population"),
```
krivard, line 318: could you use `state_code` here instead, it makes it easier to see what happened to `04000`

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -416,6 +421,15 @@ def replace_geocode(
         from_col = from_code if from_col is None else from_col
         new_col = new_code if new_col is None else new_col
 
+        if from_code == "fips" and pop_col:
+            # Zero out the duplicate population in megafips to avoid double counting.
+            megafips_to_zero = [i in self._megafips_to_zero(df[from_col]) for i in df[from_col]]
+            df.loc[megafips_to_zero, "population"] = 0
+            print(megafips_to_zero)
```
krivard, line 428:
```suggestion
```

krivard (changes requested):
Fix warning text snafu, add docs, address efficiency

_delphi_utils_python/delphi_utils/geomap.py
chinandrew: yea good idea, this was just the jhu copy paste. thanks for calling out, will fix

_delphi_utils_python/tests/test_geomap.py
chinandrew: good call, will do

_delphi_utils_python/delphi_utils/geomap.py
chinandrew: :+1:

_delphi_utils_python/delphi_utils/geomap.py
chinandrew: probably not, ill extract it to a variable

_delphi_utils_python/delphi_utils/geomap.py
chinandrew: ah right, since we just create the mask by checking inclusion into this set we don't actually need to filter the values to whats present in the df. good idea.

krivard (approved):
xlnt

chinandrew:
Superseded by #771

PR #760, opened 2021-02-02
Replace JHU megafips zeroing logic with replace_geocode

chinandrew:
### Description
Merge after #759.

Use the megafips zeroing logic in replace_geocode instead of in the JHU code. should also fix an hhs/nation bug where the megafips were double counted.

### Changelog
- Delete old megafips zeroing logic and use replace_geocode instead
- Fixed some single/double quote disparities.
- Add tests for population, verified by manually doing the population computations on the test csv file.

### Fixes 
na


chinandrew:
superseded by #771

PR #761, opened 2021-02-02
Replace USAFacts megafips zeroing logic with replace_geocode

chinandrew:
### Description
Merge after #759 
Use the megafips zeroing logic in replace_geocode instead of in the USAFacts code. should also fix an hhs/nation bug where the megafips were double counted.

Changelog

- Delete old megafips zeroing logic and use replace_geocode instead
- Move hhs and nation into the state conditional section so they also use replace_geocode
- Add tests for population

### Fixes 
- Fixes #(issue)


chinandrew:
superseded by #771

PR #763, opened 2021-02-03
Move nowcast parameters to constants file and add placeholders

chinandrew:
### Description
Moving relevant constants for nowcasting to their own file and adding a test that duplicates each value and will fail if any are changed. This is to ensure changes to parameters are done explicitely. Values are placeholders for now.

cc @mariajahja 

### Changelog
- Add parameters to a constants files and use them instead of function default parameters
- update tests to reflect lack of default constants
- Change export_data bool to export_dir string to prepare for putting export _dir into the params file and passing it in after read_params() is called.

### DIFF
--- a/nowcast/tests/test_constants.py
+++ b/nowcast/tests/test_constants.py
@@ -1,0 +1,15 @@
+"""This test is to ensure the constants are not accidentally altered."""
+
+from delphi_nowcast import constants
+from delphi_nowcast.data_containers import SensorConfig
+
+
+def test_constants():
+    assert len(dir(constants)) == 16
+    assert constants.GROUND_TRUTH_INDICATOR == SensorConfig("placeholder", "placeholder", "placeholder", 0)
+    assert constants.DELAY_DISTRIBUTION == []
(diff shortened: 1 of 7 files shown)


nowcast/tests/test_constants.py
```diff
@@ -0,0 +1,21 @@
+"""This test is to ensure the constants are not accidentally altered."""
```
krivard, line 1: This is a little bit [Frog And Toad Try Not To Eat Cookies](https://sites.fas.harvard.edu/~fs22k/Assignments/frog-toad.html) but I'm not sure there's a better way to do it so 🤷

krivard (approved):
👍

krivard:
(modulo the linting fix)

nowcast/tests/test_constants.py
chinandrew: i guess what we're protecting against is conceptually no different than someone changing the implementation of smoothing or something and us not catching it, which an output validation integration test would capture, so we could move to that once it's ready.

chinandrew:
> (modulo the linting fix)

fixed

PR #765, opened 2021-02-04
Support Wave 8

capnrefsmmat:
Supports responses in Wave 8.

Note D1b is "new" in Wave 8, but existed much earlier, so I don't need to add it back -- it's already in the list of columns.


nmdefries (commented):
Looks like item V6 ("Why don't you believe that you need a COVID vaccine?") is missing.

capnrefsmmat:
Good catch, thanks. Pushed a fix.

nmdefries (approved):

PR #766, opened 2021-02-04
Have hhs and nation prop signals be computed from the num signal in combo cases

chinandrew:
### Description

Depends on #759. Relevant commits are af2bdfa and  f612c23

Fixes a bug in the combo cases indicator, see[ this slack](https://delphi-org.slack.com/archives/CT7BUGPFC/p1612175736024000).

This change makes it so if the geo is hhs or nation, the `num` indicator is queried regardless of it is a `num` or `prop` signal, and after joining the jhu and usafacts dfs, the population column is added, the new geo resolution is mapped to through `replace_geocode`, and the proportion is calculated after.

### Changelog
- Add function for taking a county/num df and turning it into a hhs or nation df with num or prop signal
- Tests

### Fixes 
- Fixes #(issue)

### DIFF
--- a/_delphi_utils_python/delphi_utils/geomap.py
+++ b/_delphi_utils_python/delphi_utils/geomap.py
@@ -9,6 +9,7 @@ TODO:
 """
 # pylint: disable=too-many-lines
 from os.path import join
+import warnings
 
 import pandas as pd
 import pkg_resources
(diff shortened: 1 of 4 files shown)


chinandrew:
superseded by #771

PR #767, opened 2021-02-04
Add item V2 (# vaccine doses) to pipeline

nmdefries:
### Description
Add item V2 to pipeline as percent of respondents who have received 2 doses; respondents reporting receiving 1 dose are included in the denominator. "I don't know"s are rare (0.3-0.4% of responses) and are ignored for this calculation.


capnrefsmmat (approved):
Looks good. Once this is ready to deploy, you'll want to add the signal to the table here: https://github.com/cmu-delphi/delphi-epidata/blob/main/docs/api/covidcast-signals/fb-survey.md#vaccination-indicators

PR #779, opened 2021-02-07
Use 7 decimals of precision for value and se floats

### LINKED ISSUE #769: Make the CSV data exporter and archive differ use limited precision on floats
We don't need 15 digits of precision anywhere. 7 is plenty.

If we do this in the data export routine in delphi_utils, it'll propagate everywhere (good), but then the archive differ will freak out (bad). 

So, make the archive differ only check the first 7 digits, as part of the same changeset.

chinandrew:
### Description
See #769. Convert export and archive differ to round values and standard errors to 7 decimals. 

### Changelog
- Add rounding just before csv export
- Add rounding to archive differ comparison
- Unit tests

### Fixes 
- Fixes #769

### DIFF
--- a/_delphi_utils_python/delphi_utils/archive.py
+++ b/_delphi_utils_python/delphi_utils/archive.py
@@ -75,10 +75,10 @@ def diff_export_csv(
 
     before_df = pd.read_csv(before_csv, dtype=export_csv_dtypes)
     before_df.set_index("geo_id", inplace=True)
-
+    before_df = before_df.round(7)
     after_df = pd.read_csv(after_csv, dtype=export_csv_dtypes)
     after_df.set_index("geo_id", inplace=True)
-
+    after_df = after_df.round(7)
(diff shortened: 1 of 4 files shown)


_delphi_utils_python/delphi_utils/archive.py
```diff
@@ -75,10 +75,10 @@ def diff_export_csv(
 
     before_df = pd.read_csv(before_csv, dtype=export_csv_dtypes)
     before_df.set_index("geo_id", inplace=True)
-
+    before_df = before_df.round(7)
```
sgsmob, line 78: Please note in the function docstring that this rounding is happening.

sgsmob (approved):

_delphi_utils_python/delphi_utils/archive.py
chinandrew: added to both archive and export docstrings. also updated this to use `.round({"val": 7, "se": 7})` to be more explicit

PR #780, opened 2021-02-08
HHS Hosp: Have last date range get ignored if no epidata

chinandrew:
### Description
For HHS Hosp, to ensure the Epidata calls do not hit the row limit, windows of data are retrieved sequentually. For example, to get data from 2020/11/6 to 2021/2/8, the following ranges are queried:
```
 {'from': 20201106, 'to': 20201206},
 {'from': 20201207, 'to': 20210106},
 {'from': 20210107, 'to': 20210206},
 {'from': 20210207, 'to': 20210208}
```

This last date range may be anywhere from 1-30 days, and if it is a short range it will be recent enough such that the entire range returns no results and the indicator errors. This PR adds a statement to not throw an error if that is the case.

### Changelog
- Add conditional statement to ignore a "no results" output if the date range is the last one.
- Add unit test
- Removed an unused variable I stumbled upon

### Fixes 
- Fixes #(issue)

### DIFF
--- a/hhs_hosp/delphi_hhs/run.py
+++ b/hhs_hosp/delphi_hhs/run.py
@@ -75,8 +75,11 @@ def run_module():
     dfs = []
     for r in date_range:
         response = Epidata.covid_hosp(request_all_states, r)
-        if response['result'] != 1:
+        # The last date range might only have recent days that don't have any data, so don't error.
+        if response["result"] != 1 and r != date_range[-1]:
             raise Exception(f"Bad result from Epidata: {response['message']}")
+        if response["result"] == -2 and r == date_range[-1]:
+            continue
         dfs.append(pd.DataFrame(response['epidata']))
     all_columns = pd.concat(dfs)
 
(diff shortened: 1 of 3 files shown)


hhs_hosp/delphi_hhs/run.py
```diff
@@ -75,8 +75,11 @@ def run_module():
     dfs = []
     for r in date_range:
         response = Epidata.covid_hosp(request_all_states, r)
-        if response['result'] != 1:
+        # The last date range might only have recent days that don't have any data, so don't error.
+        if response["result"] != 1 and r != date_range[-1]:
             raise Exception(f"Bad result from Epidata: {response['message']}")
+        if response["result"] == -2 and r == date_range[-1]:
```
benjaminysmith, line 81: What does -2 correspond to? Maybe add a comment.

benjaminysmith (approved):

hhs_hosp/delphi_hhs/run.py
chinandrew: no results, good idea on the comment
```suggestion
        if response["result"] == -2 and r == date_range[-1]:  # -2 code means no results
```

PR #781, opened 2021-02-08
Add parsing for - delimited dates

chinandrew:
### Description
USAFacts switched up their date separators :/. Adding a statement that works for both `/` and `-`
 
### Changelog
- Have `-` be stripped from column names during check on top of `/`
- Change one test dataset to use `-`

### Fixes 
- Fixes #(issue)


krivard (approved):
lgtm!

PR #783, opened 2021-02-08
Cast time value to str in nowcast

chinandrew:
### Description
Forgot to cast the epidata return values for parsing. 

### Changelog
- Cast time values to str so they can be converted properly

### Fixes 
- Fixes #(issue)


mariajahja (approved):
good catch!

PR #785, opened 2021-02-08
changehc: fetch files based on filename not modification time

### LINKED ISSUE #600: CHNG should compute source filename instead of using delivery time
From https://github.com/cmu-delphi/covidcast-indicators/pull/595#pullrequestreview-544115304, regarding the brittleness of using delivery time to identify which files to download:

> If we run this script at 4:01 and they had posted the files already at 3, then the script will execute properly.
> 
> If they consistently post the files late, then I'm pretty sure the code will throw some sort of "File not found" error. It uses dropdate_dt to determine the filenames, so it will expect there to be files from yesterday, but the code can only download files from two days ago.
> 
> If CHNG posts the files at 5pm one day and at 4pm the next, and we run the code between 4pm and 5pm, the code will throw an assertion error because it found too many files dropped within the last 24 hours.
> 
> This code followed the outline of the HSP download scheme, but maybe a simpler solution is to **compute the expected filename beforehand** and check if it's been uploaded to their server yet. This way the errors would be more intelligible. For now, if we run the script after 8, everything should work properly.

This issue tracks work to convert the data fetcher to use the filename, not the delivery time, to determine which files to download.

We should be careful to explicitly consider the case of a late delivery (ie where the file has not yet been delivered at the time when the indicator is run). Do we want the indicator to skip that issue, and never process the file even if it is posted before the next day's run? Do we want to try to automatically detect that yesterday's missed file is now available, and notify someone to produce and load the missed issue manually?

krivard:
### Description
In order to generate the missing issues from last week caused by the bigint outage, it would be good if we could run the changehc indicator (1) outside of the production environment, and (2) to generate a specific past issue by setting `drop_date`. The current indicator code sets the expected local filenames based on `drop_date`, but downloads from SFTP based on the current local time and the modification time of the file on the server. This means that the indicator downloads the wrong files when `drop_date` is any value other than yesterday. The production environment has the correct file cached, but running a large repair like this in the production environment risks polluting it.

This PR modifies the download procedures to ignore the current time and remote file modification time, and instead fetch based on the YYYYmmdd prefix of the remote file.

### Changelog
- download_ftp_files: thread `filedate` argument through file fetcher functions
- run: pass `filedate` to file fetchers
- tests: check filename instead of modification time

### Fixes 
- Fixes #600

### DIFF
--- a/changehc/delphi_changehc/download_ftp_files.py
+++ b/changehc/delphi_changehc/download_ftp_files.py
@@ -1,4 +1,4 @@
-"""Download files modified in the last 24 hours from the specified ftp server."""
+"""Download files from the specified ftp server."""
 
 # standard
 import datetime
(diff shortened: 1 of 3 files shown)


rumackaaron (approved):
lgtm

PR #786, opened 2021-02-08
Update SirCAL logging

benjaminysmith:
### Description
Update SirCAL logging to support replicating alerts through elastic

### Changelog
- Log aggregated SLA violations
- Change some messages we don't need in the logs back to prints

### Fixes 
- Tweaks #612

### DIFF
--- a/sir_complainsalot/delphi_sir_complainsalot/run.py
+++ b/sir_complainsalot/delphi_sir_complainsalot/run.py
@@ -37,14 +37,6 @@ def run_module():
                                        params["sources"], params.get("grace", 0), LOGGER))
 
     if len(complaints) > 0:
-        for complaint in complaints:
-            LOGGER.critical(event="signal out of SLA",
-                            message=complaint.message,
-                            data_source=complaint.data_source,
-                            signal=complaint.signal,
-                            geo_types=complaint.geo_types,
-                            last_updated=complaint.last_updated.strftime("%Y-%m-%d"))
-
         report_complaints(complaints, slack_notifier)
     
     elapsed_time_in_seconds = round(time.time() - start_time, 2)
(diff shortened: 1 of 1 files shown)


sir_complainsalot/delphi_sir_complainsalot/run.py
```diff
@@ -37,14 +37,6 @@ def run_module():
                                        params["sources"], params.get("grace", 0), LOGGER))
 
     if len(complaints) > 0:
-        for complaint in complaints:
-            LOGGER.critical(event="signal out of SLA",
```
sgsmob, line 41: Why does this need to be moved?

sir_complainsalot/delphi_sir_complainsalot/run.py
```diff
@@ -37,14 +37,6 @@ def run_module():
...
-                            message=complaint.message,
-                            data_source=complaint.data_source,
-                            signal=complaint.signal,
-                            geo_types=complaint.geo_types,
-                            last_updated=complaint.last_updated.strftime("%Y-%m-%d"))
```
sgsmob, line 46: This argument is dropped below.  Was that intentional?

sgsmob (commented):
Just a few questions about what is going on here but otherwise looks good

sir_complainsalot/delphi_sir_complainsalot/run.py
benjaminysmith: Old: each line-level complaint was being logged one at a time (one line per type/source/signal/geo combo)
New: only the aggregated complaint is logged (one line per type/source; signal/geo aggregated)

(this makes it much more useful for alerting)

sir_complainsalot/delphi_sir_complainsalot/run.py
benjaminysmith: Yeah, in the new code it's been pulled directly into the aggregated message.

sgsmob (approved):

PR #794, opened 2021-02-10
Switch binary doses item to reference V2

nmdefries:
Bug fix: definition of `v_received_2_vaccine_doses` was referencing item V1 (vaccinated) instead of V2 (# doses received).


capnrefsmmat (approved):
whoops; good thing we're not shipping this publicly in the API yet... be sure to ping @krivard to deploy this fix before we start ingesting the new signals publicly

Issue #795, opened 2021-02-11
Impose parameter hierarchy in params.json files


sgsmob:
To prepare for the future scenario in which the indicators are run as part of a pipeline that includes validation and archiving steps, we should refactor the parameters to follow along accordingly.

In particular, there should be 4 top level sections:
* `common`:  parameters shared across steps
    * `export_dir`
    * `log_filename`
    * `log_exceptions`
* `indicator`:  indicator-specific parameters read in `{INDICATOR}/run.py`
* `validation`:  validation-specific parameters (see validation README)
* `archive`:  archive-specific parameters read in `_delphi_utils_python/delphi_utils/archive.py`
    * `cache_dir`
    * AWS or Github authentication parameters

Progress
- [x] cdc_covidnet
- [x] change_hc
- [x] claims_hosp
- [x] combo_cases_and_deaths
- [x] covid_act_now
- [x] doctor_visits
- [x] google_health
- [x] google_symptoms
- [x] hhs_facilities
- [x] hhs_hosp
- [x] jhu
- [x] nchs_mortality
- [x] quidel
- [x] quidel_covidtest
- [x] safegraph
- [x] safegraph_patterns
- [x] usafacts

PR #796, opened 2021-02-11
Change safegraph parameter hierarchy

sgsmob:
### Description
Update the safegraph `params.json` file and loading code to follow the conventions of #795

### DIFF
--- a/ansible/templates/safegraph-params-prod.json.j2
+++ b/ansible/templates/safegraph-params-prod.json.j2
@@ -1,14 +1,20 @@
 {
-  "static_file_dir": "./static",
-  "raw_data_dir": "/common/safegraph",
-  "export_dir": "./receiving",
-  "cache_dir": "./cache",
-  "n_core": "12",
-  "aws_access_key_id": "{{ safegraph_aws_access_key_id }}",
-  "aws_secret_access_key": "{{ safegraph_aws_secret_access_key }}",
-  "aws_default_region": "us-east-1",
-  "aws_endpoint": "https://s3.wasabisys.com",
-  "sync": true,
-  "wip_signal" : [],
-  "log_filename": "/var/log/indicators/safegraph.log"
+  "common": {
+    "export_dir": "./receiving",
+    "log_filename": "/var/log/indicators/safegraph.log"
+  },
+  "indicator": {
+    "static_file_dir": "./static",
+    "raw_data_dir": "/common/safegraph",
+    "n_core": "12",
+    "aws_access_key_id": "{{ safegraph_aws_access_key_id }}",
+    "aws_secret_access_key": "{{ safegraph_aws_secret_access_key }}",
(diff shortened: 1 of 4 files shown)


safegraph/params.json.template
```diff
@@ -1,17 +1,23 @@
...
-  "aws_secret_access_key": "",
-  "aws_default_region": "",
-  "aws_endpoint": "",
-  "sync": true,
-  "wip_signal" : ["median_home_dwell_time_7dav",
-                  "completely_home_prop_7dav",
-                  "part_time_work_prop_7dav",
-                  "full_time_work_prop_7dav"],
-  "log_exceptions": false
+  "common": {
```
chinandrew, line 2: in #795 , you use `shared` instead of `common`, is that change intentional? 

Also, is export dir  here isntead of `indicator` because it's also shared with the archive step?

chinandrew:
this also needs to be updated https://github.com/cmu-delphi/covidcast-indicators/blob/main/ansible/templates/safegraph-params-prod.json.j2. I wonder if there's a better way to keep these in sync... I've definitely forgotten it before

safegraph/params.json.template
sgsmob: No they should both be the same.  I just changed #795 to be `common` although I can switch them all to `shared` instead too.

Export dir is a parameter shared in the validation step too.

sgsmob:
> this also needs to be updated https://github.com/cmu-delphi/covidcast-indicators/blob/main/ansible/templates/safegraph-params-prod.json.j2. I wonder if there's a better way to keep these in sync... I've definitely forgotten it before

Thanks, done

safegraph/params.json.template
chinandrew: oh right, because you need to validate the output :facepalm:. `common` sounds good to me.

chinandrew (approved):
:shipit::shipit::shipit:

ansible/templates/safegraph-params-prod.json.j2
```diff
@@ -1,14 +1,20 @@
...
+  "indicator": {
+    "static_file_dir": "./static",
+    "raw_data_dir": "/common/safegraph",
+    "n_core": "12",
+    "aws_access_key_id": "{{ safegraph_aws_access_key_id }}",
+    "aws_secret_access_key": "{{ safegraph_aws_secret_access_key }}",
+    "aws_default_region": "us-east-1",
+    "aws_endpoint": "https://s3.wasabisys.com",
+    "sync": true,
+    "wip_signal" : [],
```
chinandrew, line 15:
```suggestion
    "wip_signal" : []
```

chinandrew: caught by json ci check

sgsmob: thanks

PR #799, opened 2021-02-12
Impose parameter hierarchy on NCHS

sgsmob:
### Description
Update `nchs_mortality` `params.json` files in accordance with #795.

Note that although the archive functionality is inside the indicator code, the corresponding parameters are part of the `archive` object in preparation for a separation between the archive and indicator modules.


chinandrew (approved):
:shipit::shipit::shipit:

PR #804, opened 2021-02-12
Refactor parameters for google health

sgsmob:
### Description
Update `google_health` `params.json` in accordance with #795.

Also rounds values of test comparison files to 7 places to match changes added in #779 and fixes some pylint noncompliance.


google_health/delphi_google_health/run.py
```diff
@@ -37,32 +37,27 @@ def run_module():
...
-    arch_diff.update_cache()
-    print(arch_diff)
-    # if missing start_date, set to today (GMT) minus 5 days
-    if start_date == "":
-        now = datetime.datetime.now(datetime.timezone.utc)
-        start_date = (now - datetime.timedelta(days=4)).strftime("%Y-%m-%d")
+        __name__, filename=params["common"].get("log_filename"),
+        log_exceptions=params["common"].get("log_exceptions", True))
+
+    if params["archive"]:
```
chinandrew, line 55: if there's any chance archive isn't in the json instead of having an empty value, should use `.get()` here

google_health/delphi_google_health/run.py
```diff
@@ -22,7 +22,7 @@
 from .data_tools import format_for_export
 from .pull_api import GoogleHealthTrends, get_counts_states, get_counts_dma
 from .map_values import derived_counts_from_dma
-from .constants import (SIGNALS, RAW, SMOOTHED,
```
chinandrew, line 25: was this an unused constant?

chinandrew (approved):
:shipit::shipit::shipit:

chinandrew (changes requested):
oops accidently approved, one open comment.

google_health/delphi_google_health/run.py
sgsmob: Not sure if this is important to your thinking on this, but the point here is to use the existence of the `archive` field as a marker of whether to do archiving at all rather than as a default value safeguard.

google_health/delphi_google_health/run.py
sgsmob: Yes

google_health/delphi_google_health/run.py
chinandrew: with the current approach, if no archiving is desired, you still need an `archive: {}` (or None or some other falsy value) and can't remove the archive key entirely from the json. If that's the intention that sgtm, but wasn't sure.

sgsmob: why can't you remove it entirely?

chinandrew: won't it keyerror?

sgsmob: yes of course.  Fixed to do what I meant

chinandrew (approved):
:shipit::shipit::shipit:

PR #806, opened 2021-02-12
Refactor parameters for Google symptoms

sgsmob:
### Description
Update `google_symptoms` `params.json` files in accordance with #795.  Also removes some unused parameters.

### DIFF
--- a/google_symptoms/params.json.template
+++ b/google_symptoms/params.json.template
@@ -1,8 +1,12 @@
 {
-  "export_start_date": "2020-02-20",
-  "static_file_dir": "./static",
-  "export_dir": "./receiving",
-  "cache_dir": "./cache",
-  "path_to_bigquery_credentials": "credentials.json",
-  "log_exceptions": false
+  "common": {
+    "export_dir": "./receiving",
+    "log_exceptions": false
+  },
+  "indicator": {
+    "export_start_date": "2020-02-20",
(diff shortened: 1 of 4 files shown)


google_symptoms/params.json.template
```diff
@@ -1,8 +1,12 @@
...
-  "path_to_bigquery_credentials": "credentials.json",
-  "log_exceptions": false
+  "common": {
+    "export_dir": "./receiving",
+    "log_exceptions": false
+  },
+  "indicator": {
+    "export_start_date": "2020-02-20",
+    "num_export_days": 14,
+    "path_to_bigquery_credentials": "credentials.json",
```
chinandrew, line 9: unrelated to this PR as a refactor, but why don't these appear in the ansible json?

sgsmob: @nmdefries I think you may have added this line initially

nmdefries: Yes, these params should also appear in the Ansible json. @sgsmob you can add them if you want or I'll make a followup PR.

sgsmob: I will add it here, thanks for the info

sgsmob: Done

chinandrew (approved):
:shipit::shipit::shipit:

PR #809, opened 2021-02-12
Impose parameter hierarchy on USAFacts indicator

sgsmob:
### Description
Update `usafacts` `params.json` files to use the schema in #795.

Also changes the archive functionality to be gated by the existence of the `archive` submessage.


chinandrew (approved):
:shipit::shipit::shipit:

PR #812, opened 2021-02-15
Switch nowcast to epidata async call and add tests

chinandrew:
### Description
Previously nowcasting had it's own async implementation, but after epidata 0.0.11, we can use the client's version. This PR changes that, and also adds tests for the epidata calls.

cc @mariajahja FYI

### Changelog
- Replace old async code with new Epidata.async_epidata() call
- Fix some argument ordering I had switched
- Move get_epidata up on directory since it's used by deconvolution code too.
- Add tests for the epidata calls

### Fixes 
- Fixes #(issue)

### DIFF
--- a/nowcast/delphi_nowcast/get_epidata.py
+++ b/nowcast/delphi_nowcast/get_epidata.py
@@ -5,40 +5,12 @@ from itertools import product
 
 from numpy import isnan
 from pandas import date_range
-from aiohttp import ClientSession
 
-# from ..delphi_epidata import Epidata  # used for local testing
 from delphi_epidata import Epidata
 
-from ..data_containers import LocationSeries, SensorConfig
+from delphi_nowcast.data_containers import LocationSeries, SensorConfig
 
 EPIDATA_START_DATE = 20200101
-
-async def get(params, session, sensor, location):
-    """Helper function to make Epidata GET requests."""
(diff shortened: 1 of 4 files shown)


nowcast/delphi_nowcast/get_epidata.py
```diff
@@ -57,37 +29,53 @@ def get_indicator_data(sensors: List[SensorConfig],
...
-    for response, sensor, location in responses:
+    all_params = [
+        {"source": "covidcast",
+         "data_source": sensor.source,
+         "signals": sensor.signal,
+         "time_type": "day",
+         "geo_type": location.geo_type,
+         "geo_value": location.geo_value,
+         "time_values": f"{EPIDATA_START_DATE}-{as_of.strftime('%Y%m%d')}",
+         "as_of": as_of.strftime("%Y%m%d")}
```
krivard, line 46: if len(all_combos) is ever likely to be >1000, move these strftime calls outside the loop so it only gets computed once

krivard (approved):
minor efficiency nit, otherwise looks good

nowcast/delphi_nowcast/get_epidata.py
chinandrew: :+1: thanks, updated

PR #813, opened 2021-02-15
Add better no data handling to nowcast

chinandrew:
### Description
When there is no data available in some of the objects, things were crashing, and so added additional handling
cc @mariajahja 

### Changelog
- Add exceptions when no data is present in locationseries
- move a min date statement into a try/except block
- tests

### Fixes 
- Fixes #(issue)


krivard (approved):
👍

Issue #814, opened 2021-02-16
Have `run_module()` take the params as an argument


sgsmob:
Currently, most (all?) of the indicators have very simple `__main__.py` files:
```
from .run import run_module
if __name__ == "__main__":
    run_module()
```
such that all computation is done in `run.run_module()` and its subcalls.  I propose a slight tweak to this:
```
from delphi_utils import read_params
from .run import run_module
if __name__ == "__main__":
    run_module(read_params())
```

This architecture has a few advantages:
- When called from the `IndicatorRunner`, the params can be read once and passed to the indicator, validator, and archiver.
- Documentation of the `params` argument to `run_module()` would describe what is expected and required for `params.json` to have.  Currently, these expectations are often not well documented or out of date.
- The tests would no longer need a `params.json` file.

Progress
 - [x] cdc_covidnet
 - [x] change_hc
 - [x]  claims_hosp
 - [x] combo_cases_and_deaths
 - [x]  covid_act_now
 - [x]  doctor_visits
 - [x] google_health
 - [x] google_symptoms
 - [x]  hhs_facilities
 - [x]  hhs_hosp
 - [x]  jhu
 - [x]  nchs_mortality
 - [x]  quidel
 - [x]  quidel_covidtest
 - [x]  safegraph
 - [x]  safegraph_patterns
 - [x]  usafacts

PR #818, opened 2021-02-16
Refactor params reading in claims hosp

chinandrew:
### Description
Refactor claims hosp so that `run_module()` accepts a `params` argument corresponding to parameters read from the json file. 

### Changelog

### Fixes
-  Partially addresses #814.


claims_hosp/delphi_claims_hosp/run.py
```diff
@@ -11,17 +11,42 @@
...
+               If set to null, end date is derived from drop date and n_waiting_days.
+            - "drop_date": str or null, YYYY-MM-DD format, date data is dropped. If set to
+               null, current day minus 40 hours is used.
+            - "n_backfill_days": int, number of past days to generate estimates for.
+            - "n_waiting_days": int, number of most recent days to skip estimates for.
+            - "write_se": bool, whether to write out standard errors.
+            - "obfuscated_prefix": str, prefix for signal name if write_se is True.
+            - "parallel": bool, whether to update sensor in parallel.
+            - "geos": list of str, geographies to generate sensor for.
+            - "weekday": list of bool, whether to adjust for weekday effects.
```
sgsmob, line 47: Why is this list of bools rather than just a bool?

chinandrew: don't know original design motivation, but the weekday value is [True, False] and then there's a line which is ` for weekday in params["indicator"]["weekday"]:` and loops through a version with and a version without weekday corrections. 

Could probably make the docstring more clear, but now sure what the best wording would be

sgsmob: Maybe "list of bool, which weekday adjustments to perform.  For each value in the list, signals will be generated with weekday adjustments (`True`) or without adjustments (`False`).

chinandrew: :+1: updated, thanks

claims_hosp/tests/test_update_indicator.py
```diff
@@ -16,7 +16,14 @@
 
 CONFIG = Config()
 CONSTANTS = GeoConstants()
-PARAMS = read_params()
+PARAMS = {
+    "indicator": {
+        "input_file": "test_data/SYNEDI_AGG_INPATIENT_11062020_1451CDT.csv.gz",
+        "drop_date": "2020-06-11",
+        "obfuscated_prefix": "foo_obfuscated"
+
```
sgsmob, line 24: drop newline here

sgsmob (approved):
just one nit, lgtm

claims_hosp/tests/test_update_indicator.py
chinandrew: removed

PR #821, opened 2021-02-16
Refactor params reading for CAN

chinandrew:
### Description
Refactor CAN so that `run_module()` accepts a `params` argument corresponding to parameters read from the json file. 

### Changelog
- Remove use of `params.json` in tests, which was never used

### Fixes
-  Partially addresses #814.


covid_act_now/tests/conftest.py
```diff
@@ -20,12 +19,26 @@ def run_as_module():
...
+            "archive": {
+                "cache_dir": "./cache",
+                "bucket_name": "test-bucket",
+                "aws_credentials": {
+                    "aws_access_key_id": "{{ delphi_aws_access_key_id }}",
+                    "aws_secret_access_key": "{{ delphi_aws_secret_access_key }}"
+                }
+            }
+        }
         aws_credentials = params["archive"]["aws_credentials"]
```
sgsmob, line 38: Since we aren't actually doing any archive testing, it might make sense here just to totally remove the whole s3 mocking and the archive parts of the params.

chinandrew: :+1: removed

sgsmob (approved):

PR #822, opened 2021-02-16
Refactor reading params in safegraph

sgsmob:
### Description
Refactor safegraph so that `run_module()` accepts a params argument corresponding to parameters read from the json file. Partially addresses #814.

### Changelog
- Clean up some lint issues in the test
- Change `num_cores` parameter to be an int rather than a string of an int


safegraph/delphi_safegraph/run.py
```diff
@@ -10,15 +10,31 @@
...
+        - "export_dir": str, directory to write output
+        - "log_exceptions" (optional): bool, whether to log exceptions to file
+        - "log_filename" (optional): str, name of file to write logs
+    - "indicator":
+        - "aws_access_key_id": str, ID of access key for AWS S3
+        - "aws_secret_access_key": str, access key for AWS S3
+        - "aws_default_region": str, name of AWS S3 region
+        - "aws_endpoint": str, name of AWS S3 endpoint
+        - "n_core": int, number of cores to use for multithreaded processing
+        - "raw_data_dir": directory from which to read downloaded data from AWS,
```
chinandrew, line 32:
```suggestion
        - "raw_data_dir": str, directory from which to read downloaded data from AWS,
```
```suggestion
        - "raw_data_dir": directory from which to read downloaded data from AWS,
```

chinandrew (approved):
:ship: :ship: :ship:

PR #823, opened 2021-02-16
Refactor params reading for DV

chinandrew:
### Description
Refactor DV so that `run_module()` accepts a `params` argument corresponding to parameters read from the json file. 

### Changelog
- Remove unused static file param and arg.
- 
### Fixes
-  Partially addresses #814.

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -34,7 +34,7 @@ jobs:
       run: |
         make install
     - name: Lint
-      if: ${{ matrix.packages != 'claims_hosp' }}
+      if: ${{ matrix.packages != 'claims_hosp' && matrix.packages != 'doctor_visits' }}
       run: |
         make lint
     - name: Test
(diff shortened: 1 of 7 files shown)


sgsmob (changes requested):
I was wondering why the tests weren't updated but it appears there is no `run_module()` test.  If it's not too much hassle to write a quick one that would be great for this PR, but if not, we should at least refactor the `tests/params.json.template` into a PARAMS variables in `test_run.py` and delete the json.

chinandrew:
> I was wondering why the tests weren't updated but it appears there is no `run_module()` test. If it's not too much hassle to write a quick one that would be great for this PR, but if not, we should at least refactor the `tests/params.json.template` into a PARAMS variables in `test_run.py` and delete the json.

I poked around it a little bit and am hitting some inf values when trying run_module(), so it may require a bit more investigation. Also wasn't enabled in CI, so I went ahead and did that (though linting isnt passing, so skipping it for now). Will open issues for these.

sgsmob (approved):

PR #824, opened 2021-02-16
Refactor params reading for quidel family

sgsmob:
### Description
Refactor `quidel` and `quidel_covidtest` so that `run_module()` accepts a params argument corresponding to parameters read from the json file. Partially addresses #814.

### Changelog
- Clean up some lint issues in the tests
- Change string `mode` parameter to be a bool `test_mode`


ansible/templates/quidel_covidtest-params-prod.json.j2
```diff
@@ -16,6 +16,6 @@
     },
     "bucket_name": "{{ quidel_aws_bucket_name }}",
     "wip_signal": [""],
-    "mode": ""
+    "test_model": false
```
chinandrew, line 19:
```suggestion
    "test_mode": false
```

chinandrew (approved):
:shipit::shipit::shipit:

PR #827, opened 2021-02-16
Refactor params reading for GHT

chinandrew:
### Description
Refactor GHT so that `run_module()` accepts a `params` argument corresponding to parameters read from the json file. 

### Changelog
- Update test to use params directly

### Fixes
-  Partially addresses #814.


google_health/delphi_google_health/run.py
```diff
@@ -27,19 +26,41 @@
...
+            - "log_exceptions" (optional): bool, whether to log exceptions to file
+            - "log_filename" (optional): str, name of file to write logs
+        - "indicator":
+            - "static_file_dir": str, path to DMA mapping files
+            - "data_dir": str, location of cached CSVs
+            - "start_date": str, YYYY-MM-DD format, first day to generate data for
+            - "end_date": str, YYYY-MM-DD format or empty string, last day to generate data for.
+            - "ght_key": str, GHT API key
+            - "wip_signal": list of str or bool, to be passed to delphi_utils.add_prefix
+            - "test": str, whether to run in test mode
```
sgsmob, line 52: I think we can remove this parameter entirely and just use the existence of `test_data_dir` to tell if we are in test mode (make sure to mark it as optional).

sgsmob (changes requested):

google_health/delphi_google_health/run.py
chinandrew: :+1:

sgsmob (approved):

PR #830, opened 2021-02-17
Refactor params reading NCHS

sgsmob:
### Description
Refactor NCHS so that `run_module()` accepts a params argument corresponding to parameters read from the json file. Partially addresses #814.

### Changelog
- Clean up some lint issues in the test
- Fix check that decides whether to run archiver or not
- Alter a few parameters for clarity:
    - `export_dir` -> `weekly_export_dir`
    - `cache_dir` -> `weekly_cache_dir` and move it to the `archive` object
    - `mode` -> `test_file`


nchs_mortality/tests/conftest.py
```diff
@@ -38,9 +59,7 @@ def run_as_module(date):
     with mock_s3():
```
chinandrew, line 59: is it difficult to skip this by removing the "archive" key like we did in some of the other indicators?

sgsmob: I think it is good to not omit this because the daily caching seems to be an integral part of this indicator and it probably won't be superseded by the archiver in the near future.

chinandrew (approved):
:shipit::shipit::shipit:

Issue #825, opened 2021-02-16
Fix linting for claims_hosp and DV


chinandrew:
These are the last two indicators which don't have linting passing and are being skipped in CI.

- [x] claims_hosp
- [ ] doctor_visits

PR #832, opened 2021-02-17
Get claims_hosp to pass lint

sgsmob:
### Description
Update `claims_hosp` and its tests to pass linting.

### Fixes 
- Fixes half of #825


chinandrew (approved):
thanks :shipit::shipit::shipit:

PR #833, opened 2021-02-17
Add special territory population handling to JHU

### LINKED ISSUE #829: Fix JHU US Territories Prop Signals
**Actual Behavior:**

Currently, the prop signals for the US Territories American Simoa, Guam, Virgin Islands, and Northern Mariana Islands give either nan or inf values in JHU.

**Context**

The issue is with JHU's reporting and how we store the population data. For example, doesn't report the FIPS codes

```
60010,23030
60020,1143
60030,0
60040,17
60050,31329
```

But it does report 60000. So the 60000 code gets 0 population (as megafips do) and when doing add_population the counties that do contain pop data don't get added in. So the result is a population of 0 when aggregating to any geo higher than county.

A possible solution is to just divide by a pre-computed state/hhs/nation population for territories.

chinandrew:
### Description

Since JHU only reports territories as megafips and not individual counties, the previous add_population_column call would not work with them, so we need to handle them in a custom way

In addition to the test, ran the indicator and results went for a state confirmed cumulative prop file went from 
`vi,inf,NA,NA` on main
to
`vi,2388.98548,NA,NA` to this branch

### Changelog
- Add special function to JHU to add population for us territories.
- Update test cases

### Fixes 
- Fixes #829


dshemetov (approved):
LGTM!

Issue #838, opened 2021-02-18
Update indicators with NAN codes


dshemetov:
This issue will track the production and the merging of the PRs that add bare bones NAN code support for individual indicators (enable deletions by tracking nan "val" entries and add codes for data ranges). Related work on the DB side [here](https://github.com/cmu-delphi/delphi-epidata/pull/417).

- [x] cdc_covidnet #914 
- [x] change_hc #898 
- [x] claims_hosp #902 
- [x] combo_cases_and_deaths #944 
- [x] covid_act_now #989 
- [x] doctor_visits #990
- [x] ~google_health~
- [x] google_symptoms #993 
- [x] ~hhs_facilities~ skipping for the first release
- [x] ~hhs_hosp~ skipping for first release
- [x] jhu #839
- [x] nchs_mortality #994
- [x] quidel #866 
- [x] quidel_covidtest #999
- [x] safegraph #845 
- [x] safegraph_patterns #859 
- [x] usafacts #843

dshemetov:
Only a partial fix there

nmdefries:
@dshemetov Can we close this?

dshemetov:
Still a bunch of indicator update PRs in the queue from almost 2 years ago. Might still be worth pushing them through at some point, though low priority.

PR #839, opened 2021-02-18
Add NAN code support to JHU

dshemetov:
### Description
Work as part #838. 

### Changelog
- Allow nan values to pass through in the "val" column
- Code Puerto Rico deaths at the county level as a region exception
- Code the first 6 days as "insufficient data" for the 7 day average signal

### Fixes 
- Partially addresses #838


nmdefries (approved):
Looks good!

dshemetov:
JHU doesn't exist anymore.

PR #841, opened 2021-02-18
Combo indicator speedup

chinandrew:
### Description
**Depends on https://github.com/cmu-delphi/covidcast/pull/429** 

Speed up combo indicator by making covidcast calls async. Each particular combo seems to vary quite a bit, but total run time of each loop goes from 10-30s down to ~3s on my laptop.

### Changelog
- Enable async covidcast calls

### Fixes 
- Fixes #(issue)


krivard:
Probably need to add covidcast >= 0.1.4 to setup

chinandrew:
> Probably need to add covidcast >= 0.1.4 to setup

:+1: added, thanks

krivard (approved):
🎉

PR #843, opened 2021-02-18
Add NAN code support to USAFacts

dshemetov:
### Description
Work as part #838. 

### Changelog
- Allow nan values to pass through in the "val" column
- Code the first 6 days as "insufficient data" for the 7 day average signal

### Fixes 
- Partially addresses #838


dshemetov:

PR #844, opened 2021-02-18
[fb survey] `is_selected` to map missing responses to `NA`

nmdefries:
### Description
The `is_selected` function is used to check if a given response code appears in a vector of multi-select responses. It currently maps responses of the format "" (empty string) to `NA`, under the assumption that "" corresponds to a respondent not answering the question. This does match the formatting raw Qualtrics files use, but "" is mapped to `NA` by default by `read_csv` in `load_response_one`. Thus, `is_selected` never sees the "" formatting, and erroneously maps `NA` to `FALSE`.

This results in metrics that use `is_selected` (all indicators derived from item C13) including inappropriate `FALSE` values, which biases percent calculations down. The change here remaps missing responses to `NA`.

### Changelog
- Change `is_selected` to report `NA` when a response is missing (`NA`)
- Add new test for `code_activities`
- Modify existing test for `is_selected` to expect `NA` for `NA` responses, not `FALSE`

### Impact
Historical C13-derived indicators will need to be reissued with the new definition of `is_selected`. Comparing national results for 2021-02-14 using old and new function definitions, we see increases across the board, as expected.

weighted_metric_comparisons

Sample sizes decrease by about 11%, which corresponds well to the 10-15% missingness rate we generally see on the survey.

sample_size_comparisons

### DIFF
--- a/facebook/delphiFacebook/R/variables.R
+++ b/facebook/delphiFacebook/R/variables.R
@@ -19,6 +19,9 @@ split_options <- function(column) {
 
 #' Test if a specific selection is selected
 #'
+#' Checking whether a specific selection is selected in either "" (empty
+#' string) or `NA` responses will produce `NA`s.
+#'
 #' @param vec A list whose entries are character vectors, such as c("14", "15").
 #' @param selection one string, such as "14"
 #' @return a logical vector; for each list entry, whether selection is contained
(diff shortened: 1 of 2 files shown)


facebook/delphiFacebook/R/variables.R
```diff
@@ -27,7 +27,7 @@ is_selected <- function(vec, selection) {
   selections <- sapply(
     vec,
     function(resp) {
-      if (length(resp) == 0) {
+      if (length(resp) == 0 || all(is.na(resp))) {
         # All our selection items include "None of the above" or similar, so
         # treat no selection the same as missingness.
```
capnrefsmmat, line 32: Add a comment here about how Qualtrics files code empty responses, so we remember how this should look?

capnrefsmmat: And maybe something in the `is_selected` docs about how it treats missingness

capnrefsmmat (approved):
Looks good. One documentation suggestion.

Before we merge, we should

1. Coordinate with @krivard on when we can deploy and backfill this (and decide whether to delete old data or not)
2. Draft a message for the mailing list explaining the issue
3. Update your graphs (which are great) to have labels like "Before fix"/"After fix" and dates, so we can attach them to the mailing list message to explain the effect of the change

krivard:
I'll have bandwidth for this Tuesday morning

nmdefries:
@capnrefsmmat Updated graphs above

PR #845, opened 2021-02-19
Add NAN code support to Safegraph

dshemetov:
### Description
Work as part #838. This indicator was not filtering nan values and has no privacy thresholding, so the change was simple.

### Changelog
- Add three missingness columns, default to not missing for "sample_size" and write data insufficient to "stderr" in case it's missing (which occurs only if sample size is 1).
- No region exceptions that I'm aware of

### Fixes 
- Partially addresses #838


dshemetov:
Closed because Safegraph has been deactivated.

PR #850, opened 2021-02-19
Enforce exit logging for every path in quidel

### LINKED ISSUE #689: Add logging to all indicators

benjaminysmith:
### Description
Enforce exit logging for every path in quidel. We discovered in the path where there is no data to process the module exits early. Using atexit ensures the exit code is always called.

### Changelog
- Move exit logging into an atexit registered function.

### Fixes 
#689


chinandrew:
Out of curiosity, where is this exit occurring? The `if _end_date is None:` block?

I wonder if it makes sense to use this exit pattern across all indicators

chinandrew (approved):
:shipit::shipit::shipit:

benjaminysmith:
> Out of curiosity, where is this exit occurring? The `if _end_date is None:` block?
> 
> I wonder if it makes sense to use this exit pattern across all indicators

Yeah that's the one. I am thinking similarly about this pattern. We might be able to register, start the timer, and log the exit just by importing the logger.

PR #852, opened 2021-02-19
[fb survey] Add vaccine hesitancy reasons (items V5abc) to API

nmdefries:
This must be merged *after* PR #844 since the variables here rely on the new, fixed definition of `is_selected`.

Creates marginal hesitancy reasons columns, and uses to produce corresponding smoothed output for API.


facebook/delphiFacebook/R/variables.R
```diff
@@ -225,6 +225,8 @@ code_testing <- function(input_data) {
 #' @param input_data input data frame of raw survey data
 #' @return data frame augmented with `v_covid_vaccinated` and
 #'   `v_accept_covid_vaccine`
+#'
+#' @importFrom dplyr coalesce
```
capnrefsmmat, line 229: you'll want to run `devtools::document()` and commit the `NAMESPACE` change resulting from this as well

capnrefsmmat (commented):
Looks good except for committing NAMESPACE, so I think this will be ready to go soon. Glad you got all the choices coded right and they're consistent across all three questions...

capnrefsmmat (approved):

capnrefsmmat:
@nmdefries Now that #844 is merged, should we go ahead merging this so the new signals can be deployed at the same time as the new wave and other changes we're making now?

nmdefries:
Yep!

PR #855, opened 2021-02-22
Reformat Google Symptoms params to specify values of BigQuery credentials

nmdefries:
### Description
GS params previously asked for the path to BigQuery credentials file, which doesn't work well with Ansible as used in production (specifically, this doesn't allow for encrypting the API key). This change switches from a path to putting BQ credentials directly in the params file, for both local and Ansible params files. Tests and code are updated accordingly.

### Changelog
- Switch from path to credentials file -> credential values in params files
- Add BigQuery credentials file to Ansible vault


krivard (approved):
lgtm!

PR #859, opened 2021-02-22
Add NAN code support to Safegraph Patterns

dshemetov:
### Description
Work as part #838. 

### Changelog
- Adds just very basic nans detection with an unknown code in the vals column

### Fixes 
- Partially addresses #838


dshemetov:

PR #861, opened 2021-02-23
CHNG New file format and location

rumackaaron:
Downloads new counts products files from Change Healthcare in a different folder on the server. All 6 file types are sent to the same folder `countproducts` on the FTP server. They are named `YYYYMMDD_Counts_Products_TYPE.dat.gz`.

The file format has also slightly changed. The first two columns are date then fips instead of fips then date, and the separator is a comma instead of a pipe.

The resulting signals have similar correlations to the ones before. Importantly, the max CLI is 33%, fixing the buggy 99+% values from the old counts.


krivard (commented):
The test .dat files are pretty big -- is all of that data being used?

rumackaaron:
Reduced size

krivard (approved):
👍

PR #863, opened 2021-02-23
Restructure SirCAL to use API data for lag calculation

### LINKED ISSUE #862: Switch SirCAL from metadata to API data for lag
Dashboard viz use the raw API data which means alerts and dashboards can disagree if metadata is out of sync with API data. Switching to use API data should remove this problem.

benjaminysmith:
### Description
Restructure SirCAL to use API data for lag calculation

### Changelog
- Move API retrieval up so that it runs for all signals
- Use the API lag data to generate alert (instead of the metadata lag)

### Fixes 
- Fixes #862

### DIFF
--- a/sir_complainsalot/delphi_sir_complainsalot/check_source.py
+++ b/sir_complainsalot/delphi_sir_complainsalot/check_source.py
@@ -79,18 +79,21 @@ def check_source(data_source, meta, params, grace, logger):
             geo_type=row["geo_type"]
         )
 
-        if latest_data is None:
-            logger.info("No signal data retrieved")
-            continue
+        current_lag_in_days = (now - row["max_time"]).days
+        lag_calculated_from_api = False
 
-        unique_dates = [pd.to_datetime(val).date()
-                        for val in latest_data["time_value"].unique()]
-        current_lag_in_days = (datetime.now().date() - max(unique_dates)).days
+        if latest_data is not None:
+            unique_dates = [pd.to_datetime(val).date()
+                            for val in latest_data["time_value"].unique()]
+            current_lag_in_days = (datetime.now().date() - max(unique_dates)).days
+            lag_calculated_from_api = True
+
         logger.info("Signal lag",
                     current_lag_in_days = current_lag_in_days,
                     data_source = data_source,
                     signal = row["signal"],
-                    geo_type=row["geo_type"])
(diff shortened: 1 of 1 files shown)


sir_complainsalot/delphi_sir_complainsalot/check_source.py
```diff
@@ -103,20 +83,37 @@ def check_source(data_source, meta, params, grace, logger):
...
             continue
 
-        # convert numpy datetime values to pandas datetimes and then to
-        # datetime.date, so we can work with timedeltas after
         unique_dates = [pd.to_datetime(val).date()
                         for val in latest_data["time_value"].unique()]
-        unique_issues = [pd.to_datetime(val).date()
-                        for val in latest_data["issue"].unique()]
-
         current_lag_in_days = (datetime.now().date() - max(unique_dates)).days
```
krivard, line 88: what happens if unique_dates is empty? `google_symptoms` has no data from the last 14 days.

benjaminysmith: Hmmm good point. Ok, wdyt about this version where we fallback to using metadata (and flag this in the log) if we get no data from the API?

Output using the current prod config is shown below:

{"message": "is 77 days old - (last update: 2020-12-08)", "data_source": "google-symptoms", "signal_and_geo_types": "`ageusia_raw_search: [county, hhs, hrr, msa, nation, state]`\n`ageusia_smoothed_search: [county, hhs, hrr, msa, nation, state]`\n`anosmia_raw_search: [county, hhs, hrr, msa, nation, state]`\n`anosmia_smoothed_search: [county, hhs, hrr, msa, nation, state]`\n`sum_anosmia_ageusia_raw_search: [county, hhs, hrr, msa, nation, state]`\n`sum_anosmia_ageusia_smoothed_search: [county, hhs, hrr, msa, nation, state]`\n", "event": "Signal out of SLA", "logger": "delphi_sir_complainsalot.run", "level": "critical", "timestamp": "2021-02-23T17:59:35.978353Z"}

krivard: perfect! for signals that haven't updated in two weeks, it won't matter if meta is a day behind

krivard (approved):
xlnt

PR #866, opened 2021-02-25
Add NAN code support to Quidel

dshemetov:
### Description
Work as part #838. 

### Changelog
- Add three missingness columns
- Add insuficcient data checks and codes

### To Do
- Still needs unknown data codes for when data is missing but not due to insufficient samples sensoring

### Fixes 
- Partially addresses #838


dshemetov:

PR #867, opened 2021-02-25
[fb survey] Support Wave 10

nmdefries:


facebook/delphiFacebook/R/responses.R
```diff
@@ -382,6 +387,7 @@ create_complete_responses <- function(input_data, county_crosswalk)
     "V1", "V2", "V3", "V4_1", "V4_2", "V4_3", "V4_4", "V4_5", # added in Wave 6
     "V9", # added in Wave 7,
     "V2a", "V5a", "V5b", "V5c", "V5d", "V6", "D11", # added in Wave 8
+    "V11", "V12", "V13", "V14_1", "V14_2", # added in Wave 10
```
capnrefsmmat, line 390: We're also going to need C13b and C13c added. Because we're changing that item to be specifically about indoor activities, we've renamed it.

capnrefsmmat (changes requested):
We changed some other column names that will need to be added, but this is good so far

facebook/delphiFacebook/R/variables.R
```diff
@@ -69,6 +71,37 @@ code_activities <- function(input_data) {
...
+    input_data$a_spent_time_indoors_1d <- is_selected(activities, "4")
+    input_data$a_large_event_indoors_1d <- is_selected(activities, "5")
+    
+    transit_responses <- is_selected(activities, "6")
+    if ( is.null(input_data$a_public_transit_1d) ) {
+      input_data$a_public_transit_1d <- transit_responses
+    } else {
+      # Combine with C13 responses
+      input_data$a_public_transit_1d <- coalesce(input_data$a_public_transit_1d, transit_responses)
+    }
```
capnrefsmmat, line 91: `code_activities()` is only ever called on one wave at a time, though that's not actually documented. (You can see `code_mental_health()` has an assert to enforce that.) Shouldn't it then be impossible for `a_public_transit_1d` to already have data in it?

capnrefsmmat: Though if we rely on this fact, we should ensure there's an assert or documentation to enforce it.

capnrefsmmat (approved):
looks good, just is more general than necessary given how the data is coded

facebook/delphiFacebook/R/variables.R
nmdefries: I removed this bit of logic and added the `assert`.

PR #870, opened 2021-02-26
[fb-survey] Add item C14a (7d wear mask) to API

nmdefries:
### Description
Item C14a replaces item C14 starting in Wave 8, asking "In the past 7 days, how often did you wear a mask when in public?" vs "In the past 5 days" on Wave 7.

### Changelog
- Creates `c_mask_often_7d` from item C14a
- Add assert to verify only working with one wave at a time
- Add relevant indicator
- Update tests


nmdefries:
A couple unrelated package functions didn't have updated "man" files, so I added them here. I can split them out into a separate PR if desired.

capnrefsmmat (approved):
Looks good

Issue #826, opened 2021-02-16
Doctor visits tests


chinandrew:
Neither update_sensor nor run_module are tested in the doctor visits package at the moment, and would be good to have that functionality tested. Looks like it follows similar patterns to claims_hosp and changehc, so would reference those for guidance.

PR #871, opened 2021-02-26
Add tests and lint compliance to doctor-visits

### LINKED ISSUE #825: Fix linting for claims_hosp and DV

sgsmob:
### Description
Add tests to `doctor-visits.update_sensor()` and `doctor-visits.run()`.  Also refactors some of the code to make it more easily testable.

### Changelog
- `sensor.py`
    - Change return type to be a `pd.DataFrame` rather than a dictionary containing lists to be traversed simultaneously
    - Stop exposing the `include` output parameter, opting instead to remove the non-included data before returning.  This has the added effect of not outputting any files on dates for which there is no data.
    - Rename input parameters to match the naming conventions of the call sites
- `update_sensor.py`
    - Adapt `write_to_csv()` to handle new output `pd.DataFrame` format
    - Move writing of files out of `update_sensor()`, opting instead to return the associated data frame
    - Consolidate filtering of dates in `final_sensor_idxs` rather than spreading this out
        - This could probably be improved upon even more but it is hard to debug failures since I don't fully understand what the purpose of various index filters are.
- `run.py`
   - Write output of `update_sensor()` at this level

### Fixes 
- Fixes #825 and #826


doctor_visits/tests/test_update_sensor.py
```diff
@@ -0,0 +1,23 @@
...
+            startdate="2020-02-04",
+            enddate="2020-02-05",
+            dropdate="2020-02-06",
+            geo="state",
+            parallel=False,
+            weekday=False,
+            se=False
+        )
+
+        comparison = pd.read_csv(os.path.join("./comparison/update_sensor/all.csv"),
```
chinandrew, line 21: is this `os.path.join()` used for anything?

sgsmob: No, removed.

doctor_visits/delphi_doctor_visits/sensor.py
```diff
@@ -240,4 +242,14 @@ def fit(y_data,
             np.divide((new_rates[include] * (1 - new_rates[include])), den[include]))
 
         logging.debug(f"{geo_id}: {new_rates[-1]:.3f},[{se[-1]:.3f}]")
-        return {"geo_id": geo_id, "rate": new_rates, "se": se, "incl": include}
+
+        included_indices = [x for x in final_sensor_idxs if include[x]]
+
+        def select_idx(my_list):
```
chinandrew, line 248: whats the advantage of making this its own function?

doctor_visits/delphi_doctor_visits/run.py
```diff
@@ -52,28 +50,33 @@ def run_module(params):
...
+    logging.info("last sensor date:\t{enddate}")
+    logging.info("n_backfill_days:\t{n_backfill_days}")
+    logging.info("n_waiting_days:\t{n_waiting_days}")
 
     ## geographies
     geos = ["state", "msa", "hrr", "county"]
 
+
     ## print out other vars
     logging.info("outpath:\t\t%s", params["common"]["export_dir"])
```
chinandrew, line 75:
```suggestion
    logging.info("outpath:\t\t%s", export_dir)
```

doctor_visits/delphi_doctor_visits/sensor.py
sgsmob: It made more sense in an earlier version of this code.  Removed

doctor_visits/delphi_doctor_visits/sensor.py
```diff
@@ -168,7 +169,8 @@ def fit(y_data,
         Args:
             y_data: dataframe for one geo_id, with all 7 cols
             fit_dates: list of sorted datetime for which to use as training
-            sensor_dates: list of sorted datetime for which to produce sensor values
+            burn_in_dates: list of sorted datetime for which to produce sensor values
```
chinandrew, line 172: why the rename to `burn_in` here? perhaps the docstring should be updated since; i associate burn in with values that are used as buffer before producing actual values

sgsmob: The rename reflects that we are calling this with `burn_in_dates` rather than `sensor_dates` in the caller function, but I don't totally understand why the earlier naming convention was used.  I think there is some unnecessary complexity here that is reflected in the numerous reindex operations throughout.  At a high level, it is all around performing the fit over a larger window than we actually want to write output for and accounting for some lags in the data.  I was having trouble disentangling how to migrate indexing, etc. to a single spot so I just submitted.

chinandrew: got it :+1:

chinandrew:
Remove this line  from FI: 
https://github.com/cmu-delphi/covidcast-indicators/blob/dd1798d829195610227df979c32bb03d266865e9/.github/workflows/python-ci.yml#L37

otherwise lgtm

chinandrew (approved):
:shipit::shipit::shipit:

PR #875, opened 2021-02-27
[fb-survey] improve qualtrics error reporting

krivard:
### Description
If we fail to download a Qualtrics file, that's a major problem that should halt the pipeline. This PR ensures that will happen.

### Changelog
- qualtrics.py: throw exception instead of printing error with early return


nmdefries (approved):
Looks good!

PR #876, opened 2021-03-01
Create an indicator runner function

sgsmob:
### Description
Add a utility function that wraps parameter reading, indicator running, validation, and archiving into a single top-level function.

### Changelog
- Add the `delphi_utils.run_indicator_pipeline()` function and its associated tests
- Refactor checking for correctness in `ValidationReport` into a separate function that can be used externally
- Simplify parameter naming convention using "common" rather than "global" everywhere
- Move validation into `delphi_utils` directory
- Rename `_delphi_utils_python/tests/test_utils.py` to have a unique basename from `_delphi_utils_python/tests/validator/test_utils.py` because `pytest` complained
- Do a bunch of lint compliance work in `delphi_utils` because the CI linter was complaining
- Change the Makefile to not double lint the `delphi_utils` directory.  This might actually have made the above lint changes extraneous.


validator/delphi_validator/run.py
```diff
@@ -10,8 +10,17 @@
...
+
+def from_params(params):
+    """Constructs a validator from `params`.
+    Arguments
+    ---------
+    params: Dict[str, Any]
+        Dictionary of parameters
+    """
+    if "validation" in params:
+        return Validator(params["validator"])
```
krivard, line 25: These two keys should probably match.

Also is this intended to have different behavior than run_module? `Validator(read_params())` includes the params under `common` while `Validator(params["validator"])` does not.

validator/delphi_validator/validate.py
```diff
@@ -20,7 +20,10 @@ def __init__(self, params):
         Arguments:
             - params: dictionary of user settings; if empty, defaults will be used
         """
-        suppressed_errors =  params["global"].get('suppressed_errors', [])
+        self.export_dir = params["global"]["export_dir"]
```
krivard, line 23: The production params files vary as to whether they store things like `export_dir` under `global` or `common`, but `common` is used more often:

```
covidcast-indicators$ grep '"common"' ansible/templates/*.j2
ansible/templates/changehc-params-prod.json.j2:  "common": {
ansible/templates/covid_act_now-params-prod.json.j2:  "common": {
ansible/templates/google_health-params-prod.json.j2:  "common": {
ansible/templates/google_health-params-test.json.j2:  "common": {
ansible/templates/google_symptoms-params-prod.json.j2:  "common": {
ansible/templates/hhs_hosp-params-prod.json.j2:  "common": {
ansible/templates/jhu-params-prod.json.j2:  "common": {
ansible/templates/nchs_mortality-params-prod.json.j2:  "common": {
ansible/templates/quidel_covidtest-params-prod.json.j2:  "common": {
ansible/templates/safegraph-params-prod.json.j2:  "common": {
ansible/templates/safegraph_patterns-params-prod.json.j2:  "common": {
ansible/templates/usafacts-params-prod.json.j2:  "common": {
covidcast-indicators$ grep '"global"' ansible/templates/*.j2
ansible/templates/usafacts-params-prod.json.j2:    "global": {
```

validator/delphi_validator/validate.py
```diff
@@ -20,7 +20,10 @@ def __init__(self, params):
...
+
+        validation_params = params["validation"]
```
krivard, line 25: check key name

krivard (changes requested):
Some weirdness about expected key names, and whether to pass in the toplevel params or just the validator ones

validator/delphi_validator/validate.py
sgsmob: There is a difference here between `common` and `validation.global` that is confusing since I think I have made some mistakes between them elsewhere.  I can rename `validation.global` to `validation.common` but it's not clear to me whether that is more or less prone to error.

validator/delphi_validator/run.py
sgsmob: Thanks for the catch, fixed.

sgsmob:
The linting is failing in CI (although not locally) due to not being able to `from delphi_validator.validate import Validator`.  Since I can't replicate this locally, I don't really know how to resolve this.  One possible explanation is a circular import structure at the module level (`delphi_utils.runner` imports `delphi_validator.X` imports `delphi_utils.Y`) although there aren't any cycles at the file level.  If this is the case, it seems we may want to move `validator` to the `delphi_utils` package.  Thoughts?

validator/delphi_validator/validate.py
```diff
@@ -20,7 +20,10 @@ def __init__(self, params):
         Arguments:
             - params: dictionary of user settings; if empty, defaults will be used
```
krivard, line 21: Clarify that the expected params are as read from the params file, not just the `validation` section

validator/delphi_validator/validate.py
krivard: Any of the following is okay with me:
* use the same key in both locations, highlighting the similarity of purpose ("these apply to all siblings")
* use a different key in each location, highlighting the difference of purpose ("these are shared across validators; those are shared across indicator/validation/archiving")
* factor out the key name(s) into a utility function, so that it ceases to matter

...so long as we have enough test coverage to catch any errors in the params files or the code that accesses them before they reach production.

sgsmob: switched to "common" everywhere

validator/delphi_validator/validate.py
sgsmob: done

validator/delphi_validator/validate.py
sgsmob: Done (I think)

krivard:
CI is failing -- probably needs a manual delphi_validator install inside `make install`

sgsmob:
> CI is failing -- probably needs a manual delphi_validator install inside `make install`

I think this doesn't work because of circular dependencies.  Thoughts on moving the validation to be a sub module under `delphi_utils` (i.e. `delphi_utils.validator.validate`)?

krivard:
> Thoughts on moving the validation to be a sub module under `delphi_utils`

in favor, go for it

krivard (approved):
nice work. I've created https://github.com/cmu-delphi/covidcast-indicators/issues/887 to track a future refactor to drop the static geo files, but they're fine where they are for now.

PR #878, opened 2021-03-02
[fb-survey] Switch to filtering with dplyr::filter

nmdefries:
### Description
`filter_responses` and `filter_data_for_aggregation` previously used bracket filtering (`df[logical_row_selection, ]`), which is slow and has unexpected behavior when a variable used for filtering contains `NA` (a missing value). This change swaps in dplyr's `filter` to fix these issues, and adds tests to check filter correctness and behavior with missing values.

### Changelog
- Move all `filter_data_for_aggregatation` filters to using `dplyr::filter`
- Move all `filter_responses` filters to using `dplyr::filter`
- Add tests for both functions

### FIxes
Motivating historical bug via @capnrefsmmat : "we once had a response with DistributionChannel = NA, and it turns out that with [, indexing with a logical vector containing an NA gives you a row that's completely NA... i.e. foo[c(TRUE, NA, FALSE), ] gives a df with two rows, one entirely NA"

### DIFF
--- a/facebook/delphiFacebook/R/responses.R
+++ b/facebook/delphiFacebook/R/responses.R
@@ -185,23 +185,25 @@ load_response_one <- function(input_filename, params) {
 #' @param params named list containing values "static_dir", "start_time", and
 #'   "end_time"
 #'
-#' @importFrom dplyr anti_join
+#' @importFrom dplyr anti_join filter
 #' @importFrom rlang .data
 #' @export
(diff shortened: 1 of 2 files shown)


capnrefsmmat (approved):
Looks good; hope this avoids any issues. I assume you ran the full test suite (including the gold tests) to ensure this didn't change behavior on any existing inputs

nmdefries:
Yup, all the tests are fine!

PR #880, opened 2021-03-02
[fb-package] Reduce memory usage by dropping unused columns

nmdefries:
### Description
We carry many extraneous columns along during the aggregation process. Removing those and only keeping variables directly used in calculations plus supporting columns (zip code, day) should reduce memory usage.

Tests pass as expected; pipelines before and after this change produce the same output.

### Changelog
- Use `select` to keep necessary columns in `aggregate_indicators`
- Auto-format + spelling fixes (second commit; not substantive)


capnrefsmmat (approved):
looks good, although I was getting used to those spelling mistakes

krivard:
conflicts?

capnrefsmmat:
easy fix (@nmdefries should confirm I resolved correctly)

nmdefries:
Looks good

PR #885, opened 2021-03-04
Update chng smoothing and denom thresholding

### LINKED ISSUE #881: changehc indicator outputs no data for hhs, nation when data is present at lower geo levels
[[Slack] context](https://delphi-org.slack.com/archives/C01LZ3A2UMU/p1614698588017700?thread_ts=1614663716.017100&cid=C01LZ3A2UMU)

In the changehc indicator, it seems to be possible for county/state/hrr data to pass minimum denominator thresholds when hhs/nation data does not. Our current hypothesis is that this is due to smoothing numerators and denominators separately and after geo aggregation: the smoother is being overly enthusiastic about curve fitting, such that some counties/state/hrrs may be concave-up when the majority of their siblings are concave-down, causing the overall curve for hhs or nation to fall below threshold.

There are several suspicious things going on here:
* Smoothing after geo aggregation (though doctor-visits also does this)
* Smoothing the numerator and denominator separately before computing the rate
* Backfill causes the counts to be artificially low for the most recent period of data, and the smoother is probably not a good tool for modeling this effect
* ~Zero-filling (iirc the smoother expects NaNs, not zeros)~ (this turns out to be true zeros -- by definition CHNG does not report if the count is 0)
* No corner-case tests, only completion tests
* ~[how did this ever work](https://github.com/cmu-delphi/covidcast-indicators/blob/ca43cda9ffd3e321369ad7911c5d8f44a1feb70a/changehc/delphi_changehc/sensor.py#L152)~ (turns out pandas assigns NaNs for the remaining indices, which is fine)

To fix:
* Write a test that fails when the relationship between having-data is not hierarchical
* Consider smoothing ratios instead of numerators/denominators
* Switch to using raw denominators for thresholding
* Consider whether backfill is adding 2x smoothing
* Consider reducing the bandwidth? though this would have other ramifications

[Sample data](http://rinkitink.ml.cmu.edu/delphi/misc/bad_spiky_business.csv)

chinandrew:
### Description
See #881 for context. Still needs to be run end to end to see how the signal values change with this PR vs what's in main.

A few examples of this PR vs main:
73b60a53-3b45-4bef-8c2b-4e01fb53ac20
7fc32240-e28b-4b2e-bb6b-8d2b0b4b9023
4b32a1b8-f45b-445c-93ca-745795aa8146

### Changelog
- Smooth the ratio instead of num/den separately
- Change the min den thresholding to use the raw totals and not smoothed totals
- Add a test case which previously failed (had county output data but not states, see test comment for more detail) but now passes.
- Update sensor tests to not use strict inequalities. since smoothing happens after the ratio is taken, it is now possible for the more recent values to end up as negative and be clipped to 0, whereas previously the ratio is done after smoothing and had 0.5 added to the numerator and 1 added to the denominator, so the ratio was always nonzero

### Fixes 
- Fixes #881


changehc/tests/test_sensor.py
```diff
@@ -69,5 +69,5 @@ def test_fit_fips(self):
                 # binomial standard error, hence largest possible value is
                 # 100 * (0.5 / sqrt(MIN_DEN))
                 assert np.nanmax(res0["se"]) <= 100 * (0.5 / np.sqrt(Config.MIN_DEN))
-                assert np.nanmin(res0["se"]) > 0
+                assert np.nanmin(res0["se"]) >= 0
```
krivard, line 72: Why is this necessary? With Jeffreys we shouldn't have se==0...

changehc/tests/test_update_sensor.py
```diff
@@ -81,29 +81,52 @@ def test_geo_reindex(self):
...
                 self.parallel,
                 self.weekday,
                 self.numtype,
                 self.se,
                 ""
             )
-            su_inst.update_sensor(self.small_test_data,  td.name)
+            # As of 3/3/21 (40c258a), this set of data has county outputting data, state and hhs not
+            # outputting data, and nation outputting data, which is undesirable. Ideal behaviour
+            # should be all output or only a larger region outputting if the subregions also output,
```
krivard, line 111:
```suggestion
            # should be all output or a subregion only outputting if its parent has output,
```

krivard (changes requested):
Suspicious se test, & also inverted logic on an explanatory comment

changehc/tests/test_sensor.py
chinandrew: What's Jeffreys?

For this case, sometimes rate = 0 (due to the explanation in the description), and then the se becomes 0 since its `rate(1-rate)`

krivard: The Jeffreys prior for the binomial distribution adds a pseudosample to the data: half a count to each category, or +1 total. In theory it guarantees we never have a rate or se of 0.

It's applied here in the chng indicator:
https://github.com/cmu-delphi/covidcast-indicators/blob/85e1925c0d0c606651b303a09af6f7602e36037e/changehc/delphi_changehc/sensor.py#L123-L124

but it looks like that's done before smoothing. We should double-check with a statistician whether we should move it to after smoothing or not.

chinandrew: who'd be best to consult on this with? 

If we move it to after smothing, isn't it possible we'd have a bunch of really low signals spike to 0.5/1 when they should be close to 0?

rumackaaron: Maria added the Jeffreys prior to the dv signal, where it was done before smoothing. I think we're getting significantly weirder behavior with the Change smoother, so maybe it's better to add it after smoothing to prevent the prior from getting smoothed out of existence. It's possible that we'd have low signals spike to 0.5/1 but we wouldn't be reporting those because the denominators would be way below threshold.

capnrefsmmat: I think it should be done *after* smoothing; doing it before smoothing is like adding half a count per day, instead of adding half a count per seven days, and hence biases the estimates more.

Since it's adding a half-count, its effect is small when the sample size is large. It should never change the estimates very much, unless you are reporting estimates when there are only two or three underlying data points.

chinandrew: How would we add after smoothing? e.g. say it was originally 20/100=0.8, and it gets smoothed to 0.76 , would we do something like (76+0.5) / (100 + 1) = 0.757?

krivard: Like a weighted average of (unsmoothed denominator) * (smoothed rate) and (1) * (0.5)? That seems reasonable, but also makes me wonder if that was the initial motivation behind smoothing the numerator and denominator separately. I'm not sure which approach is the most statistically correct, though I have a slight preference for doing the weighted average as you suggest, with the smoothed rate and the unsmoothed sample size.

chinandrew: oh i didn't even think of the smooth vs unsmoothed denom -- I guess using 100 as the raw denom wasn't the most unambiguous choice. I was originally thinking of using the smoothed denom (so if it was 10/50 as the original ratio, you'd still get the same answer), though you're right that there's a few ways we could do this.

capnrefsmmat: > How would we add after smoothing? e.g. say it was originally 20/100=0.8, and it gets smoothed to 0.76 , would we do something like (76+0.5) / (100 + 1) = 0.757?

That's how the survey does it:
https://github.com/cmu-delphi/covidcast-indicators/blob/e7c7c13ef988d29e8d56409db8418a1de8a062ec/facebook/delphiFacebook/R/binary.R#L195-L206

chinandrew: cool, lets go with that then.

chinandrew: ok one slight hiccup. previously when we did it, we couldn't div/0 since the denominator would always be at least one. If we change to calculating the rate, smoothing, then adding the jeffreys at the end, we run the risk of getting nan values in the first step if denom = 0. Not sure how we want to deal with this, especially given @dshemetov's work on nan coding.

dshemetov: Just to add on that last point: on my skim through the code, I didn't see any censorship in the sense of writing NANs when values are below a threshold. The backfill function seems to have a dual function of ensuring reasonable sample sizes and ensuring privacy by padding with time-averaging. So it seems to me that the only way to have a denominator of 0 would be to have a region with a week of 0 reports (which should be rare) or the same thing on the left boundary (more likely). Please correct me if I'm wrong on this mental model. @chinandrew could you point me to where the thresholding chucks the 0 denoms?

changehc/delphi_changehc/sensor.py
```diff
@@ -120,30 +120,13 @@ def fit(y_data, first_sensor_date, geo_id, num_col="num", den_col="den"):
...
-
-        # checks - due to the smoother, the first value will be NA
-        assert (
-                np.sum(np.isnan(smoothed_total_rates[1:])) == 0
-        ), "NAs in rate calculation"
-        assert (
-                np.sum(smoothed_total_rates[1:] <= 0) == 0
-        ), f"0 or negative value, {geo_id}"
+        rates = total_counts.flatten() / total_visits
+        smoothed_rate = CHCSensor.smoother.smooth(rates)
```
rumackaaron, line 124: Do we want to rewrite and call `CHCSensor.gauss_smooth` here? If not, should we just delete that function? This is the only place it would be called.

chinandrew: :+1: removed the gauss smoothed function

krivard (approved):
nice work. @rumackaaron are we good to merge?

dshemetov:
fwiw, I doubled checked too and it looked good to me

rumackaaron:
Looks good to me

PR #886, opened 2021-03-04
Fix bug in sircal where a single day of data will fail when checking gaps

benjaminysmith:
### Changelog
- When gap days is empty, do not try to calculate the max gap.


chinandrew:
Should it log something before continuing? (or instead set gap = 0 and run the rest of the code as normal)

krivard:
@benjaminysmith ping

benjaminysmith:
> Should it log something before continuing? (or instead set gap = 0 and run the rest of the code as normal)

Added a log line.

chinandrew (approved):
:shipit::shipit::shipit:

PR #888, opened 2021-03-05
Re-bodge the V4 bodge

capnrefsmmat:
If the wave has V4a but not V4, then input_data[["V4_1"]] would be NULL.
This caused the ifelse() to fail unless something was done to make it
non-NULL -- such as there being non_english responses, we would assign
NA to the column in that case.

Once we got a data file with no non_english responses, the pipeline
failed.

To avoid this, always assign NA to those columns if we're not doing so
already.


nmdefries (approved):
This looks good, but considering the error, perhaps we should also add tests of this function for when we have only English and only non-English input responses.

krivard:
I'm going to merge now so I can finish today's pipeline run, but please do add tests for that in a second PR

PR #893, opened 2021-03-08
Set minimum epidata version for combo cases/deaths

krivard:
### Description
Enforces at least delphi-epidata 0.0.12 in combod cases/deaths setup.py. This *should* cut down on the unreliability & content-type pipeline failures we've been seeing.

### Changelog
- setup.py


krivard:
combo cases deactivated

PR #894, opened 2021-03-09
[fb-survey] Add "indoors" version of work/school item

nmdefries:
Created the supporting variable in #867 but forgot to create the indicator. This is part of the set derived from C13b, the new "indoors" version of C13 (activities).


capnrefsmmat (approved):
looks good

PR #895, opened 2021-03-09
changehc testing bug

### LINKED ISSUE #896: Random changehc testing bug
Fix a small heisenbug in `test_fit_fips` in `test_sensor.py`.

dshemetov:
### Description
The test `test_fit_fips` in `test_sensor.py` has inconsistent behavior, occasionally passing and sometimes failing. To reproduce, run `pytest test_sensor.py` a few times, it took me ~10 before a failing case. 

The issue is that the function tests a random subset of the fips contained in a test data set. Some of these return all nans for the values, leading to an `assert np.nan > 0` check which fails (i.e. the fips 01039).

### Changelog
This fix removes the randomization component by testing a fixed subset of the data and adds a check to ignore fips where all entries are nan.

### Fixes 
Fixes #896.


changehc/tests/test_sensor.py
```diff
@@ -53,12 +53,14 @@ def test_backfill(self):
     def test_fit_fips(self):
         date_range = pd.date_range("2020-05-01", "2020-05-20")
         all_fips = self.combined_data.index.get_level_values('fips').unique()
-        sample_fips = nr.choice(all_fips, 10)
+        sample_fips = all_fips[:50]
```
rumackaaron, line 56: If we ignore fips that have all nan for rate, is this line necessary?

krivard:
If we're only testing 50 fips, can we get rid of the remaining unused portion of the test dataset?

If we expect behavior X when data is all nan and behavior Y when data is not all nan, should we not verify that both behaviors occur under the correct circumstances?

changehc/tests/test_sensor.py
dshemetov: Probably not necessary, but non-deterministic tests just seem like a bad idea to me in general.

dshemetov:
Sure, I'll remove the rest of the data set. The all-nan case really is a strange edge-case: we are selecting a data range for which this fips has no values, so the num and den values are all 0 and then we're running a regression fit on it, which just returns all zeros. That seems sensible, so the only real issue is that we have an assert statement assuming it can do comparison ops on nans.

dshemetov:
I could also just remove any fips that don't have values in the given date range and the all-nans case wouldn't occur.

krivard:
Since we're likely to encounter the all-nans edge case in the wild at some point, I'd prefer the following:
* split whatever test data we intend to use into not-all-nans and all-nans
* check that the not-all-nans data generates the right fits
* check that the all-nans data generates all-nans

dshemetov:
Sounds good!

dshemetov:
Updated the dataset to only include Alabama fips codes 01XXX, which includes at least one all-nans case and many non-all-nans cases. Made a check to verify that the all-nans case has all include flags set to false.

krivard:
Thanks! Would you merge main into this? It looks like you branched off of 7b7da79 instead of main and it's making the diffs hard to read.

dshemetov:
Ah sure thing!

dshemetov:
Done. Andrew's changes are still here, but my updates are just in `test_sensor.py`.

changehc/tests/test_sensor.py
```diff
@@ -53,15 +53,19 @@ def test_backfill(self):
     def test_fit_fips(self):
         date_range = pd.date_range("2020-05-01", "2020-05-20")
         all_fips = self.combined_data.index.get_level_values('fips').unique()
-        sample_fips = nr.choice(all_fips, 10)
+        sample_fips = all_fips
 
         for fips in sample_fips:
```
krivard, line 58:
```suggestion
        for fips in all_fips:
```

krivard (approved):
a wee nit, otherwise 👍

PR #897, opened 2021-03-09
[fb-package] Specify environment timezone

nmdefries:
### Description
Explicitly sets R environment timezone to LA/Pacific time.

### Fixes 
Fixes potential problem where we could get different output if we run the pipeline on a computer with a different system timezone. Based on some profiling, a small but not insignificant amount of run time is taken up by checking the computer's timezone setting, so this change stands to improve run time slightly.

### Impact
Currently, the FB pipeline runs on a machine in Pittsburgh (Eastern time), so this change may impact indicator values. TBD: what is the magnitude of this change?

### DIFF
--- a/facebook/delphiFacebook/R/dates.R
+++ b/facebook/delphiFacebook/R/dates.R
@@ -1,5 +1,6 @@
 # Time zone to use throughout package.
 tz_to <- "America/Los_Angeles"
+Sys.setenv(TZ=tz_to)
 
 #' Get the date of the first day of the previous month.
 #'


nmdefries:
This approach does not have the expected behavior, that is, runs on a computer using Eastern time and on a computer using Eastern time but with environment timezone explicitly set to Pacific time produce the same output. It may be that the environment timezone is somehow not broad enough in scope.

Given a move towards running the survey pipeline and tables in a Docker container, in which we can set the container timezone explicitly, setting the R environment timezone seems unnecessary.

PR #900, opened 2021-03-10
[fb-package] Logs + some additional parallelization

nmdefries:
### Description
Allows use of parallel `apply` functions when parallel functionality is enabled at selected bottlenecks, based on profiling the contingency table generation code. Adds basic logs to help debugging and track progress on long runs.

### Changelog
- Allows parallelization on CSV read, multi-select item code remapping (only used for contingency tables), and `is_selected`
- Adds logs
- Corrects definitions of a few derived variables


facebook/delphiFacebook/R/contingency_run.R
```diff
@@ -197,28 +215,46 @@ run_contingency_tables <- function(params) {
...
-  if (!is.null(params$end_date) & !is.null(params$n_periods)) {
-    ## Produce historical CSVs
+  if (!is.null(params$n_periods)) {
+    msg_plain(paste0("Producing CSVs for ", params$n_periods, " time periods"))
 
     if (params$aggregate_range == "week") {
       period_step <- days(7)
     } else if (params$aggregate_range == "month") {
       period_step <- months(1)
     } else if (is.null(params$aggregate_range)) {
```
capnrefsmmat, line 225: maybe this should just be an `else`, because if `params$aggregate_range == "ducks"`, the `period_step` will never be set and this will go badly

facebook/delphiFacebook/R/contingency_run.R
```diff
@@ -197,28 +215,46 @@ run_contingency_tables <- function(params) {
...
+      ymd(params$end_date) - period_step * seq(0, params$n_periods - 1)
+    ))
 
     for (end_date in end_dates) {
       period_params <- params
+      
+      # Update start/end date and time.
       period_params$end_date <- end_date
-      run_contingency_tables_many_periods_one_period(period_params, aggregations)
+      if ( !(end_date == end_dates[1]) ) {
```
capnrefsmmat, line 242:
```suggestion
      if ( end_date != end_dates[1] ) {
```

facebook/delphiFacebook/R/responses.R
```diff
@@ -190,6 +198,7 @@ load_response_one <- function(input_filename, params) {
 #' @importFrom rlang .data
 #' @export
 filter_responses <- function(input_data, params) {
+  msg_plain(paste0("Filtering data..."))
```
capnrefsmmat, line 201: all of these `msg_plain` calls with just one string don't technically need `paste0`, since you're not concatenating... doesn't matter much though

capnrefsmmat (changes requested):
Looks good overall -- my suggestions are mainly minor cleanups. What kind of speed improvements were you seeing in your testing?

facebook/delphiFacebook/R/responses.R
nmdefries: whoops, that can be attributed to some over-zealous copy-pasting

nmdefries:
> What kind of speed improvements were you seeing in your testing?

These made the contingency tables something like 50% faster on my desktop, if I recall correctly -- I haven't worked with this branch in a while. Katie's machine, where I'm running the bulk of these, has pretty slow single-threading, so improvements there may have been better.

capnrefsmmat (approved):

PR #901, opened 2021-03-10
[fb-package] Report C14a (7-day mask item) in microdata

nmdefries:
C14 ("In the past 5 days, how often did you wear a mask when in public?") was replaced starting in Wave 8 with C14a ("In the past 7 days..."), and needs to be included in the microdata. Derived indicators have already been added.


capnrefsmmat (approved):
Looks good

capnrefsmmat:
@krivard This is a column we've been missing in the microdata since Feb 8, if it's possible to merge and reissue.

PR #905, opened 2021-03-11
[fb-package] Add V6 to API

nmdefries:
Add V6-derived indicators (reasons for not believing respondent needs a COVID vaccine) to API.


capnrefsmmat (approved):
Code looks good. But we do need to be aware that this is treating blank responses as missing, i.e. assuming that if no reasons apply, the respondent will pick "Other". Just like in the other items where we assume they'll pick "None of the above".

I don't think we have an alternative here, but we'll want to include this fact in the API docs

PR #906, opened 2021-03-11
[combo cases/deaths] restrict issues fetched

### LINKED ISSUE #903: Consider limiting the issues pulled by combo cases/deaths
Currently, combo cases/deaths pulls all dates and uses the archive differ to reduce the results to only the new and changed rows. As we go forward in time that becomes a harder ask for the server, with millions of rows returned for each county signal.

However, we don't need to pull the latest issue for all dates -- just the issues that have been added since the last time combo was run. We should pad that range to account for runs where one or the other cases/deaths source fails, but it should still substantially reduce the amount of data the server needs to assemble. This might be enough on its own to reduce the frequency of 502 Bad Gateway error responses to tolerable levels.

krivard:
### Description

We're currently generating random 502 Bad Gateway errors on combo cases runs. One hypothesis is that at this point, combo cases pulls so many rows (~1M or more) that some of the replicas are getting overtaxed and the load balancer gives up on them.

This PR will generate the same number of _queries_, but the vast majority of them should return zero results. It also adds an optional config param if you for some reason want to manually specify the range of issues, or switch back to fetching the most recent issue for each date pulled no matter how old it is.

The default behavior will only fetch issues from the last week or so.

### Changelog
- generalize date range calculations (new|all|yyyymmdd-yyyymmdd) in their own function
- specify `issues` in calls to `covidcast.signal`
- drop now-superfluous "Fetching jhu-csse|usa-facts..." printcruft

### Fixes 
- Fixes #903

### DIFF
--- a/combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
+++ b/combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
@@ -86,7 +86,7 @@ def compute_special_geo_dfs(df, signal, geo):
     return df
 
 
-def combine_usafacts_and_jhu(signal, geo, date_range, fetcher=covidcast.signal):
+def combine_usafacts_and_jhu(signal, geo, date_range, issue_range=None, fetcher=covidcast.signal):
     """Add rows for PR from JHU signals to USA-FACTS signals.
 
     For hhs and nation, fetch the county `num` data so we can compute the proportions correctly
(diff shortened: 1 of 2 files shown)


chinandrew:
Do any params files need to be updated with this?

krivard:
No, the default behavior is the desired production behavior, and it won't hurt dev usage either:

https://github.com/cmu-delphi/covidcast-indicators/blob/a8df19b74d8a9f4f75f66283f3d7150f797af12f/combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py#L195

chinandrew (approved):
:shipit::shipit::shipit:

PR #908, opened 2021-03-12
Stop sending slack messages directly from Sircal

benjaminysmith:
Remove the slack channel from the sircal config since we are alerting through ELK monitoring.


krivard (approved):
lgtm; just for my future self's sanity here's the guard that makes this effective

https://github.com/benjaminysmith/covidcast-indicators/blob/a8ed4a193e7a4db4a74e11c63444733cd3de9f74/sir_complainsalot/delphi_sir_complainsalot/run.py#L31

PR #914, opened 2021-03-12
Add NAN code support to CDC CovidNet

dshemetov:
### Description
Work as part #838. 

### Changelog
- Add three missingness columns, default to not applicable for "sample_size" and "stderr" columns
- Code nan values in "val" column as unknown for now

### Fixes 
- Partially addresses  #838


dshemetov:
@kyminbb oh, never mind on the review, looks like we don't publish this indicator.

dshemetov:

PR #916, opened 2021-03-12
Change CAN export dir to dashes instead of underscores

chinandrew:
### Description
Dashes are consistent with the rest of the sources.

### Changelog
- Update ansible params.json to use receiving folder with dashes.


krivard (approved):
👍

PR #917, opened 2021-03-12
[fb-package] Improvements in speed and clarity to contingency pipeline

nmdefries:
### Description
Decrease memory usage by filtering out low-sample size groups before aggregations are produced.

### Changelog
- Create megacounties and for all geolevels drop too-small groups *before* aggregation step. Reduces resource use when spawning threads.
- Replaces bracket filtering with `dplyr::filter` for speed and intuitive handling of `NA`s


nmdefries:
Changes originally introduced here were separated out into #952 and #934 for shorter and more focused PRs.

PR #926, opened 2021-03-16
Add slack channel back again

benjaminysmith:
This is a temporary fix to get logging back for ELK. Proper fix will decouple from the slack channel.

### Fixes 
- Temp fix for #925


krivard (approved):
👍

PR #928, opened 2021-03-16
[fb-survey] Support new wording of V4a "local health" item

nmdefries:
### Description
Create new indicator for changed wording of V4a subitem 2.

### Fixes
In Wave 8, the subitem was switched from asking about "Local health workers" to "Doctors and other health professionals you go to for medical care". This resulted in a sharp increase in reported trust that is misleading when the new and old versions are reported under the same signal name.

### Impact
@krivard We will need to reissue survey data going back to Feb 8, the Wave 8 release date, to incorporate this change.


capnrefsmmat (approved):
looks good

Issue #868, opened 2021-02-25
Add GitHub workflow for survey indicator unit tests


capnrefsmmat:
It'd be good to run the survey indicator unit tests in CI so we don't break it, since it's... easy to break. The tests are pretty slow to run, though, so we may need to either make them faster or change something about how the testing works so they're practical to run on CI.

A similar workflow for the covidcast R package: https://github.com/cmu-delphi/covidcast/blob/main/.github/workflows/r_covidcast_ci.yml

nmdefries Assigning to you, but I think this is a longer-term goal, so you can think about it once you have the current data demands out of the way

nmdefries:
#1081 (and some followups) adds unit tests to CI. These test individual functions and don't rely on running the actual data generation, which is the slow part. Integration tests, which check validity of data generation output, were excluded from CI due to long set-up time. For reference, tests take ~15 min to run in GitHub Actions. About 2 minutes of that is on unit tests and 13 minutes on integration tests.

Without changing the integration tests themselves, it seems like the only way to feasibly run them in CI is to do so only when the survey code has changed.

krivard:
Two ways to do that:
* action workflows can be configured to run only when a path has been modified; see [the delphi_utils one for an example](https://github.com/cmu-delphi/covidcast-indicators/blob/48c1ca058faeb5881493487dcb6405ffe9f527ce/.github/workflows/release-delphi-utils.yml#L3-L9). this may run them more often than we'd like.
* set up a special branch that we PR into to stage survey code changes, and have a workflow that runs only when that branch is modified. we'd have to remember to use that branch whenever we want the integrations to run, but it would let us make unrelated changes in the survey tree without the 15 minute penalty.

PR #930, opened 2021-03-16
[fb-survey] Support new wording of V4a "local health" item followup

nmdefries:
Forgot supporting wave variable in #928.


capnrefsmmat (approved):
aha, another reason to add CI in #868

PR #932, opened 2021-03-17
[Draft] Combo indicators bug fix

### LINKED ISSUE #921: Combo cases/deaths reporting nonsensical values
Combo cases is having some serious trouble. Here is a plot of nation data:

image

At first I thought this was a problem with switching the queries from default versioning (as_of=today) to a limited issue range (from "last update day minus a week" to "today"). Switching to a limited issue range was necessary to reduce the load on the API server pool and cut down on 502 errors, which it did.. However, any region that updated more than once during the issue range would have >1 row in the result, since the API returns all issues requested. I fixed that in #910/#918 by dropping all but the most recent issue for each date+region, but that didn't resolve the behavior depicted in the plot above.

My new hypothesis is that when we aggregate up geographically, the only regions in the df are the ones that were changed. If this is indeed what's happening, then we should do something like this to make sure we're getting a fully-populated map for all updated dates:

```
changes = covidcast.signal(..., issues=issue_range)
changed_dates = changes.time_value.unique()
query_date_pairs = date_list_to_start_end_pairs(changed_dates)
dfs = []
for start,end in query_date_pairs:
  dfs.append(covidcast.signal(..., start_day=start, end_day=end))
df = pandas.concat(dfs)
```

which is, yknow, pretty gross, but I'm not sure what else to do. 

and then we'll have to figure out how to handle the 7dav signals...

dshemetov:
### Description
Fix #921.

### Changelog
Implements @krivard's suggestion
```
changes = covidcast.signal(..., issues=issue_range)
changed_dates = changes.time_value.unique()
query_date_pairs = date_list_to_start_end_pairs(changed_dates)
dfs = []
for start,end in query_date_pairs:
  dfs.append(covidcast.signal(..., start_day=start, end_day=end))
df = pandas.concat(dfs)
```
The way lag is handled needs to be fixed (currently being totaled by geo aggregation).

### Fixes 
Fix #921.

### DIFF
--- a/combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
+++ b/combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
@@ -14,6 +14,7 @@ import time
 
 import covidcast
 import pandas as pd
+from numpy import timedelta64
 
 from delphi_utils import add_prefix, get_structured_logger
 from delphi_utils.geomap import GeoMapper
@@ -86,15 +87,49 @@ def compute_special_geo_dfs(df, signal, geo):
     return df
 
 
-def combine_usafacts_and_jhu(signal, geo, date_range, issue_range=None, fetcher=covidcast.signal):
-    """Add rows for PR from JHU signals to USA-FACTS signals.
+def merge_dfs_by_geos(usafacts_df, jhu_df, geo_to_fetch):
+    """Combines the queried usafacts and jhu dataframes based on the geo type."""
+    # State level
+    if geo_to_fetch == 'state':
+        combined_df = maybe_append(
+            usafacts_df,
+            jhu_df if jhu_df is None else jhu_df[jhu_df["geo_value"] == 'pr']) # add territories
+    # County level
+    elif geo_to_fetch == 'county':
+        combined_df = maybe_append(
+            usafacts_df,
+            jhu_df if jhu_df is None else jhu_df[jhu_df["geo_value"].str.startswith("72")])
+    # For MSA and HRR level, they are the same
+    elif geo_to_fetch == 'msa':
+        # df = GMPR._load_crosswalk("fips", "msa")
+        # puerto_rico_fips = ["72" + str(i).zfill(3) for i in range(999)]
+        # puerto_rico_msas = df[df["fips"].isin(puerto_rico_fips)]["msa"].unique()
+        puerto_rico_msas = [
+            '10380',
+            '11640',
+            '25020',
+            '32420',
+            '38660',
+            '41900',
+            '41980',
+            '49500'
+        ]
+        combined_df = maybe_append(
+            usafacts_df,
+            jhu_df if jhu_df is None else jhu_df[jhu_df["geo_value"].isin(puerto_rico_msas)])
+    else:
+        combined_df = usafacts_df
+    combined_df.rename(COLUMN_MAPPING, axis=1, inplace=True)
 
-    For hhs and nation, fetch the county `num` data so we can compute the proportions correctly
-    and after combining JHU and USAFacts and mapping to the desired geos.
-    """
+    return combined_df
+
+
+def get_updated_dates(signal, geo, date_range, issue_range=None, fetcher=covidcast.signal):
+    """Return the unique dates of the values that were updated in a given issue range in a geo."""
     is_special_geo = geo in ["hhs", "nation"]
     geo_to_fetch = "county" if is_special_geo else geo
     signal_to_fetch = signal.replace("_prop", "_num") if is_special_geo else signal
+
     usafacts_df = fetcher(
         "usa-facts", signal_to_fetch,
         date_range[0], date_range[1],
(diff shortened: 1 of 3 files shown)


combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
```diff
@@ -86,15 +86,49 @@ def compute_special_geo_dfs(df, signal, geo):
...
+        puerto_rico_msas = [
+            '10380',
+            '11640',
+            '25020',
+            '32420',
+            '38660',
+            '41900',
+            '41980',
+            '49500'
+        ]
```
krivard, line 115: what's going on here? i'm not super hyped about hard-coded MSAs, since if we ever need to fix them in the geo utils then we'll have to remember to fix them here too

krivard: (would it be easier/faster/clearer to have GMPR just add a fips column to the msa data, then filter by fips?)

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
```diff
@@ -107,25 +141,56 @@ def combine_usafacts_and_jhu(signal, geo, date_range, issue_range=None, fetcher=
...
+    return unique_dates
+
+
+def combine_usafacts_and_jhu(signal, geo, date_range, issue_range=None, fetcher=covidcast.signal):
+    """Add rows for PR from JHU signals to USA-FACTS signals.
+
+    For hhs and nation, fetch the county `num` data so we can compute the proportions correctly
+    and after combining JHU and USAFacts and mapping to the desired geos.
+    """
+    is_special_geo = geo in ["hhs", "nation"]
```
krivard, line 161: this occurs at least twice; probably time to put it into a function

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
```diff
@@ -107,25 +141,56 @@ def combine_usafacts_and_jhu(signal, geo, date_range, issue_range=None, fetcher=
...
+    geo_to_fetch = "county" if is_special_geo else geo
+    signal_to_fetch = signal.replace("_prop", "_num") if is_special_geo else signal
+
+    unique_dates = get_updated_dates(signal, geo, date_range, issue_range, fetcher)
+    # This occurs if the usafacts ~and the jhu query were empty
+    if unique_dates is None:
         return pd.DataFrame({}, columns=COLUMN_MAPPING.values())
+    # Requery each date with as_of=today, so that every geo is represented
+    query_date_pairs = [(date, date) for date in unique_dates]
```
krivard, line 170: this is probably fine but we should check the running time; worst case scenario this would generate 365 calls to fetcher instead of 1 if all dates in the past year were touched by some issue

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
```diff
@@ -107,25 +141,56 @@ def combine_usafacts_and_jhu(signal, geo, date_range, issue_range=None, fetcher=
...
+            start, end,
+            geo_to_fetch,
+        )
+        jhu_df = fetcher(
+            "jhu-csse", signal_to_fetch,
+            start, end,
+            geo_to_fetch,
+        )
+        dfs.append(merge_dfs_by_geos(usafacts_df, jhu_df, geo_to_fetch))
+    # Always non-empty, since the dates were already present in at least on query
```
krivard, line 184: 💯

combo_cases_and_deaths/tests/test_run.py
```diff
@@ -47,38 +48,51 @@ def test_unstable_sources(mock_covidcast_signal):
...
+        elif geo == "msa":
+            geo_val = "10580"
+        else:
+            geo_val = "36001"
         return pd.DataFrame(
-            [(date.fromordinal(c[0]),"ny" if geo == "state" else "36000",1,1,1)],
+            [(date.fromordinal(c[0]),geo_val,1,1,1)],
             columns="time_value geo_value value stderr sample_size".split())
     def make_mock(geo):
+        # The first two in each row provide the date for the query of the latter two
```
krivard, line 72: 💯

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
dshemetov: yea, so GMPR doesn't have ways to go from msas to fips. i hardcoded these because the alternative was to use a private function in gmpr (commented code). not sure which of the two we'd rather have

krivard: oh shoot, we'd need the weights to invert it wouldn't we.

I think I prefer the private function, but you shouldn't need to generate all 999 PR fips codes, just filter by fips.startswith("72")

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
dshemetov: ah, looks like i didn't actually need it the second time.

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
dshemetov: the runtime is definitely a concern. I will test something like 100 calls to fetcher for 100 separate dates vs 1 call to fetcher for a range of 100 dates.

krivard: (it will matter what geo you use there -- fetcher will automatically split county queries into one API call per day, but it bundles state/hhs/nation queries in bigger batches)

dshemetov: oof, ok so the new method of calling fetcher once for each date seems to be 10x slower than just calling a whole range. for both state and county geos. the performance gap shrinks a little when there's sparsity in `unique_dates`, but even for `date_range = pd.date_range("2020-08-1", "2021-03-1", freq='W')`, fetching the whole range in a single call is ~4 times faster than querying those individual days.

so i think it makes more sense to do a single fetch call to `unique_dates.min(), unique_dates.max()` instead.

combo_cases_and_deaths/delphi_combo_cases_and_deaths/run.py
dshemetov: yea, the weights would be tricky since fips -> msa is an incomplete mapping (not defined for all fips).

got it, good call on `startswith`

krivard (approved):
Looks good!

krivard:
Is this still draft or ready to merge?

dshemetov:
Let me double check and run this PR on the past 2 months and make sure the numbers look reasonable.

dshemetov:
The error reported (nation):
image

Same data in the new PR:
image

I added a new param called `issue_days` so that we can choose how far back in the issues we should look for a date range. This should help if we want to reissue values older than a week, which we'll probably have to judging by the first plot here (the default is still 7 though).

dshemetov:
Should be good to go.

krivard:
Can you paste the params.json you used to generate the plot above? For some reason prod is producing the same tiny numbers as before.

krivard:
Never mind, I'm looking at `prop` instead of `num`

PR #934, opened 2021-03-17
[fb-package] Usability improvements + fix to hesitant_and_* vars

nmdefries:
### Description
General usability improvements to contingency pipeline. Decrease memory usage. Improve clarity and conciseness. Fix some variable definitions (will need to reissue Box tables).

### Changelog
- Initially filter data to only keep columns used in aggregations.
- Filter list of files from `params$input` to leave out unnecessary files, i.e. if a file doesn't contain any responses within the requested aggregation date range. Assumes standard CSV name format.
- Introduce new params setting `use_input_asis` to suppress above feature. Necessary if inputting files with non-standard names or if want to load all data files regardless of specified start and end dates.
- Simplify code that handles removing aggregations when necessary columns are missing.
-  Fix definitions of `hesitant_and_*` variables -- previously denominator was all responses to V3 (vaccine acceptance item), but the intent was to do the calculation *among* hesitant responses only. (This change has already been made in the report data and other data sent to FB.)
- Add default weekly aggs (same as default monthly aggs for now).


krivard:
looks like you ran into a conflicting change in contingency_variables.R

facebook/delphiFacebook/R/contingency_aggregate.R
```diff
@@ -158,37 +167,43 @@ post_process_aggs <- function(df, aggregations, cw_list) {
...
-    if ( is.null(df[[col_var]]) ) {
-      aggregations <- aggregations[aggregations$metric != col_var &
-                                     !mapply(aggregations$group_by,
-                                             FUN=function(x) {col_var %in% x}), ]
-      msg_plain(
-        paste0(
-          col_var, " is not defined. Removing all aggregations that use it. ", 
-          nrow(aggregations), " remaining")
-      )
+  group_cols <- unique(do.call(c, aggregations$group_by))
```
capnrefsmmat, line 172: hmm, why `do.call`? Do you want `c(aggregations$group_by, recursive = TRUE)`?

capnrefsmmat: and how does this variable differ from `group_vars`?

capnrefsmmat (commented):
just one question; otherwise I think this looks fine

capnrefsmmat (approved):

PR #938, opened 2021-03-19
update usafacts to use new lag style

sgsmob:
### Description
Update `usafacts` validation to use the conventions introduced in #915.

### Fixes 
- Partial step for #725

### DIFF
--- a/ansible/templates/usafacts-params-prod.json.j2
+++ b/ansible/templates/usafacts-params-prod.json.j2
@@ -19,7 +19,7 @@
     "common": {
       "data_source": "usa-facts",
       "span_length": 14,
-      "end_date": "latest",
+      "end_date": "today",
       "suppressed_errors": [
         {"check_name": "check_val_lt_0"},
         {"check_name": "check_test_vs_reference_avg_changed",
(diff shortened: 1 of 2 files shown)


jingjtang (approved):
lgtm

ansible/templates/usafacts-params-prod.json.j2
```diff
@@ -19,7 +19,7 @@
     "common": {
       "data_source": "usa-facts",
       "span_length": 14,
-      "end_date": "latest",
+      "end_date": "today",
```
krivard, line 22:
```suggestion
      "end_date": "today-2",
```

usafacts/params.json.template
```diff
@@ -19,7 +19,7 @@
     "common": {
       "data_source": "usa-facts",
       "span_length": 14,
-      "end_date": "latest",
+      "end_date": "today",
```
krivard, line 22:
```suggestion
      "end_date": "today-2",
```

krivard (commented):
Update reflecting min_lag pulled from metadata (which has some signals at min_lag 1, some signals at min_lag 2, guessing we'd rather use the worst case there?)

sgsmob:
> Update reflecting min_lag pulled from metadata (which has some signals at min_lag 1, some signals at min_lag 2, guessing we'd rather use the worst case there?)

Thank you for making these suggestions on this PR and the others.  Done

PR #939, opened 2021-03-19
Create preliminary validation params for safegraph_patterns

sgsmob:
### Description
Create a preliminary set of validation params for `safegraph_patterns` and turn on running the validation in production.

### Changelog
Validation file turns off validation for all signals in `safegraph` since they share a common `data_source` in the API. 

### Fixes 
- Partially addresses #725

### DIFF
--- a/ansible/templates/safegraph_patterns-params-prod.json.j2
+++ b/ansible/templates/safegraph_patterns-params-prod.json.j2
@@ -18,7 +18,7 @@
     "common": {
       "data_source": "safegraph",
       "span_length": 14,
-      "end_date": "today",
+      "end_date": "today-4",
       "suppressed_errors": [
         {"signal": "completely_home_prop"},
         {"signal": "completely_home_prop_7dav"},
--- a/ansible/templates/safegraph_patterns-params-prod.json.j2
+++ b/ansible/templates/safegraph_patterns-params-prod.json.j2
@@ -14,7 +14,30 @@
     "sync": true,
     "wip_signal" : []
   },
-  "archive": {
-    "cache_dir": "./cache"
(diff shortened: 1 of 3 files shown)


nmdefries (commented):
From #915 it looks like `end_date` should be of the format `today-<n>`; I think `today` will cause an error.

sgsmob:
> From #915 it looks like `end_date` should be of the format `today-<n>`; I think `today` will cause an error.

Good catch, fixed in #943

nmdefries (approved):

ansible/templates/safegraph_patterns-params-prod.json.j2
```diff
@@ -14,7 +14,30 @@
     "sync": true,
     "wip_signal" : []
   },
-  "archive": {
-    "cache_dir": "./cache"
+  "validation": {
+    "common": {
+      "data_source": "safegraph",
+      "span_length": 14,
+      "end_date": "today",
```
krivard, line 21:
```suggestion
      "end_date": "today-4",
```

safegraph_patterns/params.json.template
```diff
@@ -13,7 +13,30 @@
     "aws_endpoint": "",
     "sync": true
   },
-  "archive": {
-    "cache_dir": "./cache"
+  "validation": {
+    "common": {
+      "data_source": "safegraph",
+      "span_length": 14,
+      "end_date": "today",
```
krivard, line 20:
```suggestion
      "end_date": "today-4",
```

krivard (commented):
Updates to end_date referencing min_lag from meta

PR #940, opened 2021-03-19
[fb-package] Add screening test positivity to API

nmdefries:
Report test positivity rate among those who were tested for screening purposes, and had no symptoms or prior risky behavior


facebook/delphiFacebook/R/variables.R
```diff
@@ -280,6 +280,40 @@ code_testing <- function(input_data) {
...
+    input_data$t_tested_reason_crowd <- is_selected(testing_reasons, "6")
+    input_data$t_tested_reason_visit_fam <- is_selected(testing_reasons, "7")
+    input_data$t_tested_reason_other <- is_selected(testing_reasons, "8")
+    
+    input_data$t_tested_reason_screening <- case_when(
+      input_data$t_tested_reason_sick == TRUE ~ 0,
+      input_data$t_tested_reason_contact == TRUE ~ 0,
+      input_data$t_tested_reason_crowd == TRUE ~ 0,
+      
+      input_data$t_tested_reason_medical == TRUE ~ 1,
```
capnrefsmmat, line 301: pretty sure that in 650, Taylor would have told you to do

```suggestion
      input_data$t_tested_reason_medical ~ 1,
```

because you don't need to compare against booleans like this. But I think it is actually more readable this way, for this specific case. If you agree, I'm fine with leaving it

capnrefsmmat (approved):
Looks good, although I wonder if the logic is complicated enough that you want to add a test for it (particularly for empty answers or for ones with multiple selections) just in case we screw it up later

facebook/delphiFacebook/R/variables.R
nmdefries: Good point. I've been using this format since some of our binary variables are coded as numeric rather than explicit logical and cause an error. I'm fine with either method.

capnrefsmmat: I think we can leave these for clarity, but set the `is.na` one below to use `!is.na()` instead. Somehow it reads better to me

nmdefries:

PR #941, opened 2021-03-19
Preliminary validation params for safegraph

sgsmob:
### Description
Create a preliminary set of validation params for `safegraph` and turn on running the validation in production.

### Changelog
Validation file turns off validation for all signals in `safegraph_patterns` since they share a common `data_source` in the API. 

### Fixes 
- Partially addresses #725

### DIFF
--- a/ansible/templates/safegraph-params-prod.json.j2
+++ b/ansible/templates/safegraph-params-prod.json.j2
@@ -14,7 +14,31 @@
     "sync": true,
     "wip_signal" : []
   },
-  "archive": {
-    "cache_dir": "./cache"
+  "validation": {
+    "common": {
+      "data_source": "safegraph",
+      "span_length": 14,
+      "end_date": "today",
+      "suppressed_errors": [
+        {"signal": "bars_visit_num"},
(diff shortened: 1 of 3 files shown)


ansible/templates/safegraph-params-prod.json.j2
```diff
@@ -14,7 +14,31 @@
     "sync": true,
     "wip_signal" : []
   },
-  "archive": {
```
benjaminysmith, line 17: Why does the archive parameter get unset?

sgsmob: As far as I can tell it is not being used and when we get the indicator runner up and running it will trigger running the archiver before it is configured.

krivard: confirmed; there is no safegraph prefix in the S3 bucket we use for archiving.

benjaminysmith: Thanks.

benjaminysmith (approved):

PR #943, opened 2021-03-19
Allow "today" as an option for lag in validator

sgsmob:
### Description
#915 introduced the ability to lag days the validator checks the recently produced data, but inexplicably required the user to specify using today's data as "today-0".  This PR allows "today" as a valid option.


dshemetov (approved):
Looks good.

PR #944, opened 2021-03-20
Add NAN code support to Combo Indicator

dshemetov:
### Description
Work as part #838. 

### Changelog
- Allow nan values to pass through in the "val" column (in anticipation of those columns being returned from the covidcast client)
- Add three missingness columns, default to NA for "sample_size" and "stderr" columns, set "val" to either NOT_MISSING or to UNKNOWN; if we want to be more granular, we will need to think through how geo aggregation should interact with missing codes 

### Fixes 
- Partially addresses #838


krivard:
this PR outlived the actual combo indicator 😳

PR #945, opened 2021-03-21
add vi/gu to US state count

mariajahja:
### Description
Drops of hosp claims data contain observations from VI/GU. Increase the number of possible states to account for this. 

### Changelog
- One line change in `config.py`


krivard (approved):
👍

krivard:
(at some point we may want to consider dropping those constants and using `delphi_utils.geomap.GeoMapper. get_geo_values()`, but this is fine for now)

PR #952, opened 2021-03-22
[fb-package] Change megacounty behavior

nmdefries:
### Description
Decrease memory usage by filtering out low-sample size groups before aggregations are produced. Large numbers of aggregation groups, especially a problem at the county level, cause `mclapply` to eat up all the memory and slow way down.

Remove megacounty creation for contingency tables -- megacounties were originally intended to make the map look nice, but are probably not of particular interest in finer-level aggregations and can be backed out from state + rest-of-counties data if desired.

### Changelog
- For all geolevels drop too-small groups *before* aggregation step. Reduces resource use when spawning threads.
- Stop creating megacounty values for contingency tables.
- Add test to check county contingency tables.


facebook/delphiFacebook/R/contingency_aggregate.R
```diff
@@ -231,17 +231,21 @@ post_process_aggs <- function(df, aggregations, cw_list) {
 #'   "num_filter"
 #'
 #' @importFrom dplyr inner_join bind_rows
-#' @importFrom parallel mclapply
+#' @importFrom parallel mclapply splitIndices
```
capnrefsmmat, line 234: Is `splitIndices` used anywhere? I don't see it in this diff

capnrefsmmat (approved):
I think this looks good; one nitpick with imports, but otherwise great

PR #953, opened 2021-03-22
combo cases: 

### LINKED ISSUE #947: Combo cases/deaths should not publish dates where only JHU is available
Because combo cases/deaths re-aggregates HHS and nation geo levels, any dates where only JHU is available wind up with only puerto rico counting toward HHS=2 and nation=us, which is more wrong than not publishing anything for those dates.

Adjust the null data frame checker to generate the following behavior:

| USAFacts | JHU | result |
|-----------|-----|-----------|
|  data | data | usafacts + jhu |
| data | null | usafacts |
| null | data | ~jhu~ **null** |
| null | null | null |

krivard:
### Description
If JHU data extends farther into the future than USAFacts data, stop where USAFacts stops.

### Changelog
- run.py - multiple logic changes to discard scenarios when we only have JHU data
- test_run.py - fix existing tests to account for differences in fetcher call counts, add test for nation when JHU extends past USAFacts

### Fixes 
- Fixes #947

### DIFF
--- a/combo_cases_and_deaths/tests/test_run.py
+++ b/combo_cases_and_deaths/tests/test_run.py
@@ -235,7 +235,6 @@ def test_no_nation_jhu(mock_covidcast_signal):
         20200101,
         "county"
     )
-    #assert False, result
     pd.testing.assert_frame_equal(
         result,
         pd.DataFrame({"timestamp":[20200101],
--- a/combo_cases_and_deaths/tests/test_run.py
+++ b/combo_cases_and_deaths/tests/test_run.py
@@ -2,7 +2,7 @@
 from datetime import date
 from itertools import product
 import unittest
-from unittest.mock import patch
+from unittest.mock import patch, call
 import pandas as pd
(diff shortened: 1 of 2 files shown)


combo_cases_and_deaths/tests/test_run.py
```diff
@@ -206,6 +206,45 @@ def test_combine_usafacts_and_jhu_special_geos(mock_covidcast_signal):
...
+    result = combine_usafacts_and_jhu("_num", "nation", date_range=(0, 1), fetcher=mock_covidcast_signal)
+
+    assert mock_covidcast_signal.call_args_list[-1] == call(
+        "jhu-csse",
+        "_num",
+        20200101,
+        20200101,
+        "county"
+    )
+    #assert False, result
```
dshemetov, line 238: Probably don't need this?
```suggestion
```

dshemetov (approved):
Looks good after a minor nit.

Issue #955, opened 2021-03-23
CDC COVIDnet tests failing


krivard:
cdc_covidnet tests are failing in main with 

```
raise ConnectionError(e, request=request)
E           requests.exceptions.ConnectionError: HTTPSConnectionPool(host='gis.cdc.gov', port=443): Max retries exceeded with url: /grasp/covid19_3_api/GetPhase03InitApp?appVersion=Public (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f1cb39899d0>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))
```

I've opened a PR to drop this indicator from CI (since it's never been deployed) but we should still fix this test to mock out all network calls. Unit tests should have repeatable results regardless of the day they are run and the status of the network.

PR #956, opened 2021-03-23
CI: Drop cdc_covidnet

krivard:
### Description
cdc_covidnet was in CI for some reason, though it's not a production indicator and we have no immediate plans to release it. Normally this wouldn't be a problem, but #955 is causing checks to fail, so we should drop it for the time being.

### Changelog
- python-ci.yml


chinandrew (approved):
:shipit::shipit::shipit:

PR #959, opened 2021-03-23
Remove unnecessary second loop of signals

### LINKED ISSUE #768: Remove hardcoded signal names in validation
https://github.com/cmu-delphi/covidcast-indicators/blob/10af56f3dd3bfdff3f6b3a233727f3b225a422e1/validator/delphi_validator/dynamic.py#L121

These signal names should be dynamically generated so this validation works with other indicators besides JHU

sgsmob:
### Description
Remove a bug in the validator where some signals were hardcoded to loop over.  See #768 

### Fixes 
- Fixes #768


chinandrew (approved):
:shipit::shipit::shipit:

PR #960, opened 2021-03-23
Use geomapper to validate geo value correctness

sgsmob:
### Description
Migrate validation to use geomapper rather than hardcoded local files when checking validity of geo values.  This both makes our system more robust and also deals with an annoying issue of specifying the path to the old local files.


rumackaaron (approved):
lgtm

PR #961, opened 2021-03-25
Allow more expressive geo value specification in validation

sgsmob:
### Description
Allow a wider array of geo values to be accepted as valid in the validator

### Changelog
- Users can specify `additional_valid_geo_values` for each geo type to be accepted as valid
- State level FIPS values are accepted as valid for counties


_delphi_utils_python/delphi_utils/validator/static.py
```diff
@@ -130,6 +134,22 @@ def check_df_format(self, df_to_test, nameformat, report):
...
+        if geo_type == "state":
+            geomap_type = "state_id"
+        elif geo_type == "county":
+            geomap_type = "fips"
+        else:
+            geomap_type = geo_type
+
+        gmpr = GeoMapper()
+        valid_geos = gmpr.get_geo_values(geomap_type)
+        valid_geos |= set(self.params.additional_valid_geo_values.get(geo_type, []))
```
chinandrew, line 148: today I learned what `|=` does, cool

_delphi_utils_python/tests/validator/test_static.py
```diff
@@ -237,6 +237,20 @@ def test_empty_df(self):
...
 
+    def test_state_level_fips(self):
+        validator = StaticValidator(self.params)
+        report = ValidationReport([])
+        df = pd.DataFrame(["37183", "56000", "04000", "60000", "78000"], columns=["geo_id"])
+        validator.check_bad_geo_id_value(df, FILENAME, "county", report)
+
+        assert len(report.raised_errors) == 0
+
+        df = pd.DataFrame(["37183", "56000", "04000", "79000" "60000", "78000"], columns=["geo_id"])
```
chinandrew, line 248: Is there supposed to be a comma between "79000" and "60000"? or is this intentionally a typo. If the former, perhaps do something like 

```suggestion
        df = pd.DataFrame(["37183", "56000", "04000", "60000", "78000", "99000"], columns=["geo_id"])
```
so the first 5 are the same as the valid values in the example above, and the last one is the new and more obviously invalid value

_delphi_utils_python/tests/validator/test_static.py
```diff
@@ -302,6 +316,31 @@ def test_invalid_geo_id_value_nation(self):
         assert len(report.raised_errors) == 1
         assert report.raised_errors[0].check_name == "check_bad_geo_id_value"
 
+    def test_additional_valid_geo_ids(self):
+        params = self.params.copy()
+        params["static"] = {
+            "additional_valid_geo_values": {
+                    "state": ["confusion"],
+                    "county": ["left", "right"]
```
chinandrew, line 324: nit: could you name these something like "additional_county1" so the below dataframe values are fairly clear

sgsmob: done

chinandrew: update state name too? otherwise lgtm

chinandrew (approved):
:shipit::shipit::shipit:

PR #963, opened 2021-03-25
[fb-package] Add in-person schooling aggs to API

nmdefries:
### Description
Adds percent of people with children attending in-person school full-time and percent with children attending part-time.


nmdefries:
Haven't heard back from the ICL team yet on verifying the definition. Should we wait?

nmdefries:
Okay, this is good to go!

facebook/delphiFacebook/R/variables.R
```diff
@@ -471,3 +471,32 @@ code_vaccines <- function(input_data) {
...
+#'
+#' @param input_data input data frame of raw survey data
+#' @return data frame augmented with `hh_number_total`
+code_schooling <- function(input_data) {
+  if ("E2_1" %in% names(input_data)) {
+    # Coded as 2 = "Yes", 3 = "No", 4 = "I don't know"
+    input_data$s_inperson_school_fulltime <- case_when(
+      input_data$E2_1 == 2 ~ 1,
+      input_data$E2_1 == 3 ~ 0,
+      TRUE ~ NA_real_
```
capnrefsmmat, line 485: Given the discussion we just had on "Don't know" answers, some thoughts. For this one, "I don't know" could be:

* school is changing rules or things are uncertain
* temporary break means they don't know what to answer
* respondent confused by the question

I think it makes sense to exclude "I don't know" here, but you could argue the other way. What's a typical rate of "I don't know" for this question?

capnrefsmmat (approved):
Looks good; we'll just want to document the "I don't know" behavior clearly, or maybe run it past Sarah. The global instrument doesn't have this item, so there's no problem with consistency, but maybe she has an opinion on the strategy we should use

facebook/delphiFacebook/R/variables.R
nmdefries: "I don't know" accounts for about 0.75-1% of responses.

capnrefsmmat: Ok, so I'm pretty comfortable ignoring it

nmdefries: Sounds good.

krivard:
It's totally feasible, only question is whether to support as_of for those past dates or whether we can dump all the past dates in e.g. tomorrow's issue

nmdefries:
> whether to support as_of for those past dates or whether we can dump all the past dates in e.g. tomorrow's issue

Hm, I'm leaning towards putting all the past dates in a single issue, e.g. tomorrow's, since this is the first time these values are being reported (vs. recreating the issue dates when fixing some bug for values that had been issued before). But I'm not sure if people are using the `as_of` feature and would find it useful to have here.

@capnrefsmmat thoughts?

capnrefsmmat:
I think that's fine with me.

PR #964, opened 2021-03-25
Add null check names to `safegraph*` validations

sgsmob:
### Description
`safegraph` and `safegraph_patterns` validation was dying in production (no consequences) due to the fact that the required `check_name`s for the `suppressed_errors` were missing.  These suppression rules were intended to block validation failures due to the fact that `safegraph` and `safegraph_patterns` both are part of data source `safegraph` in the API; without suppression, we would have failures from `safegraph` not producing the signals from `safegraph_patterns` and vice-versa.

Since `null`/`None` values are used as wildcards, we add an explicit `null` to each such suppression entry.


sgsmob:
Closed in favor of #965

PR #965, opened 2021-03-25
Simplify syntax to suppress all validation checks for a signal/geo/time

sgsmob:
### Description
#939 and #941 specified that all checks corresponding to a particular set of signals should be suppressed by adding `{"signal": "$SIGNAL"}` to the `validation.common.suppressed_errors` lists of the params files.  This syntax, however, is not valid and causes the validator to die.  Instead of "fixing" the params files by adding `"check_name": null` to each entry (see #964), this PR changes the validator params syntax to allow the aforementioned specifications to work as intended.

### Changelog
- `check_name` argument to `ValidationFailure` constructor is now optional and set to None when omitted.


chinandrew (approved):
:shipit::shipit::shipit:

PR #967, opened 2021-03-26
Enable setting end date in Google Symptoms

nmdefries:
### Description
The GS pipeline previously assumed that `export_end_date` was "now", which was inconvenient for retrieving a small set of historical data (e.g. for filling in missing data in the API). This adds a new setting to the params file to allow data to be fetched between two specific dates.

### Changelog
- Supports new params setting `export_end_date`; defaults to "today" if not set
- Updates tests.


google_symptoms/delphi_google_symptoms/pull.py
```diff
@@ -94,7 +94,7 @@ def preprocess(df, level):
     return df
 
 
-def get_date_range(export_start_date, num_export_days):
+def get_date_range(export_start_date, export_end_date, num_export_days):
     """Produce date range to retrieve data for.
```
krivard, line 99: Clarify what `end date ("now")` means in the below

krivard (approved):
Minor documentation nit, otherwise good

krivard:
... that _really_ should not have caused tests to start failing. Is something mocked incorrectly?

nmdefries:
A new version of pylint came out today... It appears to have a bug where it doesn't correctly load the pylintrc file.

PR #972, opened 2021-03-29
Add --dry_run flag to validator

### LINKED ISSUE #969: HHS Hosp fails when validation fails
a la

```
[...]
Number of expected geo region-signal combinations: 12
937 checks run
49 checks failed
0 checks suppressed
0 warnings

make: *** [run] Error 1
   Fail [cmd=sudo -u indicators -s bash -c "cd /home/indicators/runtime/hhs_hosp && make run"]
```

I've patched in prod by removing validation from the `run` target, but we'll need a more permanent fix before the next time we deploy prod -- probably just need to modify the validation line in the target to ignore nonzero exit status.

sgsmob:
### Description
Running the validator from the command line now has the option of turning off failing with non-zero exit status when failures are raised.

### Changelog
Add `--dry_run` flag to `delph_utils.validator` that, when invoked, exits with success regardless of the number of failures.

### Fixes 
- Fixes #969

### DIFF
--- a/_delphi_utils_python/delphi_utils/validator/report.py
+++ b/_delphi_utils_python/delphi_utils/validator/report.py
@@ -92,7 +92,7 @@ class ValidationReport:
 
     def print_and_exit(self, die_on_failures=True):
         """Print results and exit.
-        
+
         Arguments
         ---------
         die_on_failures: bool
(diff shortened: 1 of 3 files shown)


eujing (approved):
Looks good to me!

PR #973, opened 2021-03-29
Google Symptoms uses metadata to determine how many days of data to fetch

nmdefries:
### Description
Instead of pulling a fixed 14 days of data, determine number of days to pull based on metadata `min_lag` value. Pre-existing functionality (`num_export_days` can be set in params to an integer or to "all") still works.

### Changelog
- Pull GS-specific `min_lag` from metadata if `num_export_days` is not set in params.
- Remove 14-day `num_export_day` setting from template and Ansible params files.

### Fixes 
Makes GS pipeline more robust to BigQuery outages. Minimize data pulled in anticipation of change to daily updates.

### DIFF
--- a/google_symptoms/delphi_google_symptoms/run.py
+++ b/google_symptoms/delphi_google_symptoms/run.py
@@ -7,6 +7,7 @@ when the module is run with `python -m delphi_google_symptoms`.
 import time
 from datetime import datetime
 from itertools import product
+import covidcast
 
 import numpy as np
 from delphi_utils import (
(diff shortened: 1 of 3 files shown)


google_symptoms/delphi_google_symptoms/run.py
```diff
@@ -46,7 +47,14 @@ def run_module(params):
     export_start_date = datetime.strptime(
         params["indicator"]["export_start_date"], "%Y-%m-%d")
     export_dir = params["common"]["export_dir"]
-    num_export_days = params["indicator"].get("num_export_days", "all")
+
+    if "num_export_days" in params["indicator"]:
```
jingjtang, line 51: Why we have such a check here?

google_symptoms/delphi_google_symptoms/run.py
```diff
@@ -46,7 +47,14 @@ def run_module(params):
...
+        num_export_days = params["indicator"]["num_export_days"]
+    else:
+        # Get number of days based on what's missing from the API.
+        metadata = covidcast.metadata()
+        gs_metadata = metadata[(metadata.data_source == "google-symptoms")]
+        num_export_days = max(gs_metadata.min_lag)
```
jingjtang, line 57: Does it mean we would never have backfill data for Google Symptoms signals?

nmdefries: Yeah, as far as we can tell, they don't have any backfill in their system so we don't need to re-fetch data that we already have.

google_symptoms/delphi_google_symptoms/run.py
jingjtang: Isn't the `num_export_days` deleted from the `params` file? If you want to check this, I think it is better to keep the `num_export_days` term in `params` but set it to `None` in default. Otherwise, this check does not make sense to me.

jingjtang (approved):
lgtm

PR #976, opened 2021-03-30
Update params documentation

sgsmob:
### Description
Add documentation about expected `params.json` hierarchy compatible with indicator runner.

### DIFF
--- a/_template_python/delphi_NAME/run.py
+++ b/_template_python/delphi_NAME/run.py
@@ -13,7 +13,6 @@ following structure:
             True if all signals in the registry are works in progress, or False if only
             unpublished signals are.  See `delphi_utils.add_prefix()`
         - Any other indicator-specific settings
-If the indicator is m
 """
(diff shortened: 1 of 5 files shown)


_template_python/delphi_NAME/run.py
```diff
@@ -2,20 +2,34 @@
 """Functions to call when running the function.
 
 This module should contain a function called `run_module`, that is executed
-when the module is run with `python -m MODULE_NAME`.
+when the module is run with `python -m MODULE_NAME`.  Its lone argument should be a nested
```
chinandrew, line 5: The "lone argument" here is the argument to run_module() and not to the command line command right? 

```suggestion
when the module is run with `python -m MODULE_NAME`.  `run_module`'s lone argument should be a nested
```

chinandrew (approved):
:shipit::shipit::shipit:

sgsmob:
The added line is too long; pushing a fix in a sec.

PR #977, opened 2021-03-30
[fb-package] Allow custom aggs to override in-package aggs

nmdefries:
### Description
Allow user to provide an R script defining new aggs to produce, overriding default aggs included in the package. Move the default aggs to separate file to improve readability of `contingency_run.R`.

### Changelog
- Check for new params setting `aggs_in`, specifying a file to read custom aggs from. If defined, use instead of default aggs.
- Create file `contingency_indicators.R` to  store default aggs.

### Fixes
Lets custom/one-off tables be run on the main branch.


facebook/delphiFacebook/R/contingency_indicators.R
```diff
@@ -0,0 +1,276 @@
...
+#'
+#' Compute functions must be one of the `compute_*` set (or another function
+#' with similar format can be created). Post-processing functions should be one
+#' of the `jeffreys_*` set or post_convert_count_to_pct or the identity `I`,
+#' which does not modify the data.
+#'
+#' @return named list
+#' 
+#' @importFrom tibble tribble
+set_aggs <- function() {
```
capnrefsmmat, line 23: `get_aggs` seems like a more natural name

facebook/delphiFacebook/R/contingency_run.R
```diff
@@ -286,7 +8,16 @@ set_aggs <- function() {
 #'
 #' @export
 run_contingency_tables <- function(params) {
-  aggs <- set_aggs()
+  if ( !is.null(params$aggs_in) && file.exists(params$aggs_in) ) {
```
capnrefsmmat, line 11: it probably would save some confusion if you issued a warning if the parameter is not null but the file does not exist

facebook/delphiFacebook/R/contingency_run.R
```diff
@@ -286,7 +8,16 @@ set_aggs <- function() {
...
+    # Run non-default aggregates. File should create an object called `aggs`.
+    source(params$aggs_in)
+    
+    if ( !exists("aggs") || all(class(aggs) != "data.frame") ) {
```
capnrefsmmat, line 15:
```suggestion
    if ( !exists("aggs") || !inherits(aggs, "data.frame") ) {
```

capnrefsmmat (changes requested):
Some minor suggestions. I'm not a huge fan of sourcing a bare R file as configuration, but until we convert this to some kind of static configuration file, I guess it's the best we can do.

nmdefries:
> I'm not a huge fan of sourcing a bare R file as configuration

Yeah, it's kludgy and has security problems. The intention of this feature is for me to make custom tables for people without needing to make a new branch or make edits to the package itself, so I'll be running it on my own computer using files that I've created. I don't think we'd ever use this in an automated way, so I'm not super concerned about the security aspect.

As an alternative, I was considering having the aggs be set in the params file, but that requires additional processing and potentially validation steps. Plus the lists of aggregates can get really long (at least for FB).

capnrefsmmat (approved):

PR #988, opened 2021-04-06
[fb-package] Automate weekly and monthly contingency table runs

nmdefries:
### Description
Add `make` commands to support automated contingency table generation and upload to the Delphi AWPS website.

Weekly cron job would run: `make weekly-contingency-pipeline`. Monthly cron job would run: `make monthly-contingency-pipeline`.

### Changelog
- Add contingency-specific SFTP post command (currently goes to the Delphi AWPS website which will either need manual publishing or for publishing to be automatically on -- perhaps this can be done for a single "automation" user), receiving directory, and destination directory
- Add contingency receiving dir to tidying and cleaning commands.
- Add weekly and monthly contigency params generation.
- Add weekly and monthly contigency pipelines.
- Add weekly and monthly contigency SFTP posting.
- Build function into package to fetch every fourth day of data when params file lists no input files. Creates list of input files for recordkeeping.


nmdefries:
@krivard I updated the Makefile as we discussed, with one exception. Instead of creating a command line-callable function to print every fourth date of input data, I put that bit of logic directly into the contingency pipeline. So if the list of input files is empty or not provided in the params file, the function will pull a list of appropriate input file names from `input_dir`. (Will plan to update to use the list of dormant surveys as exclusion patterns.)

facebook/Makefile
```diff
@@ -5,9 +5,11 @@ DRY:=yes
 ifeq ($(DRY),yes)
 	OUTSIDE_EMAIL_FLAGS:=-d
 	SFTP_POST:=echo -e "Would run: sftp -b <(echo -e \"\$${BATCH}\") -P 2222 fb-automation@ftp.delphi.cmu.edu\n$${BATCH}"
+	CONTINGENCY_SFTP_POST:=echo -e "Would run: sftp -b <(echo -e \"\$${BATCH}\") ???@publishing.andrew.cmu.edu\n$${BATCH}"
```
nmdefries, line 8: Will need to set up a daemon account for the AWPS website, preferably with exclusive auto-publishing capabilities.

krivard: I have an open ticket with facilities on this; they don't currently support it but they're discussing it today and will let us know. we may need to fall back to a delphi-operated hosting solution.

krivard:
> does not produce params files that are particularly useful for a historical log

That's fine -- you can have it output the list of files to a text file, and store it in the archive during the tidy step.

krivard:
Re:duplication, see https://github.com/cmu-delphi/covidcast-indicators/pull/1026

facebook/Makefile
krivard: Update: They've started working on a tentative solution. We will however need to create a dedicated AWPS collection to host them, since there's no way to set up autopublish on just part of a collection.

nmdefries: CCing @capnrefsmmat

krivard: delphi team leads just approved turning on autopublish for delphi-web and just making sure everyone with push access knows

krivard:
This is waiting on passwordless auth from IT

nmdefries:
> This is waiting on passwordless auth from IT

Is this still pending?

krivard:
Yep

nmdefries:
Survey has ended. This was never used.

PR #989, opened 2021-04-07
Add NAN code support to CAN

dshemetov:
### Description
Work as part #838. Branched off #1252, so that one should be merged first.

### Changelog
- Add three missingness columns, default to NA for "sample_size" and "stderr" columns, depending on the signal

### Fixes 
- Partially addresses #838


krivard:
Nothing risky until after we're clear of the JHU patch next week, which probably means after Thanksgiving 😞

PR #996, opened 2021-04-09
Add C17a to microdata output list

### LINKED ISSUE #987: Item C17a missing from survey microdata files
**Actual Behavior:**

It seems that in wave 8, we changed C17 to C17a (wording change) but did not add it to the output files. We should fix this. Nobody seems to be using it (it's the question about flu vaccination), but eventually we should regenerate the affected microdata files.

**Expected behavior**

C17a included.

nmdefries:
### Description
C17a, a revision of C17, was introduced in Wave 8 but not added to the microdata output list.

### Fixes
Closes #987.

@krivard We'll need to reissue the microdata going back to Feb 8 so we can include this item.

### DIFF
--- a/facebook/delphiFacebook/R/responses.R
+++ b/facebook/delphiFacebook/R/responses.R
@@ -455,7 +455,7 @@ create_complete_responses <- function(input_data, county_crosswalk)
     "C16", "C17", "E1_1", "E1_2", "E1_3", "E1_4", "E2_1", "E2_2", "E3", # added in Wave 5
     "V1", "V2", "V3", "V4_1", "V4_2", "V4_3", "V4_4", "V4_5", # added in Wave 6
     "V9", # added in Wave 7,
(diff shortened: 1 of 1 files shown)


capnrefsmmat (approved):
With Katie going on PTO, will there be a good way of reissuing in the mean time?

nmdefries:
I don't think so... I am able to produce the backfill files locally and can presumably add them to the SFTP site. But since the automatically-run code needs to be updated manually as far as I'm aware, it will continue making the old version of the microdata.

capnrefsmmat:
@nmdefries Reminder that this will require a reissue of microdata; can you coordinate with @krivard?

krivard:
Pulled, built, and installed; first issue of microdata with this code will go out tomorrow morning. @nmdefries you're clear to upload backfilled files through today at will.

PR #1000, opened 2021-04-10
Add nancodes enum util

dshemetov:
### Description
This is part of #838. Adds the nancode constants used throughout the indicators.

### Changelog
Add the nancodes.py of constants.

### Fixes 
- Fixes #838

PR #1001, opened 2021-04-12
Add mock to setup.py

chinandrew:
### Description
Builds are failing right now since the `import mock` statements are failing with Module not found. I don't know why it's only GS and utils either, since other packages use the same import

### Changelog
- Add `mock` to the setup.py. Decided to do this isntead of replacing `import mock` with `from unittest import mock` since the latter changes the imported package, which im not sure is desired. we can refactor later.

### Fixes 
- Fixes #(issue)

### DIFF
--- a/_delphi_utils_python/setup.py
+++ b/_delphi_utils_python/setup.py
@@ -6,6 +6,7 @@ required = [
     "covidcast",
     "freezegun",
     "gitpython",
+    "mock",
     "moto",
     "numpy",
     "pandas>=1.1.0",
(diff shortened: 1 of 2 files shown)


nmdefries (approved):
This seems fine -- tests pass, and I'm not getting any ModuleNotFound errors.

However, pydocstyle is flagging a bunch of problems, which seems to be causing build failures as well, [for example](https://github.com/cmu-delphi/covidcast-indicators/pull/1000/checks?check_run_id=2310335101).

chinandrew:
> This seems fine -- tests pass, and I'm not getting any ModuleNotFound errors.
> 
> However, pydocstyle is flagging a bunch of problems, which seems to be causing build failures as well, [for example](https://github.com/cmu-delphi/covidcast-indicators/pull/1000/checks?check_run_id=2310335101).

i think that's a separate issue for that PR

PR #1002, opened 2021-04-13
Add new version to sg-patterns

chinandrew:
### Description
Safegraph changed their s3 directory structure, so updating the code to handle it.

### Changelog
- Add new version with new s3 path
- Add new brand info file. Using yyyymmdd instead of previous yyyymm since there are two april versions.

### Fixes 
- Fixes #(issue)


jingjtang (approved):
lgtm

PR #1006, opened 2021-04-13
Add NAN code support to HHS Facilities

dshemetov:
### Description
Work as part #838. 

### Changelog
- Now codes the value "-999,999", which stands for the range [0-4], as 2 instead of nan

### Fixes 
- Partially addresses #838


jingjtang (approved):
lgtm

PR #1007, opened 2021-04-13
[DRAFT] Refactor indicators to use delphi utils pypi install

chinandrew:
### Description
Change indicators to use the pypi utils package instead of the local package. 

This change is still under discussion, hence the draft status.

### Changelog
- Replace local utils install with the util in the setup requirements

### Fixes 
- Fixes #(issue)


krivard:
Given #1009 I think we want to continue to install from local instead of pypi just to avoid github vs jenkins race conditions

PR #1012, opened 2021-04-14
Update changehc to use export util

dshemetov:
### Description
Part of #497. I can't quite tell if I achieved my goal of making the code simpler or not; this ended up being more complicated than I thought it would be.

### Changelog
- form the data into a dataframe and use the export util to write data instead of a local function
- updated tests
- update export util with a flag to allow writing empty CSVs for missing dates


krivard:
@rumackaaron ping

rumackaaron (approved):
I think this is much simpler, thanks! The code looks good to me

PR #1017, opened 2021-04-21
[fb-package] Create tool to find changes in survey waves using .qsf files

nmdefries:
### Description
This makes a command-line tool to diff two .qsf files, printing out questions that differ between the two with respect to item wording in answer choices and subquestions (for matrix items), and display logic, and questions that have been added to/removed from the survey.

Example output comparing Wave 8 and Wave 10:
```
 Added: item C13b (QID158)
 Added: item C13c (QID207)
 Added: item V11 (QID195)
 Added: item V12 (QID217)
 Added: item V13 (QID210)
 Added: item V14 (QID212)

 Removed: item A3b (QID8)
 Removed: item B2b (QID11)
 Removed: item C11 (QID18)
 Removed: item C12 (QID19)
 Removed: item C13 (QID57)
 Removed: item C13a (QID58)
 Removed: item Q93 (QID145)

 Question wording changed: item B10a (QID154)
 Question wording changed: item B10b (QID156)
 Question wording changed: item B2 (QID151)
 Question wording changed: item C6 (QID163)
 Question wording changed: item C8 (QID165)
 Question wording changed: item D7 (QID171)
 Question wording changed: item V2a (QID179)
 Question wording changed: item V5a (QID184)
 Question wording changed: item V5b (QID186)
 Question wording changed: item V5c (QID188)
 Question wording changed: item V5d (QID190)

 Display logic changed: item B11 (QID54)
 Display logic changed: item B2c (QID215)
 Display logic changed: item B7 (QID49)
 Display logic changed: item E1 (QID86)
 Display logic changed: item V5d (QID190)
 Display logic changed: item V6 (QID193)
 Display logic changed: item V9 (QID104)

 Answer choices changed: item A5 (QID148)
 Answer choices changed: item B10b (QID156)
 Answer choices changed: item B12a (QID56)
 Answer choices changed: item B7 (QID49)
 Answer choices changed: item Q67 (QID173)
 Answer choices changed: item Q78 (QID175)
 Answer choices changed: item V2a (QID179)

 Matrix subquestions changed: item A1 (QID146)
```

This provides the change type, item name, and QID for all changed questions.

The display logic change detection may be oversensitive. I've set it up to ignore changes in conditional question wording, but in some cases (e.g. for E1 above) it appears that wording stays the same despite the QID of the conditioning question changing, which triggers a "change" notice anyway. Any changes in comparison operator (<, >, =), value, comparison type, etc will also trigger a change notice.

### DIFF
--- a/facebook/translation-differ.R
+++ b/facebook/translation-differ.R
@@ -75,7 +75,13 @@ get_qsf_file <- function(path) {
     
     # These "questions" are not shown to respondents and thus don't need to be
     # accounted for in documentation or code. Skip.
-    if (question$QuestionText == "Click to write the question text") {next}
+    if (question$QuestionText == "Click to write the question text") {
+      warning(paste0("Item ", question$DataExportTag, " (", question$QuestionID,
+                     ") has question text '", question$QuestionText,
+                     "' and appears to be a remnant of the survey creation process; please remove."
+      ))
+      next
+    }
     
     # For matrix questions, "Choices" are the subquestion text and code. For all
     # other questions, "Choices" are the answer choices and corresponding
--- a/facebook/translation-differ.R
+++ b/facebook/translation-differ.R
@@ -116,7 +116,9 @@ diff_questions <- function(old_qsf, new_qsf) {
 }
 
 ## Compare a single question field.
-diff_question <- function(names, change_type, comparator, old_qsf, new_qsf) {
+diff_question <- function(names, change_type=c("answers", "wording", "display logic", "subquestions"), comparator, old_qsf, new_qsf) {
+  change_type <- match.arg(change_type)
+  
   changed <- c()
   for (question in names) {
     if ( !identical(old_qsf[[question]][[comparator]], new_qsf[[question]][[comparator]]) ) {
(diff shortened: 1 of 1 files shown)


capnrefsmmat:
Are you sure that diff is between Wave 8 and Wave 10? Wave 10 only introduced V11, V12, V13, and V14.

Is it possible the other "Added item" entries are actually changed items, but they have a new QID because we implemented the change by replacing them with a new question with the same variable name?

nmdefries:
Good catch! I used a Wave 8 .qsf file from the FB CSS drive, so maybe it was a partial file somehow. I'll rerun with a full version from Wichada.

capnrefsmmat:
Oh yeah, it's possible their version may only have changed questions or something, since they needed to translate those specific items.

nmdefries:
@capnrefsmmat This is ready for you to take a look at. I updated the example output above. It mostly matches the official changelog. There are some differences where we changed the wording of a question, but forgot to change the name in Qualtrics and instead made that change in code (e.g. C6 -> C6a). As mentioned above, the biggest difference is in official vs detected changes in display logic.

@WichadaLMK Please feel free to leave feedback on the format and contents of the example output. I can add question text or other question info in if you'd find that helpful.

WichadaLMK:
The output looks great to me. But I should probably admit now that I have no idea where this output is from. HA!

As far as the C6 change, this is something we caught after launch, so Nat saved me and changed to code to name it C6a to make it right for the CSVs. We will have to undo this for Wave 11.

nmdefries:
@WichadaLMK This output is created by running `Rscript translation-differ.R path/to/old/qsf path/to/new/qsf` at the command line. Output can be saved to a file in the standard bash way, `Rscript ... > path/to/output/file.txt`.

The script reads in the current survey data and the new survey data from the .qsf files. It figures out what survey items were displayed in each survey (items in any block that's not "Trash"), and compares their existence and content between the survey versions.

The script has some [info on .qsf structure](https://github.com/cmu-delphi/covidcast-indicators/blob/fb-package-translation-diff/facebook/translation-differ.R#L13) at the top. But essentially, for each question, we get raw text, display logic, answer choices, etc, that we compare for equality between the two surveys. If a given question has *any* change in wording, e.g., we print it out. If a question differs in multiple ways (e.g. question wording and answer choice wording), it will be printed out multiple times.

These comparisons are fairly sensitive. For example, the raw text for a question includes HTML formatting so if formatting/emphasis changes it will trigger a "change" notice. `Have you <strong>ever</strong> ...` vs `Have you ever ...` is considered changed.

capnrefsmmat:
I don't think I'll be able to code-review this thoroughly soon, but we shouldn't let that stop us from using it for this wave and using the results to improve things.

facebook/translation-differ.R
```diff
@@ -0,0 +1,240 @@
...
+#' @param path path to Qualtrics translation file in .qsf format
+#'
+#' @return A named list
+get_qsf_file <- function(path) {
+  # Read file as json.
+  qsf <- read_json(path)
+  
+  ## Block
+  block_filter <- sapply(qsf$SurveyElements, function(elem) { elem[["Element"]] == "BL" })
+  block_out <- qsf$SurveyElements[block_filter][[1]]$Payload
```
capnrefsmmat, line 71: Seems like a use case for `Filter()`?

capnrefsmmat (approved):
This looks great. I think you could use `Filter()` in a couple spots, instead of doing it in two steps with `sapply` and indexing, but otherwise this is nice and well-documented.

facebook/translation-differ.R
nmdefries: The object is a list of nested lists, not a dataframe, so 'filter' isn't applicable here. I was originally trying to manipulate the qsfs as dataframes, but changes in fields between questions/waves means it is sometimes hard to convert to dataframe, so I'm handling the qsfs exclusively as lists.

capnrefsmmat: Right, but I meant capital-F `Filter()`, not `filter()`. But maybe `Filter()` is for vectors and not for lists, I'm not sure

PR #1019, opened 2021-04-23
Add export_day_range into params.json

jingjtang:
### Description
EXPORT DAY RANGE is a variable stored in `constants.py` previously to control the number of dates for the updates everyday. We used to let it equal to 40 to have daily reports from today-45days to today-5days. 

Since we need to backfill data for dates ranging from 11-15-2020 to 12-09-2020, it's more convenient to have EXPORT DAY RANGE to be a variable that can be easily changed via `params.json`.

### Changelog
- remove `EXPORT_DAY_RANGE` from `constants.py` 
- read `export_day_range` from `params.json` in `run.py`
- minor updates in `tests/test_run.py`


chinandrew (approved):
:shipit::shipit::shipit:

PR #1025, opened 2021-04-24
[fb-package] Remove contingency aggregations with no or censored output

nmdefries:
### Description
Drop aggregates from being reported if they have no output to prevent tables including whole indicators as `NA` The main use of this is to not display indicators that are based on discontinued survey items.


capnrefsmmat (approved):
It's tempting to make you change this to use `Filter()`, but it's fine as-is

nmdefries:
Switching everything to `filter()` is still on my long-term nice-to-have list :laughing:

PR #1027, opened 2021-04-27
Remove the update step for the available signal matchlist from the Jenkins workflow

korlaxxalrok:
### Description
We are decommissioning api-staging.delphi.cmu.edu. There is a step in the
indicators Jenkinsfile that runs some Ansible to update a text file used by this
server. To keep Jenkins from failing we will remove this step.


krivard (approved):
👍

PR #1028, opened 2021-04-27
Remove google_health from indicator list to stop Jenkins from trying to build it

korlaxxalrok:
### Description
google_health is deprecated and removed from the repo. Jenkins was failing because
it was trying to build against a missing directory.

### Changelog
Removed `google_health` entry from `indicator_list`.


krivard (approved):
Nice catch

PR #1039, opened 2021-05-04
[fb-package] Create script to combine contingency tables by group over all dates

nmdefries:
### Description
Create script to combine into a single file the contingency tables that use the same set of grouping variables but over different time periods. These rollup files will be posted on the AWPS website so users can more easily browse through the files and, download all dates for a given grouping.

The user calls the script with the path to the input files and the path to the output files. Input files are grouped according to what output file they will go in (based on grouping variables used + geo level + aggregation time range). Generally this will be used to add a set of weekly or monthly tables to the output files, so most of the time we expect a one-to-one mapping of input and output names.

If files have not been seen before (based on list kept in `seen.txt`), they are directly appended to the pre-existing output file so that we don't have to read in the potentially large output data every time. Then the newly-added filenames are added to the already-seen list. Again, we expect this to be the case the majority of the time. If files *have* been seen before, the output data is read in so that we can deduplicate it. For each unique group, we keep the row with the most recent issue date.

Output names have a similar format to the individual tables (e.g. `20210418_20210424_weekly_nation_age_edulevel.csv`), but drop the initial dates and are compressed using gzip (`....csv.gz`).

### Changelog
- Create `contingency-combine.R` script as a command-line tool.

### Fixes 
[Asana task](https://app.asana.com/0/1199704294865989/1200043573046771/f)

### DIFF
--- a/facebook/contingency-combine.R
+++ b/facebook/contingency-combine.R
@@ -1,0 +1,144 @@
+#!/usr/bin/env Rscript
+
+## Combine and compress contingency tables by aggregation.
+##
+## Usage:
+##
+## Rscript contingency-combine.R path/to/individual/files/ path/to/rollup/files/
+##
+## Appends a set of newly-generated contingency tables to a rollup CSV that
+## contains all dates for a given set of grouping variables. Can also be used to
+## combine a directory of tables spanning multiple time periods.
+
+suppressPackageStartupMessages({
+  library(dplyr)
+  library(readr)
+  library(purrr)
+})
+
+
+#' Fetch all tables in a chosen directory and combine according to grouping
+#' used.
+#'
+#' @param input_dir Directory in which to look for survey CSV files, relative to
+#'   the current working directory.
+#' @param output_dir Directory in which to look for existing rollup files or
+#'   create new ones, relative to the current working directory.
+#' @param pattern Regular expression indicating which files in that directory to
+#'   open. By default, selects all `.csv` files with standard table date prefix.
+run_rollup <- function(input_dir, output_dir, pattern = "^[0-9]{8}_[0-9]{8}.*[.]csv$") {
+  files <- list.files(input_dir, pattern = pattern)
+
+  if (length(files) == 0) {
+    stop("No matching data files.")
+  }
+
+  files <- map_dfr(files, get_file_properties)
+  
+  # Reformat files as a list such that input files with same grouping variables
+  # (and thus same output file) are in a character vector named with the output
+  # file.
+  files <- lapply(split(files, files$rollupname), function(x) {x$filename})
+  
+  for (output_name in names(files)) {
+    combine_and_save_tables(
+      file.path(input_dir, files[[output_name]]), 
+      file.path(output_dir, output_name))
+  }
+  
+  return(NULL)
+}
+
+## Helper function to extract info from each file's filename.
+get_file_properties <- function(filename) {
+  short <- strsplit(filename, ".", fixed = TRUE)[[1]][1]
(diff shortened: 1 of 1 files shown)


facebook/contingency-combine.R
```diff
@@ -0,0 +1,182 @@
...
+##
+## Appends a set of newly-generated contingency tables to a rollup CSV that
+## contains all dates for a given set of grouping variables. Can also be used to
+## combine a directory of tables spanning multiple time periods.
+
+suppressPackageStartupMessages({
+  library(dplyr)
+  library(readr)
+  library(purrr)
+  library(delphiFacebook)
```
capnrefsmmat, line 17: I didn't see delphiFacebook get used anywhere in here, but maybe I missed it

capnrefsmmat: Oh, is it just `assert`?

facebook/contingency-combine.R
```diff
@@ -0,0 +1,182 @@
...
+  } else {
+    assert(file.exists(output_file),
+           paste0("The output file ", output_file, " does not exist, but non-zero",
+                  " files using the same grouping variables have been seen before."))
+    
+    output_df <- read_csv(output_file, col_types = cols)
+    
+    # Use all columns up to the first "val" column to find unique rows.
+    group_names <- names(output_df)
+    ind_first_val_col <- min(which(startsWith(group_names, "val_")))
```
capnrefsmmat, line 147: Is it clearly documented in the contingency table generator that the grouping columns must always precede the `val_` columns, and that, say, the standard error can't come before the `val_`, to prevent this from messing up?

capnrefsmmat (approved):
This looks good to me overall; my comments aren't blockers. I am a bit nervous, though, since the merging logic is a bit tricky and it's hard to test because it does the IO directly in `combine_and_save_tables`. I wonder if it's worth splitting that into `combine_tables` and `save_tables` so that `combine_tables` just does the data frame merge step, and then you can easily test that step to ensure it always combines correctly.

facebook/contingency-combine.R
nmdefries: Yep, just `assert`. I'm surprised R doesn't have a built-in version, it seems pretty basic/essential.

capnrefsmmat: The closest thing is `stopifnot`, which doesn't allow a message to be provided. I do like [`assertthat`](https://cran.r-project.org/package=assertthat) though.

facebook/contingency-combine.R
nmdefries: The `write` function explicitly reorders the columns in a table so that the grouping columns and table metadata (date range, US vs global survey, etc) always come first. I've added more info to the function description to make that clear.

> the standard error can't come before the val_

Good point. I can make the column-finder bit more general so that it only uses up to the first column that's not `val`, `se`, `sample_size`, etc.

nmdefries:
@capnrefsmmat Could you take another look at this?

The biggest change is the removal of the functionality to directly append to the output file if an input table has not been seen before. It complicated the logic a lot and I realized it's not going to be used that much. Even if an input table hasn't been seen before, it can still have a column mismatch with existing output, for example, if a new indicator was added or if an old indicator was discontinued. So most of the time we still want to read in the output so the input-output merge is done correctly (by name rather than position).

Because of this change, the `seen.txt` file that tracks tables we've already input isn't strictly necessary, but I'm thinking of keeping it for now so there's a record of which tables have been processed. Removing the seen-file would make the code shorter/simpler, though. Thoughts?

capnrefsmmat (approved):
Looks good. Nice commit messages too

PR #1044, opened 2021-05-07
Rename nancodes before release

dshemetov:
### Description
After some discussion, we have decided to rename PRIVACY to CENSORED and UNKNOWN to OTHER. Since these codes have not been used in the system yet, no other changes will be necessary.


krivard (approved):
👍

Issue #970, opened 2021-03-29
Add ability to tune validation outlier thresholds


sgsmob:
One validation check is to compare the actual signal value and a "predicted" signal value from the recent historical API data.  If these values are too far apart, we classify the signal as anomalous and trigger a validation error.  Currently this is done using a set of [hardcoded thresholds](https://github.com/cmu-delphi/covidcast-indicators/blob/67a10e08ec81f0005497a857570128b0975e26f7/_delphi_utils_python/delphi_utils/validator/dynamic.py#L525), which gives the user two options:  use these thresholds or turn off this type of validation check entirely.  It would nice if we had a method of adjusting these thresholds from the `params.json` file to allow indicators with more variance to still have this type of validation.

sgsmob:
It is not obvious how we should do this, but here are some ideas.
1.  Expose the thresholds to the user
    1. (+) Most customizable
    2. (-) Hard for the user to meaningfully set since they have no inherent interpretation (?)
2.  Expose a multiplicative factor to apply to the thresholds
    1.(+) Only need single parameter
    2. (-) Potentially even less interpretable than option 1
3.  Expose discreet sensitivity levels to the user ("high", "medium", "low")
    1. (+) Obvious interpretation
    2. (-) How to we set the levels meaningfully?
    3. (-) Extending to higher granularity could be messy


@nmdefries any additional thoughts?

nmdefries:
Options 3 sounds the most reasonable. If I recall correctly, the metric that is being used (relative mean difference) asymptotes to 2, so a user setting a multiplicative factor or a threshold directly could easily create a threshold that never triggers an error.

> Hard for the user to meaningfully set since they have no inherent interpretation (?)

Strong agree here. The threshold values were never set to mean anything in particular, they were eyeballed to keep the false positive rate reasonable (only based on the Facebook survey pipeline, I believe).

Relative mean difference is a measure of dispersion so maybe thresholds can be mapped to percentiles (and back)? But if we're open to changing how this check works, it'd be a lot more interpretable if we used a percentile threshold directly.

nmdefries:
It appears that the overzealous errors are because the existing metric doesn't take variance into account. This approach works fine for relatively smooth signals, but in regions with small populations (e.g. small counties) and rare signals (e.g. deaths) variance is high causing many errors to be raised.

[...]

PR #1051, opened 2021-05-10
Use z-score to account for variance in validation check of current vs historical values

nmdefries:
### Description
The validation check `check_test_vs_reference_avg_changed` previously used a metric tracking average change (across all regions and dates-to-be-checked) in historical average value vs average value in test set compared to historical average + current average. Think half-way between multiplicative increase in current average value compared to historical average value and relative mean difference. This approach does not account for variance, causing errors in high-variance signals, which we tend to see at fine geo levels (county, e.g.) due to small population.

To account for variance and to allow more intuitive tuning of hard-coded thresholds, this switches `check_test_vs_reference_avg_changed` to using z-score instead.

Initial z-score thresholds were set using JHU signals only in consideration of #752.

### Changelog
- Use z-score in place of "relative mean difference". Variances of 0 are replaced with the smallest non-zero variance.
- Increase size of reference window used for smoothed signals to avoid interaction with moving average smoothing window.
- Auto-formatting.

### Fixes 
This PR makes #970 obsolete by switching to a metric that controls for historical variance. There should be no need for users to adjust metric thresholds on a per-signal basis.

### DIFF
--- a/_delphi_utils_python/delphi_utils/validator/dynamic.py
+++ b/_delphi_utils_python/delphi_utils/validator/dynamic.py
@@ -1,5 +1,4 @@
 """Dynamic file checks."""
-import math
 from dataclasses import dataclass
 from datetime import date, timedelta
 from typing import Dict, Set
--- a/_delphi_utils_python/delphi_utils/validator/dynamic.py
+++ b/_delphi_utils_python/delphi_utils/validator/dynamic.py
@@ -409,8 +409,8 @@ class DynamicValidator:
         # These variables are interpolated into the call to `api_df_or_error.query()`
         # below but pylint doesn't recognize that.
         # pylint: disable=unused-variable
-        api_frames_end = min(api_frames["time_value"].max(
-        ), source_frame_start-timedelta(days=1))
+        api_frames_end = min(api_frames["time_value"].max(),
+                             source_frame_start-timedelta(days=1))
         # pylint: enable=unused-variable
         outlier_df = all_frames.query(
             'time_value >= @api_frames_end & time_value <= @source_frame_end')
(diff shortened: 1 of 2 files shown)


_delphi_utils_python/delphi_utils/validator/dynamic.py
```diff
@@ -461,89 +473,84 @@ def check_avg_val_vs_reference(self, df_to_test, df_to_reference, checking_date,
...
-
-        df_all = pd.concat([df_to_test, df_to_reference])
+        # Calculate reference mean and standard deviation for each geo_id.
+        reference_mean = df_to_reference.groupby(['geo_id'], as_index=False)[
+            ['val', 'se', 'sample_size']].mean().assign(type="reference mean")
+        reference_sd = df_to_reference.groupby(['geo_id'], as_index=False)[
+            ['val', 'se', 'sample_size']].std().assign(type="reference sd")
+
+        # Replace standard deviations of 0 with non-zero min sd for that type. Ignores NA.
+        replacements = {"val": {0: df_to_test.val[df_to_test.val > 0].min()},
```
rumackaaron, line 483: Why do we need to replace mean-0 signals with a mean greater than 0?

nmdefries: Whoops, the replacement values should've been taken from `reference_sd`. Thanks for the catch!

This is intended to replace the standard deviation of a zero-variance signal with the smallest non-zero standard deviation seen in another geo region of the same type (e.g. county) for the same signal (e.g. cases), since otherwise *any* change would trigger an error (z-score = inf with sd = 0).

nmdefries: The other approach would be to just ignore/remove the zero-variance geo regions. Not sure if it makes a practical difference one way or the other. About 2% of counties have periods of zero-variance, based on a single spot-check.

rumackaaron: My question is specifically about the mean replacement. I don't know of any signals that could have a negative mean, but presumably it could be possible and the z-score can still be calculated and interpretable for negative means.

nmdefries: Hm, I'm confused. Maybe there's a bug I'm not seeing, but this shouldn't be replacing mean values, only standard deviations.

[This bit above calculates standard deviations](https://github.com/cmu-delphi/covidcast-indicators/blob/24da6c0eec94fae80279dccca1bc3afc86628c59/_delphi_utils_python/delphi_utils/validator/dynamic.py#L479) for each of the `val`, `se`, and `sample_size` columns. So in the `reference_sd` dataframe, we end up with a column called `val` that is the standard deviation of the original `val` column (and `se` for sd of `se`, etc), it's not the mean (means are stored separately in `reference_mean`). The replacements defined here are [*only* being used on the `reference_sd` dataframe](https://github.com/cmu-delphi/covidcast-indicators/blob/24da6c0eec94fae80279dccca1bc3afc86628c59/_delphi_utils_python/delphi_utils/validator/dynamic.py#L486).

rumackaaron: Ah I understand now, I'm glad my confusion allowed you to catch a bug

rumackaaron (approved):

PR #1053, opened 2021-05-12
Turn on async covidcast fetch for sirCAL

krivard:
### Description

The nightly SirCAL run has been pushing an hour in running time, so it's probably a good idea to switch to async calls.

### Changelog
- check_source.py


chinandrew:
Any concerns with hitting the 502 errors that combo was seeing?

krivard:
We haven't had one of those in combo since April 2, so I think we should be fine

chinandrew (approved):
:shipit: :shipit: :shipit:

PR #1054, opened 2021-05-12
[fb-package] Support Wave 11

nmdefries:
### Description
Update microdata, API, and contingency table pipelines to support new and changed items in Wave 11.

[Wave 10 vs 11 diff as of 05/17](https://github.com/cmu-delphi/covidcast-indicators/files/6494618/Wave10_vs_11_diff_after_surveychecks.txt)

### Changes
In API,
- `screening_tested_positive_14d`: "large outdoor event" and "crowded indoor environment" options removed; "for travel" option was added as a new screening reason
- `anxious_7d`: C8a discontinued, now defined using C18a
- `depressed_7d`: C8a discontinued, now defined using C18b
- `tested_positive_14d`: B10a removed, now defined using B10c (same meaning and logic, minor change in wording)
- `tested_14d`: no longer uses B8 ("ever tested";  removed) to fill in B10 for better coverage (if respondent never tested, also not tested in last 14d). Kept same name.
- `smoothed_wothers_masked_public` replaces `smoothed_wothers_masked`: H2 replaces C16 with change in wording
- set of `trust_covid_info_<source>` replaces set of `vaccine_likely_<source>`: I6 replaces V4a with change in meaning, wording, and info sources being asked about.
- `covid_vaccinated_appointment_or_accept` replaces `covid_vaccinated_or_accept`: V3a replaces V3 with change in logic (V3a only asked of respondents without vaccine appointment)
- `appointment_or_accept_covid_vaccine` replaces `accept_covid_vaccine`: same
- `accept_covid_vaccine_no_appointment` replaces `accept_covid_vaccine`: same. The change means that acceptance based on V3a only is among respondents who have not been vaccinated and do not have an appointment to do so, versus V3 being among respondents who have not been vaccinated but may or may not have an appointment
- Handful of signals discontinued
- Similar changes in contingency tables; some name changes differ slightly


facebook/delphiFacebook/R/variables.R
```diff
@@ -373,6 +415,15 @@ code_vaccines <- function(input_data) {
...
     )
+  } else if ("V3a" %in% names(input_data)) {
+    input_data$v_covid_vaccinated_or_accept <- case_when(
+      input_data$V1 == 1 ~ 1,
+      input_data$V3a == 1 ~ 1,
+      input_data$V3a == 2 ~ 1,
+      input_data$V3a == 3 ~ 0,
+      input_data$V3a == 4 ~ 0,
+      TRUE ~ NA_real_
+    )
```
capnrefsmmat, line 426: A key question: What do we do with `v_covid_vaccinated_or_accept` now that V3a has different display logic?

Specifically, V11a is now before V3a, and asks respondents if they have an appointment. They are only asked V3a if they do not.

I think this means we need a "vaccinated, appointment, or accept" signal, and the question is just whether this signal keeps the `covid_vaccinated_or_accept` name or gets a new name to indicate that there may be a discontinuity.

facebook/delphiFacebook/R/responses.R
```diff
@@ -457,6 +459,10 @@ create_complete_responses <- function(input_data, county_crosswalk)
     "V9", # added in Wave 7,
     "C14a", "C17a", "V2a", "V5a", "V5b", "V5c", "V5d", "V6", "D11", # added in Wave 8
     "C6a", "C8a_1", "C8a_2", "C8a_3", "C13b", "C13c", "V11", "V12", "V13", "V14_1", "V14_2", # added in Wave 10
+    "B10c", "B13 ", "C18a", "C18b", "C7a", "D12", "E4",
```
capnrefsmmat, line 462:
```suggestion
    "B10c", "B13", "C18a", "C18b", "C7a", "D12", "E4",
```

capnrefsmmat: Is there a good way to compare this list against the QSF diff to verify that nothing is missing?

facebook/delphiFacebook/R/variables.R
```diff
@@ -238,6 +251,18 @@ code_mask_contact <- function(input_data) {
...
     input_data$c_others_masked <- NA
   }
+  
+  if ("H2" %in% names(input_data)) {
+    # added in wave 11, replaces C16. most/all *others* seen in public wearing
+    # masks; exclude respondents who have not been in public.
+    input_data$c_others_masked_public <- case_when(
+      is.na(input_data$H2) ~ NA,
+      input_data$H2 == 6 ~ NA,
+      input_data$H2 == 4 | input_data$H2 ==  5 ~ TRUE,
```
capnrefsmmat, line 261:
```suggestion
      input_data$H2 == 4 | input_data$H2 == 5 ~ TRUE,
```

facebook/delphiFacebook/R/variables.R
```diff
@@ -292,7 +328,8 @@ code_testing <- function(input_data) {
     input_data$t_tested_reason_crowd <- is_selected(testing_reasons, "6")
     input_data$t_tested_reason_visit_fam <- is_selected(testing_reasons, "7")
     input_data$t_tested_reason_other <- is_selected(testing_reasons, "8")
-    
+    input_data$t_tested_reason_travel <- is_selected(testing_reasons, "9")
```
capnrefsmmat, line 331: Some of these testing reasons were removed in Wave 11, and as written, the code will just show their estimates go to 0 after Wave 11 is deployed. Should we change the code so they're marked NA for Wave 11+, rather than `is_selected` just reporting that they're not selected?

capnrefsmmat (changes requested):
Some comments from before you marked it draft, noting a few things that we'll need to figure out

facebook/delphiFacebook/R/responses.R
nmdefries: This is taken directly from the diff (+ formatting changes, of course), so it should be exactly the same.

facebook/delphiFacebook/R/variables.R
nmdefries: Creating new vaccinated+appointment+accept, appointment+accept, and accept+no appointment indicators.

nmdefries:
Items that had slightly changed but I thought similar enough to leave as-is:

* tested_positive_14d: B10a replaced by B10c; same meaning and logic, different wording
* screening_tested_positive_14d: same as tested_positive_14d
* tested_14d: B8 was previously used to fill in B10 for better coverage (if respondent never tested, also not tested in last 14d)

capnrefsmmat (approved):
I believe this looks good. I will send you a sample CSV from Qualtrics so you can test the code runs before we deploy, so we don't do the usual thing of having the pipeline crash the first time...

nmdefries:

PR #1056, opened 2021-05-13
Add logging to CAN

### LINKED ISSUE #962: Add logging to CAN indicator
[Guidelines for adding logging are here](https://docs.google.com/document/d/1PFUbnZiRGsCoi3Pr0QnM-lC1bKKgAcGO0t-ZdtOXLxo/edit?usp=sharing), though they're slightly outdated so cross-check with one of the other indicators.

chinandrew:
### Description
CANs had no logging, so adding basic time log so it shows in elastic.

### Changelog
- Add total elapsed time logging to CAN
- Update template and ansible params files with log file.

### Fixes 
- Fixes #962


covid_act_now/delphi_covid_act_now/run.py
```diff
@@ -97,3 +104,6 @@ def run_module(params):
         # Report failures: someone should probably look at them
         for exported_file in fails:
             print(f"Failed to archive '{exported_file}'")
+
+    elapsed_time_in_seconds = round(time.time() - start_time, 2)
+    logger.info("Completed indicator run", elapsed_time_in_seconds=elapsed_time_in_seconds)
```
krivard, line 109: does it make sense to add any of the other summative stats? a la jhu:
https://github.com/cmu-delphi/covidcast-indicators/blob/0398be979c39e29cec2b719c96f0cbc91e7a7790/jhu/delphi_jhu/run.py#L171-L175

chinandrew: good idea, i can add min and max exported dates. Though there is a weird pattern where it exports twice (one for positivity rate and one for total tests), so perhaps we need to split them up. Do we have any guidelines on when we want to use multiple logging statements vs bundling them all into one?

krivard: The dates are the same for both exports, so you should be able to count them using something like
```
csv_export_count += 2*exported_csv_dates.size
```

chinandrew: added

covid_act_now/delphi_covid_act_now/run.py
```diff
@@ -83,6 +83,7 @@ def run_module(params):
             sensor=SIGNALS[1])
 
         earliest, latest = min(exported_csv_dates), max(exported_csv_dates)
+        files_exported = df.size * 2  # multiply by two to count both positivity and tests signals
```
krivard, line 86:
```suggestion
        files_exported = exported_csv_dates.size * 2  # multiply by two to count both positivity and tests signals
```

we want the number of files, not the number of lines

covid_act_now/delphi_covid_act_now/run.py
```diff
@@ -106,4 +107,8 @@ def run_module(params):
             print(f"Failed to archive '{exported_file}'")
 
     elapsed_time_in_seconds = round(time.time() - start_time, 2)
-    logger.info("Completed indicator run", elapsed_time_in_seconds=elapsed_time_in_seconds)
+    logger.info("Completed indicator run",
+                elapsed_time_in_seconds=elapsed_time_in_seconds,
+                csv_export_count=files_exported,
+                earliest_export_date=earliest.strftime("%Y-%m-%d"),
+                latest_export_date=latest.strftime("%Y-%m-%d"))
```
krivard, line 114: `earliest` and `latest` are local mini/maximums, you probably want a global one here

krivard (changes requested):

covid_act_now/delphi_covid_act_now/run.py
chinandrew: ah right i forget we have multiple geo values. should be the sum of all lines across all geos right? 

also i wonder whats going on with that comment formatting...

covid_act_now/delphi_covid_act_now/run.py
chinandrew: :+1:

covid_act_now/delphi_covid_act_now/run.py
```diff
@@ -76,6 +86,9 @@ def run_module(params):
             sensor=SIGNALS[1])
 
         earliest, latest = min(exported_csv_dates), max(exported_csv_dates)
+        min_dates_exported.append(earliest)
+        max_dates_exported.append(latest)
+        num_exported_files += df.size * 2  # x2 to count both positivity and tests signals
```
krivard, line 91: ?
```suggestion
        num_exported_files += exported_csv_dates.size * 2  # x2 to count both positivity and tests signals
```

krivard:
(lint)

krivard (approved):
🎉

PR #1057, opened 2021-05-18
[fb-survey] permit fetching qualtrics files for past days

krivard:
### Description
Currently the delphi_facebook python package fetches the most recent 7 days of data regardless of the date set in the params file. This PR makes it look at the `end_date` parameter and compute which 7 days to fetch from qualtrics.

### Changelog
- Makefile -- set `end_date` before fetching qualtrics data
- run.py -- set `qualtrics.date` from `end_date`
- qualtrics.py -- configure POST parameters using `qualtrics.date`

### DIFF
--- a/facebook/Makefile
+++ b/facebook/Makefile
@@ -81,6 +81,7 @@ $(TODAY):
 	done; \
 	$(SFTP_POST); \
 	echo "SUCCESS: $(DRY_MESSAGE)Posted `echo -e $${BATCH} | wc -l` raw files" >> $(MESSAGES)
+	touch -d $(YESTERDAY) params.json
 	mv tmp $@
 
 params.json: $(TODAY)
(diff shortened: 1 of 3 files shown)


facebook/Makefile
```diff
@@ -72,6 +72,7 @@ $(TODAY):
 	[ -f $(QUALTRICS) ] || mkdir -p $(QUALTRICS)
 	rm -rf tmp
 	touch -d "$(TODAY) 01:00" tmp
+	$(PYTHON) -m delphi_utils set end_date $(YESTERDAY)
```
nmdefries, line 75: Looks like this procedure already appears on line 91, in the `params.json` target. I'm not understanding why this is happening twice.

nmdefries: Nvm, got it.

nmdefries (approved):

PR #1060, opened 2021-05-20
Include random module assignment

nmdefries:
### Description
Process and save to microdata variable indicating if respondent was randomized to Module A or Module B.

### Changelog
- Create variable `module`, coded as A/B/`NA`, based on column `FL_23_DO`. Note that this column name may change in later survey waves.
- Add `module` variable to microdata output.
- Ask Qualtrics API for display order info.


capnrefsmmat (approved):
R code portion looks fine to me; I'll let @krivard look at the one line of Python

krivard (approved):
legit

krivard:
is this urgent enough for a friday deploy, or hold off for monday in case something goes sideways?

capnrefsmmat:
If we deploy Monday, we'll want to reissue the microdata files from this weekend (the first ones with Wave 11) so they have the module column added to them. Though I guess that will happen automatically, since the past few days get reissued.

krivard:
deployed, ready for tomorrow morning. @nmdefries would you reissue the microdata?

PR #1064, opened 2021-05-26
Adjust Waves 6 and 11 logic in contingency tables

nmdefries:
### Description
Previously, tables using January data were run starting from Jan 6 rather than the the first date in the relevant time range, Jan 1 for monthly tables and Jan 3 for weekly tables. Jan 6 is the date when the display logic of items V3 ("would you get the vaccine if offered today?") and V4 ("whose recommendation would make you more or less likely to get the vaccine?") changed to be shown only to respondents who indicated that they had not been vaccinated.

This code explicitly sets pre-Jan 6 values in V3 and V4 to `NA` so that we can produce historical tables without mixing pre-Jan 6 and post-Jan 6 display logic.

Set `hesitant_barrier_*` and `defno_barrier_*` items to missing for responses to Wave 11 and later. This fixes the logic of the #1061 wave patch.

### Changelog
- Set `b_accept_vaccine`, `mc_accept_cov_vaccine`, and all `b_vaccine_likely_*` values to `NA` where the survey was started prior to Jan 6 at midnight Pacific.
- Set `hesitant_barrier_*` and `defno_barrier_*` values to `NA` where the respondent took Wave 11 or later.

### DIFF
--- a/facebook/delphiFacebook/R/contingency_variables.R
+++ b/facebook/delphiFacebook/R/contingency_variables.R
@@ -358,7 +358,11 @@ remap_responses <- function(df) {
 #' @param df Data frame of individual response data.
 #'
 #' @return data frame of individual response data with newly derived columns
+#' 
+#' @importFrom lubridate ymd
 create_derivative_columns <- function(df) {
+  wave6_mod_date <- ymd("2021-01-06", tz=tz_to)
+    
   ###---
   # Grouping variables.
   ###---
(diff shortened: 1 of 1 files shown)


nmdefries:
@davidkretch Could you take a look at this?

facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -979,13 +1001,11 @@ create_derivative_columns <- function(df) {
...
     df$b_hesitant_barrier_religious <- as.numeric(is_selected(hesitancy_reasons, "15"))
     
-    if (wave >= 11) {
-      input_data$b_hesitant_barrier_allergic <- NA
-      input_data$b_hesitant_barrier_not_recommended <- NA
-      input_data$b_hesitant_barrier_distrust_vaccines <- NA
-      input_data$b_hesitant_barrier_health_condition <- NA
-      input_data$b_hesitant_barrier_pregnant <- NA
-    }
+    df$b_hesitant_barrier_allergic[df$wave >= 11] <- NA
```
davidkretch, line 1004: I feel like there should be a comment here stating that these must explicitly be set these to missing because `is_selected` will always return false if an option isn't selected but there is an answer to the question, even if the option is no longer available. It took me a little bit to understand.

davidkretch:
Looked good to me and worked in my tests. Had one suggestion for a clarifying comment.

nmdefries:
@capnrefsmmat Could you take a look at this and/or merge?

capnrefsmmat (approved):

PR #1067, opened 2021-06-01
Better support for patching using Makefile pipeline

krivard:
### Description
You can now set TODAY in environment variables, and the Makefile will set all other dates (YESTERDAY, etc) based on that value. This makes it easier to run the pipeline for missed days due to outages.

### Changelog
- Makefile


nmdefries (approved):
Looks good

PR #1070, opened 2021-06-08
[fb-package] Standardize variable creation

nmdefries:
### Description
Reformat contingency indicator creation function to match the style of the API in `variables.R`. A standard style improves maintainability, and defining indicators in `load_response_one` allows for easier single-wave processing.

These changes do not change table output, with the exception of the "smoker" grouping variable being recoded as True/False rather than 1/0 to match the standard boolean grouping variable format.

### Changelog
- Add arg to `load_responses_all` to indicate if data is being used for API (default) or contingency tables, in which case additional variables are created.
- Remove auto-recoding by indicator prefix (`b_`, `mc_`, etc) and for a select set of survey items (education, etc). This functionality was not being used and added complexity.
- Separate contingency indicator creation by topic into smaller functions in the style of `code_*` in `variables.R`.


nmdefries:
@davidkretch Could you verify that the output from the changed code here is the same as that from the prior version? It's the same for me (aside from "smoker" going from 1/0 -> True/False), but I'd appreciate a second pair of eyes.

davidkretch (approved):
Sorry for the delay! The only differences in outputs I see are also in the smoker files, where values 0 and 1 changed to FALSE and TRUE respectively, as expected. The code changes looked good to me as well.

nmdefries:
Thanks, @davidkretch!

capnrefsmmat (approved):
Looks fine to me. Ready to merge?

nmdefries:
Ready to merge!

PR #1074, opened 2021-06-21
Switch to sendmail for stale weights notifications

krivard:
### Description
Surveys is moving to a temporary home after three strikes on the server under my desk on campus crashing with no evidence as to why. The Makefile emails for help if the weights are stale, and need to be converted from `s-nail` to `sendmail`.

### Changelog
- Makefile - more detailed dry-run handling for sending email via sendmail instead of snail


nmdefries (approved):
Looks good

PR #1075, opened 2021-06-22
Add raw data logging statement to USAFacts

chinandrew:
### Description
USAFacts has had a few instances where it ran without error but nothing new was ingested. It's unclear if it's something in our pipeline or if it's from the source, but the latest instance shows that the max export date did not change for 3 days in a row, suggesting that raw data may not have updated. This PR adds additional logging on the raw data.

Another thing we might consider doing in the future is just storing all the raw data to s3 as an intermediate step.

### Changelog
- Add log statement with info about the raw dataframe downloaded.
    - dimensions
    - min/max dates
    - checksum

### Fixes 
- Fixes #(issue)

### DIFF
--- a/usafacts/delphi_usafacts/pull.py
+++ b/usafacts/delphi_usafacts/pull.py
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
 """Functions for pulling data from the USAFacts website."""
+import hashlib
+from logging import Logger
+
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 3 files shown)


usafacts/delphi_usafacts/pull.py
```diff
@@ -52,6 +55,13 @@ def pull_usafacts_data(base_url: str, metric: str) -> pd.DataFrame:
     """
     # Read data
     df = pd.read_csv(base_url.format(metric=metric))
+    date_cols = [i for i in df.columns if i.startswith("2")]
```
krivard, line 58: ??? why `if i.startswith("2")`

chinandrew: The county names are like 
```
countyFIPS            County Name State  StateFIPS  2020-01-22  2020-01-23  ...  2021-06-16  2021-06-17  2021-06-18  2021-06-19  2021-06-20  2021-06-21
```

So the dates start with 2. To be a bit more rigorous I could regex or try to date parse them, or just add a comment explaining.

krivard (approved):
👍

Issue #1073, opened 2021-06-14
HHS data should include population-normalized signals, perhaps 7-day averages


capnrefsmmat:
The [HHS source](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/hhs.html) includes absolute counts of hospital admissions, but no `_prop` signals or `smoothed_` signals. This makes it inconvenient when I want to plot hospital admissions against other signals, like survey signals or any other per-population quantity.

- [x] smoothed variants
- [x] population-proportionate variants

chinandrew:
Is this closed with #1199 ?

krivard:
Yep!

PR #1077, opened 2021-06-23
Add smoothing to hhs hosp

chinandrew:
### Description
Add 7dav smooth signal to hhs. Follows similar pattern to JHU indicator with the smoother as part of the list iteration. 

Using product() is slightly inefficient since it recomputes some of the dataframes extra times, but doing it like

```
for sensor in SIGNALS:
    df = geo_mapper.add_geocode(make_signal(all_columns, sensor),
                                "state_id",
                                "state_code",
                                from_col="state")
    for geo in GEOS:
        df = make_geo(df, geo, geo_mapper)
        for smoother in SMOOTHERS:
            df = smooth_values(df, smoother[0])
            sensor_name = smoother[1] + sensor
            create_export_csv(df, params["common"]["export_dir"], geo, sensor_name)
```

leads to some weird behavior since the dataframes are modified in place, opening up state issues within the inner loops. Case in point, the above code will throw an error, while the implementation in this PR (below) works. Was a real head scratcher trying to figure out what was going on.
```
for sensor, smoother, geo in product(SIGNALS, SMOOTHERS, GEOS):
    df = geo_mapper.add_geocode(make_signal(all_columns, sensor),
                                "state_id", 
                                "state_code",
                                from_col="state")
    df = make_geo(df, geo, geo_mapper)
    df = smooth_values(df, smoother[0])
    sensor_name = smoother[1] + sensor
    create_export_csv(df, params["common"]["export_dir"], geo, sensor_name)
```
Fixing it would require making copies at each step which is a bit ugly. 

The whole indicator runs in <2 min on my laptop so figured it should be ok for now to generate a few extra dfs. Main currently runs in 1 min.

### Changelog
- Add 7dav smoother signal, with signal names being "7dav_" prepended to existing signal names
- Add test verifying all files were generated
- Some old test style cleanup, plus one more test case for the date range function.

### Fixes 
- Partially addresses #1073. Prop signals not implemented.


chinandrew:
When trying to verify new output with old output, I'm seeing a couple one line differences in the occasional file. Looking into it now, and will switch this PR to draft mode so it doesnt accidentally get merged. I think it's something to do with the data types, since some of the epidata columns appear to be getting cast to `object` instead of `float` or `int`, maybe due to weird nan behavior. While that may explain some of the NA values becoming 0.0, there's a few where the value actually changes which is a bit worrying. I'll look into it more tomorrow, but worst case we just cast everything to float before doing any operations.

chinandrew:
> When trying to verify new output with old output, I'm seeing a couple one line differences in the occasional file. Looking into it now, and will switch this PR to draft mode so it doesnt accidentally get merged. I think it's something to do with the data types, since some of the epidata columns appear to be getting cast to `object` instead of `float` or `int`, maybe due to weird nan behavior. While that may explain some of the NA values becoming 0.0, there's a few where the value actually changes which is a bit worrying. I'll look into it more tomorrow, but worst case we just cast everything to float before doing any operations.

Resolved by removing smoothing imputation

krivard (approved):
👍

PR #1078, opened 2021-06-23
[Fix] Make Smoother "identity" smoother a true no-op by default (no impute)

dshemetov:
### Description
Currently, the identity smoother imputes nan values using the "savgol" smoother. This was the intended behavior for other smoothers (since not imputing while smoothing would lead to nan propagation), but the intention for the identity smoother is to be a true no-op by default. This fixes that issue.

### Changelog
Change the default argument handling: the default for "identity" is no imputation, while the default for anything else is "savgol". User-supplied impute arguments are handled normally.

### Fixes 
- Probably helps #1077. 
- May change the JHU and USAFacts raw signals (difference plots incoming).

### DIFF
--- a/_delphi_utils_python/delphi_utils/smooth.py
+++ b/_delphi_utils_python/delphi_utils/smooth.py
@@ -153,10 +153,9 @@ class Smoother:  # pylint: disable=too-many-instance-attributes
     def _select_imputer(self, impute_method, smoother_name):
         if impute_method is None and smoother_name != "identity":
             return "savgol"
-        elif impute_method is None and smoother_name == "identity":
+        if impute_method is None and smoother_name == "identity":
             return "identity"
-        else:
-            return impute_method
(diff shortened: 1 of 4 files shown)


krivard:
Fix your tests please

dshemetov:
Done!

dshemetov:
Don't merge yet, I will need to update Andrew's fix

chinandrew:
> Don't merge yet, I will need to update Andrew's fix

My fix? If you're referring to #1077, I just specified `impute_method = None`, which should work with both this PR and what is currently in main. Perhaps updating it to "identity" or removing the impute_method kwarg would be a bit cleaner once this is merged, but functionally the same.

dshemetov:
You're right, we don't have to update that. Ready to merge whenever.

krivard (approved):
LGTM!

PR #1081, opened 2021-06-24
Add workflow to automate fb package testing

nmdefries:
### Description
Add fb package unit tests to CI.

### Changes
- Add R survey GitHub Actions workflow file. Use R dependency caching to reduce set-up time. If dependencies are not in the cache, they are installed; if they are, they are updated (not reinstalled, to save time) to the newest version. Caches are [versioned using the current month and year](https://github.com/actions/cache#creating-a-cache-key) in the key name to keep required dependency updates small.
- Update handful of failing tests.
- Remove run on push to release branch from python CI workflow.

### Context
In its full form, the workflow takes awhile to run, ~25 min for all (unit and integration) tests. About half the time is spent on installing R packages and half on actually running the tests. Run time is reduced to ~15 min when using caching for R packages. This is still much longer than other CI tests take, mostly due to integration tests setup, so (for now) this will run unit tests only.


.github/workflows/r-ci.yml
```diff
@@ -0,0 +1,59 @@
...
+# documentation.
+#
+# See https://github.com/r-lib/actions/tree/master/examples#readme for
+# additional example workflows available for the R community.
+
+name: R facebook survey
+
+on:
+  push:
+    branches: [ main, prod, 'release/*' ]
```
chinandrew, line 13: I think we can remove the `'release/*' ` (also in the python one) since we don't use those anymore, is that right @krivard?

krivard: That's correct

chinandrew, line 13:
```suggestion
    branches: [ main, prod]
```

chinandrew (approved):
:shipit::shipit::shipit:

PR #1090, opened 2021-06-30
Standardization of Completed logging

### LINKED ISSUE #1087: Standardize contents of "Completed indicator run" messages
Currently, the only guaranteed field in "Completed indicator run" logging messages is "elapsed_time_in_seconds".

Other fields that may or may not be present include:

* max_lag_in_days
* csv_export_count

Ideally we want all indicators to provide both figures so we can more easily direct investigations into problems raised by Sir Complains-a-lot.

krivard:
### Description
Submit max min lag, min max date, and csv export count for all indicators.

I tried to do this largely the same way for all indicators wherever possible. The general strategy is:
1. for each signal configuration saved, grab the max date and the number of files exported
2. store as a list of tuples called `stats`
3. right before the Completed log, grab the min (earliest) max date and compute max lag
4. sum the number of files exported
5. log these in Completed

This was impossible for:
* doctor-visits
* hospital-admissions
* fb-survey
& we'll have to pick those up when we productionize them.

Variations noted in changelog below.

### Changelog
- changehc - had to nest `stats` in groups used within `update_sensor.py`
- covid_act_now - was already computing stats using another method, just added lag
- hhs_facilities - above method
- hhs_hosp - above method
- nchs_mortality - doesn't use create_export_csvs because epiweeks, so I added a `return` to its custom `export.py` to simulate it
- quidel_covidtest - extended `log_exit` to do final stats computation, since it uses atexit registration to perform Completed logging
- safegraph_patterns - added `stats` list to the function signature used as partials for workers in `process.py` (nb Python lists are thread-safe so this should be fine)

Remaining indicators (jhu, usa-facts, google-symptoms) already log these items using various other methods; I opted not to normalize the code.

### Fixes 
- Fixes #1087

### DIFF
--- a/changehc/delphi_changehc/run.py
+++ b/changehc/delphi_changehc/run.py
@@ -156,8 +156,7 @@ def run_module(params: Dict[str, Dict[str, Any]]):
         se = params["indicator"]["se"])
 
     ## start generating
-    min_max_date = None
-    csv_export_count = 0
+    stats = []
     for geo in params["indicator"]["geos"]:
         for numtype in params["indicator"]["types"]:
             for weekday in params["indicator"]["weekday"]:
@@ -182,18 +181,17 @@ def run_module(params: Dict[str, Dict[str, Any]]):
                 elif numtype == "cli":
                     data = load_cli_data(file_dict["denom"],file_dict["flu"],file_dict["mixed"],
                              file_dict["flu_like"],file_dict["covid_like"],dropdate_dt,"fips")
-                all_dates = su_inst.update_sensor(
+                more_stats = su_inst.update_sensor(
                     data,
                     params["common"]["export_dir"]
                 )
-                max_date = min(max(dates) for dates in all_dates)
-                if min_max_date is None or min_max_date > max_date:
-                    min_max_date = max_date
-                csv_export_count += sum(len(dates) for dates in all_dates)
+                stats.extend(more_stats)
 
             logger.info("finished processing", geo = geo)
 
     elapsed_time_in_seconds = round(time.time() - start_time, 2)
+    min_max_date = stats and max(s[0] for s in stats)
+    csv_export_count = sum(s[-1] for s in stats)
     max_lag_in_days = min_max_date and (datetime.now() - min_max_date).days
     formatted_min_max_date = min_max_date and min_max_date.strftime("%Y-%m-%d")
     logger.info("Completed indicator run",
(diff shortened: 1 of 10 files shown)


nchs_mortality/delphi_nchs_mortality/export.py
```diff
@@ -24,11 +24,14 @@ def export_csv(df, geo_name, sensor, export_dir, start_date):
...
+    dates = df["timestamp"].unique()
+    for date in dates:
         t = Week.fromdate(pd.to_datetime(str(date)))
         date_short = "weekly_" + str(t.year) + str(t.week).zfill(2)
         export_fn = f"{date_short}_{geo_name}_{sensor}.csv"
         result_df = df[df["timestamp"] == date][["geo_id", "val", "se", "sample_size"]]
         result_df.to_csv(f"{export_dir}/{export_fn}",
                          index=False,
                          float_format="%.8f")
+    print(dates)
```
krivard, line 36:
```suggestion
```

changehc/delphi_changehc/run.py
```diff
@@ -180,12 +181,21 @@ def run_module(params: Dict[str, Dict[str, Any]]):
                 elif numtype == "cli":
                     data = load_cli_data(file_dict["denom"],file_dict["flu"],file_dict["mixed"],
                              file_dict["flu_like"],file_dict["covid_like"],dropdate_dt,"fips")
-                su_inst.update_sensor(
+                more_stats = su_inst.update_sensor(
                     data,
                     params["common"]["export_dir"]
                 )
+                stats.extend(more_stats)
```
chinandrew, line 188: `update_sensor` returns the output of `create_export_csv`, which is a pd.Series of dates. Is there some tuple conversion somewhere that I missed?

changehc/delphi_changehc/run.py
```diff
@@ -180,12 +181,21 @@ def run_module(params: Dict[str, Dict[str, Any]]):
...
+
             logger.info("finished processing", geo = geo)
 
     elapsed_time_in_seconds = round(time.time() - start_time, 2)
+    min_max_date = stats and max(s[0] for s in stats)
```
chinandrew, line 193: Ignoring the above comment for a moment, is min_max_date supposed to be the smallest max date? e.g. `min(max(s) for s in states)`

EDIT: nevermind I see it now.

krivard: 🤦‍♀️

chinandrew (approved):
:shipit::shipit::shipit:

PR #1091, opened 2021-06-30
Survey codebook generator tuning on all QSFs

nmdefries:
### Description
Cleanup and tuning of codebook generator.

### Changelog
- Remove straggling `static_microdata_fields.csv`. New version already exists in the `static` dir.
- Explicitly set non-randomized questions to `NA` in one substep of the module assignment parsing. Prevents `NULL`s from causing problems in the `unlist()` step.
- Remove questions that aren't replacing any old items (replacement is set to `NA`) to simplify warning output.

### Fixes
Gets codebook generator to correctly ingest QSFs from all survey waves.

### DIFF
--- a/facebook/qsf-tools/generate-codebook.R
+++ b/facebook/qsf-tools/generate-codebook.R
@@ -198,7 +198,7 @@ process_qsf <- function(path_to_qsf,
     }
   })) %>%
     map(unique) %>%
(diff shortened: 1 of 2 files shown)


TaraLakdawala (commented):
static_microdata_fields.csv is removed and generate-codebook.R seems to handle all replacements of NULL with NA and removes questions that don't replace anything. This looks good to me!

PR #1099, opened 2021-07-07
Update validation logger

qx-teo:
### Description
Bugfix: allowing logger to work for indicators using the runner

### Fixes 
- Added __name__ param to fix indicators using the runner not having validator log files saved to Elastic
- Added summary output (checks run/failed/suppressed/warnings) to the logger


qx-teo:
@nmdefries

krivard:
@qx-teo ah sorry; you'll want to resubmit these as branches instead of forks so that the CI tests will run

krivard:
😕  mmmmm I'm used to PRs submitted from forks _not_ running CI tests, but it looks like they ran here. I guess this is safe? Regardless in future do them as branches; I've added you to @cmu-delphi/covidcast-engineering so you should have write access now

krivard:
Superseded by #1103

PR #1101, opened 2021-07-07
Validator bugfixes

qx-teo:
### Description
Bug fixes/improvements to missing_date, positive_negative_spikes, min_max_date, max_max_date, and geo_sig_date combos

### Changelog
- Incorporating max-lag and min-lag for date-related errors:
- check_min_max_date and check_max_max_date:
Instead of comparing whether it is today - expected lag <= recent date <= today, we use today - max_expected_lag <= recent_date <= today - min_expected_lag
- check_geo_sig_date: Only give an error if the missing geo_sig_date combo happens after the maximum lag
- check_missing_date: Missing dates only checked if the dates in the time window aren't within the expected lag period
- fix_positive_negative_spikes: Making sure that outliers can only be considered within the time window, and not the frame a day before, Making sure that df[-len(lower_index)] works in the case where lower_df is empty
- add_raised_error: raised_errors.append(Error) wasn't properly adding an error. Standardized the two instances with the other syntax of add_raised_error() method.

@nmdefries


nmdefries (changes requested):
CI is failing because of a lint error in `dynamic.py`. You can take a look at the checks below for details.

qx-teo:
Superceded by #1105

PR #1102, opened 2021-07-07
[fb-package] Split tests into unit and integration sections that can be run separately

nmdefries:
### Description
Split survey package tests into unit and integration sections.

### Changelog
- Move unit tests (those that don't require an initial pipeline run) into `unit-tests` dir.
- Move integration tests (those that do require an initial pipeline run) into `integration-tests` dir.
- Update `test` `make` target to run both sets of tests. (Each `R CMD check` call runs some other tests that we'd prefer to only run once, but as far as I can tell these can't be suppressed and `R CMD check` can only run tests in one dir at a time.)

### Fixes 
Allows fast (unit) and slow (integration, due to setup that runs the pipeline) tests to be run separately to help speed up survey checks in CI (#1081).


krivard (approved):
This is good. If we grow annoyed with the duplicated `R CMD check` work, we can always see if --test-dir will descend recursively and just wrap them both up in a general `tests` directory. Any numbers on how quickly the unit tests run?

nmdefries:
On my computer, the general package/format checks with unit tests take 30 s and with integration tests take ~4 min (seems faster than what GitHub Actions has been doing).

> see if --test-dir will descend recursively and just wrap them both up in a general tests directory

PR #1104, opened 2021-07-07
Validation params changes

qx-teo:
- Added "sunday+m,n" functionality for indicators that update at a set time each week
- Split up expected lag parameter into minimum and maximum expected lag, shifted it to common param
- Added 'all' keyword for default values in expected lag dictionaries
- Adjusted off by one error in span length, fixed tests

- Adjusted validation params:
-- Added min_expected_lag and max_expected_lag params
-- Fixed missing se/missing sample size allowed on certain indicators
-- Added log filepath into individual indicator templates
-- Added validation params to indicators that didn't have it
-- Added additional valid geo values to Safegraph due to geo coding bug

### Description
Adding functionality to validator params, and fixing default template values for params (along with a few minor bugfixes)

### Changelog
- dynamic.py and static.py: adding dependencies, class parameters
- utils: Fixing off by one error in TimeWindow, adding functionality for "sunday+"
- test_utils: Fixing test for TimeWindow error
- templates: in ansible folder and individual indicators

@nmdefries


combo_cases_and_deaths/params.json.template
```diff
@@ -1,13 +1,33 @@
 {
   "common": {
     "log_exceptions": false,
-    "export_dir": "./receiving"
+    "export_dir": "./receiving",
+    "log_filename": "./indicator-combination.log"
```
nmdefries, line 5: Not sure we want logs saved to a log file for local/dev runs (and etc for other template params). I believe it keeps them from being printed to std out. But also no strong feelings one way or the other.

covid_act_now/params.json.template
```diff
@@ -13,5 +13,33 @@
...
+         "signal": "pcr_specimen_total_tests"},
+        {"check_name": "check_se_not_missing_and_in_range",
+         "signal": "pcr_specimen_total_tests"},
+        {"check_name": "check_n_missing",
+         "signal": "pcr_specimen_total_tests"},
+        {"check_name": "check_se_0_when_val_0"}         
+      ]
+    },
+    "static": {
+      "minimum_sample_size": 100,
```
nmdefries, line 35: Covid Act Now was the one with really low sample sizes (like 0.7) being produced, right? Were you able to verify that we actually want to report those, or is 100 the desired sample size threshold?

ansible/templates/safegraph_patterns-params-prod.json.j2
```diff
@@ -19,6 +19,8 @@
       "data_source": "safegraph",
       "span_length": 14,
       "end_date": "today-4",
+      "min_expected_lag": {"all": "sunday+4,4"},
+      "max_expected_lag": {"all": "sunday+4,4"},
```
nmdefries, line 23: What does this mean vs (from HHS)

```
"min_expected_lag": {"all": "1"},
"max_expected_lag": {"all": "7"},
```

'cause HHS also updates weekly, so presumably we'd want to use the "sunday+" format.

covid_act_now/params.json.template
qx-teo: Yep, I'll need to change that - Katie says treat it as 0 for now

ansible/templates/safegraph_patterns-params-prod.json.j2
qx-teo: I think my checks found that HHS doesn't always update on the same day, and the documentation says "HHS issues updates to this timeseries once a week, and occasionally more often." I'll double check on whether or not to use "sunday+" for this.

nmdefries: Makes sense.

If min expected lag is 1, that means that for data issued today, the most recent date available is yesterday -- is that right? But I'm getting mixed messages on what the lag should be: According to [the HHS documentation](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/hhs.html#lag-and-backfill) lag is 0-6 days. Taking a quick look at our as-of/issue dates in the API, it looks like lag on issue is often 2 days.

qx-teo: The way I've written it right now, yes. I also saw that line in HHS documentation, but from my experience lag is rarely 0 days?

nmdefries: Okay, maybe the docs are out of date, then.

_delphi_utils_python/delphi_utils/validator/utils.py
```diff
@@ -77,3 +80,41 @@ def aggregate_frames(frames_list):
...
+            value_num = int(value)
+        return value_num
+
+    # Add 'all' to lag_dict keys, give it default value of 1
+    lag_dict["all"] = lag_dict.get("all", "1")
+
+    # Converting strings to numeric output
+    output_dict = {sig:value_interpret(lag_dict.get(
+        sig)) for sig in lag_dict.keys()}
+    return output_dict
```
qx-teo, line 120: lag_converter should be out of sync with validator_bugfixes now since I added a commit there, just putting a note here so I don't forget

nmdefries (approved):
Looks good! 👏  Great work figuring out what all the lags and update cycles are.

This is waiting on #1101 to merge first.

krivard:
You've got conflicts -- maybe due to colliding changes with #1105, which replaced #1101.

qx-teo:
@krivard conflicts should be resolved!

PR #1105, opened 2021-07-07
Validator bugfixes

qx-teo:
### Description
Bug fixes/improvements to missing_date, positive_negative_spikes, min_max_date, max_max_date, and geo_sig_date combos

### Changelog
- Incorporating max-lag and min-lag for date-related errors:
- check_min_max_date and check_max_max_date:
Instead of comparing whether it is today - expected lag <= recent date <= today, we use today - max_expected_lag <= recent_date <= today - min_expected_lag
- check_geo_sig_date: Only give an error if the missing geo_sig_date combo happens after the maximum lag
- check_missing_date: Missing dates only checked if the dates in the time window aren't within the expected lag period
- fix_positive_negative_spikes: Making sure that outliers can only be considered within the time window, and not the frame a day before, Making sure that df[-len(lower_index)] works in the case where lower_df is empty
- add_raised_error: raised_errors.append(Error) wasn't properly adding an error. Standardized the two instances with the other syntax of add_raised_error() method.

@nmdefries


_delphi_utils_python/delphi_utils/validator/dynamic.py
```diff
@@ -150,14 +155,18 @@ def validate(self, all_frames, report):
...
-                        ValidationFailure("check_missing_geo_sig_date_combo",
-                                          checking_date,
-                                          geo_type,
-                                          signal_type,
-                                          "test data for a given checking date-geo type-signal type"
-                                          " combination is missing. Source data may be missing"
-                                          " for one or more dates"))
+                    # If checking_date is within max_lag days, don't give error
+                    min_thres = timedelta(days = self.params.max_expected_lag.get(
+                        signal_type, self.params.max_expected_lag.get('all', 10)))
```
nmdefries, line 160: I thought default lag was going to be 1?

nmdefries (commented):
Initial comments, I haven't had the chance to go through the logic thoroughly yet.

_delphi_utils_python/delphi_utils/validator/static.py
```diff
@@ -92,7 +95,11 @@ def check_missing_date_files(self, daily_filenames, report):
             daily_filename[0][0:8], '%Y%m%d').date() for daily_filename in daily_filenames}
 
         # Diff expected and observed dates.
-        check_dateholes = list(set(self.params.time_window.date_seq).difference(unique_dates))
+        expected_dates = self.params.time_window.date_seq
+        max_expected_lag_overall = max(self.params.max_expected_lag.values())
+        expected_dates = [date for date in expected_dates if
+            abs((datetime.today().date() - date).days) > max_expected_lag_overall]
```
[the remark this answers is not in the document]
qx-teo: Maybe `expected_dates `was a bad term to use, what I meant is that `expected_dates `are all the days that should 100% be there. So if a particular date that falls between min and max expected lags aren't present, there shouldn't necessarily be a `missing_date_files `error.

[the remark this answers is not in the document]
nmdefries: Ah, yeah, I remember now :sweat_smile: Could be good to add a brief explanation like that as a comment.

_delphi_utils_python/delphi_utils/validator/dynamic.py
qx-teo: I'm not too sure about this, my idea is that max expected lag should be a decently large number and min expected lag should be decently small - if max lag defaults to 1 when a param is missing (which shouldn't happen as of now since I've put in all the lags), it might give a lot of false positives?

nmdefries: Okay, I'll revisit this when I go through the logic/design more carefully.

_delphi_utils_python/delphi_utils/validator/utils.py
```diff
@@ -77,3 +77,40 @@ def aggregate_frames(frames_list):
...
+    """Convert a dictionary of lag values into the proper format.
+
+    Parameters
+    ----------
+    lag_dict: Dict[str, str]
+        Keys are either 'all', or signal names
+        Values are either numeric or days of the week, represented by
+            "Sunday+[0-7],[0-9]", first part represents the day of the week
+            the upload happens, while the second number represents lag upon
+            upload.
```
nmdefries, line 91: It could be nice to support numeric lag inputs (`1` instead of `"1"`) as well.

_delphi_utils_python/delphi_utils/validator/dynamic.py
```diff
@@ -27,8 +27,10 @@ class Parameters:
         max_check_lookbehind: timedelta
         # names of signals that are smoothed (7-day avg, etc)
         smoothed_signals: Set[str]
-        # how many days behind do we expect each signal to be
-        expected_lag: Dict[str, int]
+        # maximum number of days behind do we expect each signal to be
+        max_expected_lag: Dict[str, int]
+        # minimum number of days behind do we expect each signal to be
+        min_expected_lag: Dict[str, int]
```
nmdefries, line 33: We also need to remove the code that handles the old `lag` style. I guess that's in `dynamic.py` since they were dynamic-specific settings.

_delphi_utils_python/delphi_utils/validator/utils.py
```diff
@@ -77,3 +77,40 @@ def aggregate_frames(frames_list):
...
+    Dict[str, int]
+        Keys are all active signal names for an indicator, or "all"
+        Values are obtained by looking at signal name entries,
+            If not present then by the 'all' indicator,
+            else 1.
+    """
+    def value_interpret(value):
+        """Convert value from string to numeric, including sunday+m,n."""
+        if value.startswith("sunday+"):
+            value_num = (date.today().isoweekday() - int(value[7:8]) - 1) % 7 + 1
```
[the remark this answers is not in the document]
qx-teo: I tried to make this clearer, but it's still a little confusing - is this better?

_delphi_utils_python/delphi_utils/validator/utils.py
```diff
@@ -77,3 +77,47 @@ def aggregate_frames(frames_list):
...
+            # Expected lag is 7 (in case updates aren't in yet)
+            value_num = (date.today().isoweekday() - int(value[7:8]) - 1) % 7 + 1
+            # Add on expected lag to lag from weekdays
+            value_num += int(value[9:])
+        else:
+            value_num = int(value)
+        return value_num
+
+    # Add 'all' to lag_dict keys, give it default value of 1
+    lag_dict["all"] = lag_dict.get("all", "1")
```
nmdefries, line 118: I'm not understanding the purpose of this bit. Above, like in `dynamic.py`, we're calling `self.params.min_expected_lag.get('all', 1)` and `self.params.max_expected_lag.get('all', 10)`, so we're already providing defaults for the "all" key. Also, you mentioned above that you wanted to use a larger value for max_lag, but here min and max lags are always going to be the same.

qx-teo: Hm, this is definitely a bit of redundancy. Would you prefer I remove it then? If yes, should I remove this line and keep the 1/10?

nmdefries: > should I remove this line and keep the 1/10?

Yes, I think it makes more sense to remove it here and fill in as needed in other parts of the code because: 1) you wanted to use different "all" defaults for min vs max lag and this doesn't allow that, and 2) we don't always need to have the "all" key.

_delphi_utils_python/delphi_utils/validator/static.py
```diff
@@ -96,7 +96,9 @@ def check_missing_date_files(self, daily_filenames, report):
 
         # Diff expected and observed dates.
         expected_dates = self.params.time_window.date_seq
-        max_expected_lag_overall = max(self.params.max_expected_lag.values())
+        max_expected_lag_overall = 10 if self.params.max_expected_lag else 
+            max(self.params.max_expected_lag.values())
```
nmdefries, line 100: Looks like the logic here is flipped (or you need a `not`)

qx-teo: Yeah i confused myself, should be fixed now

nmdefries (approved):
Looks good!

PR #1110, opened 2021-07-12
Reference API DF support

qx-teo:
### Description
Takes missing dates from API dataframe, and supplements it with test data for days before the checking date. Useful for non-daily updates (or large batch updates in event of outages)

### Changelog
- dynamic.py

### Fixes 
- Fixes #(issue)

### DIFF
--- a/_delphi_utils_python/delphi_utils/validator/dynamic.py
+++ b/_delphi_utils_python/delphi_utils/validator/dynamic.py
@@ -192,6 +192,21 @@ class DynamicValidator:
                                           "be performed"))
                     continue
 
+                # If API data is missing (e.g. non-daily updates),
+                # Take data from geo_sig_df
+                reference_api_df_max_date = reference_api_df.time_value.max()
+                if reference_api_df_max_date < reference_end_date:
+                    # Making sure geo_sig_df is similar:
+                    geo_sig_df_supplement = geo_sig_df.query(
+                        'time_value <= @reference_end_date & time_value > \
+                        @reference_api_df_max_date')[[
+                        "geo_id", "val", "se", "sample_size", "time_value"]]
+                    geo_sig_df_supplement["time_value"] = \
+                        pd.to_datetime(geo_sig_df_supplement["time_value"],
+                            format = "%Y-%m-%d %H:%M:%S")
+                    reference_api_df = pd.concat(
+                        [reference_api_df, geo_sig_df_supplement])
+
                 self.check_max_date_vs_reference(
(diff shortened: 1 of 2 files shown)


qx-teo:
@nmdefries

_delphi_utils_python/delphi_utils/validator/dynamic.py
```diff
@@ -192,6 +192,8 @@ def validate(self, all_frames, report):
                                           "be performed"))
                     continue
 
+                self.check_reference_api_df(reference_api_df, geo_sig_df, reference_end_date)
```
nmdefries, line 195: This isn't actually saving the returned padded ref API data anywhere, so subsequent checks will still use the un-padded reference data.

_delphi_utils_python/delphi_utils/validator/dynamic.py
```diff
@@ -254,6 +256,32 @@ def check_max_allowed_max_date(self, max_date, geo_type, signal_type, report):
 
         report.increment_total_checks()
 
+    def check_reference_api_df(self, reference_api_df, geo_sig_df, reference_end_date):
```
nmdefries, line 259: Looks like this isn't actually checking anything, so a different name would be more suitable, e.g. "pad_to_date" or something.

nmdefries (changes requested):

_delphi_utils_python/delphi_utils/validator/dynamic.py
```diff
@@ -254,6 +256,32 @@ def check_max_allowed_max_date(self, max_date, geo_type, signal_type, report):
...
+        if reference_api_df_max_date < reference_end_date:
+            # Querying geo_sig_df, only taking relevant rows
+            geo_sig_df_supplement = geo_sig_df.query(
+                'time_value <= @reference_end_date & time_value > \
+                @reference_api_df_max_date')[[
+                "geo_id", "val", "se", "sample_size", "time_value"]]
+            # Matching time_value format
+            geo_sig_df_supplement["time_value"] = \
+                pd.to_datetime(geo_sig_df_supplement["time_value"],
+                    format = "%Y-%m-%d %H:%M:%S")
```
nmdefries, line 280: This logic is fairly complicated (easy to have an off-by-one error on the dates, for example) so it'd good to add a few tests.

_delphi_utils_python/delphi_utils/validator/dynamic.py
qx-teo: Good catch! I'll add it in

nmdefries (approved):

PR #1111, opened 2021-07-12
Cast value column to float in HHS before downstream operations

chinandrew:
### Description
The value column was getting dropped by the geomapper during a groupby, which seems to be due to a new pandas release (downgrading fixes the issue). Casting to a float prevents this from happening.

### Changelog
- Cast value column in dataframe to a float right after it's generated

### Fixes 
- Fixes #(issue)


krivard:
@nmdefries the R tests (well, building the dependencies) ran for 81 minutes before github cut it off; I've restarted the job but if it happens again we may have to back out #1081

krivard (approved):
👍  pending build

PR #1113, opened 2021-07-12
[fb-package] Modify output when debug flag is on

nmdefries:
### Description
When `debug=true` in params, print message and raise warning, and add message to beginning of contingency table names. Add supporting tests.

### DIFF
--- a/facebook/delphiFacebook/R/contingency_run.R
+++ b/facebook/delphiFacebook/R/contingency_run.R
@@ -8,6 +8,12 @@
 #'
 #' @export
 run_contingency_tables <- function(params) {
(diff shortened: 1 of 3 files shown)


capnrefsmmat (approved):
Looks good

capnrefsmmat:
Actually, is there any circumstance we could make the same mistake when manually generating API data?

nmdefries:
Yeah, probably. There aren't any filters on sample size after data generation.

capnrefsmmat:
Ok, so we might add the same warning to the code that generates the files that go  into receiving.

PR #1118, opened 2021-07-13
Reduce fb package testing setup time

nmdefries:
### Description
Take 2 of #1081 to reduce package install time. Takes ~8 min to run with no cache; time with cache of dependencies takes ~2 min, as in #1081.

### Changelog
- Don't install "suggested" dependencies of survey package's dependencies or of `rcmdcheck` and `mockr`.


chinandrew (approved):
:shipit: :shipit: :shipit:

PR #1119, opened 2021-07-14
Prevent multiple log outputs

qx-teo:
Multiple calls to get_structured_logger creating multiple FileHandlers, this checks to see if a handler is already present, and if so doesn't add another one.

### Description
Bug fix: Preventing multiple duplicate log messages in the output due to more than one FileHandler being instantiated.

### Changelog
- logger.py (get_structured_logger function)

### Fixes 
Checks to see if a handler is already present, and if so doesn't add another one. Referring to [this link](https://stackoverflow.com/a/6729713) for solution


nmdefries (approved):
Looks good.

PR #1124, opened 2021-07-17
Change logger summary output for elastalert compatibility

qx-teo:
### Description
Change logger summary output for elastalert compatibility

### Changelog
- Broke summary output down from having all the information under the "event" to separate fields, that elastalert can parse
- Also standardized event message so that elastalert can pull all the summary messages.


qx-teo:
Superceded by #1130

PR #1127, opened 2021-07-19
Add missing sources to sirCAL

krivard:
### Description

Adds nchs-mortality, covid-act-now, and hhs (hospitalizations) to sir CAL dev and prod params.

### Changelog
- ansible/templates/sir_complainsalot-params-prod.json.j2
- sir_complainsalot/params.json.template

### DIFF
--- a/ansible/templates/sir_complainsalot-params-prod.json.j2
+++ b/ansible/templates/sir_complainsalot-params-prod.json.j2
@@ -48,7 +48,7 @@
       "retired-signals": ["raw_pct_negative","smoothed_pct_negative","raw_tests_per_device","smoothed_tests_per_device"]
     },
     "nchs-mortality": {
-      "max_age":13,
+      "max_age":16,
       "maintainers": []
     },
     "covid-act-now": {
--- a/ansible/templates/sir_complainsalot-params-prod.json.j2
+++ b/ansible/templates/sir_complainsalot-params-prod.json.j2
@@ -46,6 +46,18 @@
       "max_age":6,
       "maintainers": ["U01AP8GSWG3","U01069KCRS7"],
       "retired-signals": ["raw_pct_negative","smoothed_pct_negative","raw_tests_per_device","smoothed_tests_per_device"]
+    },
+    "nchs-mortality": {
+      "max_age":13,
+      "maintainers": []
+    },
+    "covid-act-now": {
+      "max_age":5,
+      "maintainers": []
+    },
+    "hhs": {
+      "max_age":8,
+      "maintainers": []
     }
   }
 }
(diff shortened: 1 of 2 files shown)


ansible/templates/sir_complainsalot-params-prod.json.j2
```diff
@@ -46,6 +46,18 @@
       "max_age":6,
       "maintainers": ["U01AP8GSWG3","U01069KCRS7"],
       "retired-signals": ["raw_pct_negative","smoothed_pct_negative","raw_tests_per_device","smoothed_tests_per_device"]
+    },
+    "nchs-mortality": {
+      "max_age":13,
```
qx-teo, line 51: Just checking if max_age is similar to the max_lag_in_days argument in the indicator lags - if yes, I believe that NCHS has lags of up to 15/16?

qx-teo (changes requested):
Everything looks good, just double checking on the values themselves

ansible/templates/sir_complainsalot-params-prod.json.j2
krivard: ooh interesting. Yes it should generally be the same (or one day greater to allow for acquisition lag), but I was going off the most recent lag pattern from the last month, which has been 12 at most over the ~3 weekly cycles that the indicator has been logging max lags.

image

Is the 15-16 figure coming from some other output I should be looking at, or from the indicator logic?

qx-teo: Hm, I downloaded the info from most likely the same source and got up to 16:

image

The graph values actually seem to resemble what I have, but there's a truncating at 12. Not too sure why...

qx-teo: The search I used was something like log.file.path:/var/log.indicators/*.log AND message: max_lag_in_days. It's also 7 days old, so many mine's just outdated.

krivard: serves me right, turns out I set a max of 12 in the graph y-axis limits, and kibana doesn't make it obvious when you've got data that exceeds the limits 😖 

17 it is!

krivard, line 51:
```suggestion
      "max_age":16,
```

qx-teo:
Hey @krivard, any idea why the check is failing? I don't know if I should approve this since it looks fine otherwise.

qx-teo:
@nmdefries , the R FB check seems to be angry again!

sir_complainsalot/params.json.template
```diff
@@ -47,6 +47,18 @@
       "max_age":6,
       "maintainers": ["U01AP8GSWG3","U01069KCRS7"],
       "retired-signals": ["raw_pct_negative","smoothed_pct_negative","raw_tests_per_device","smoothed_tests_per_device"]
+    },
+    "nchs-mortality": {
+      "max_age":13,
```
qx-teo, line 52:
```suggestion
      "max_age":16,
```

Mirroring the ansible template changes:

qx-teo (changes requested):

qx-teo (approved):
All good!

Issue #1122, opened 2021-07-15
Validator check_se_not_missing_and_in_range not working for indicators with SE but no sample size


qx-teo:
**Summary**
Validator check_se_not_missing_and_in_range uses the following formula to generate upper limits for standard errors:
`df_to_test.eval('se_upper_limit = (val * sample_size + 50)/(sample_size + 1)', inplace=True)`

This creates NAs when sample sizes aren't available, and automatically fails the test.

Left as an issue since currently all indicators either have both SE and N, or have neither of those (therefore bug does not appear in real-world tests)

**Possible Solutions**
Modifying code to provide warnings when N not available
Using only [0,50] limit rather than [0, min(50, val * (1 + eps))]
Finding alternative upper bound when N is not available

qx-teo:
@nmdefries

nmdefries:
Talked to Katie about this: It's not clear why we need to check SEs against a calculated upper limit. Verifying that SEs are positive and non-zero is sufficient, and we can just remove the upper limit bit.

PR #1128, opened 2021-07-20
Remove SE Upper Limit

qx-teo:
### Description
Removing SE upper limit from checks, as per discussion in #1122 

### Changelog
- static.py

### Fixes 
- Removed upper limit since we only need to check for non-negativity

### DIFF
--- a/_delphi_utils_python/delphi_utils/validator/static.py
+++ b/_delphi_utils_python/delphi_utils/validator/static.py
@@ -324,22 +324,18 @@ class StaticValidator:
         Returns:
             - None
         """
-        # Add a new se_upper_limit column.
-        df_to_test.eval(
-            'se_upper_limit = (val * sample_size + 50)/(sample_size + 1)', inplace=True)
 
         df_to_test['se'] = df_to_test['se'].round(5)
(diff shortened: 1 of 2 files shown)


nmdefries (changes requested):
Lines 328-342 also need revising.

_delphi_utils_python/delphi_utils/validator/static.py
```diff
@@ -324,34 +324,29 @@ def check_bad_se(self, df_to_test, nameformat, report):
         Returns:
             - None
         """
-        # Add a new se_upper_limit column.
-        df_to_test.eval(
-            'se_upper_limit = (val * sample_size + 50)/(sample_size + 1)', inplace=True)
-
         df_to_test['se'] = df_to_test['se'].round(5)
```
nmdefries, line 327: We should be able to get rid of this rounding step as well.
```suggestion
```

nmdefries (approved):
One more line that can be removed for conciseness, but other than that looks good!

nmdefries:
There's a linting issue.

nmdefries (approved):

nmdefries:
Closes #1122.

PR #1129, opened 2021-07-20
[fb-package] Fix failing contingency CI due to `readr` version change

nmdefries:
### Description
Prevent survey CI from updating to new major release of `readr`. A couple bugs ([apparently spurious compilation errors on install ](https://github.com/tidyverse/readr/issues/1233) and [failure to write empty files](https://github.com/tidyverse/readr/issues/1234)) cause test failures. I've filed issues and hopefully those are fixed in the next minor version; will keep an eye on future `readr` releases.

### Changelog
- Remove `path` argname. Argument will by default use the new `file` arg.
- Exclude `readr` from being updated in Actions workflow.

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -48,6 +48,10 @@ jobs:
             ${{ runner.os }}-r-facebook-survey-
       - name: Install R dependencies
         run: |
+          if ( packageVersion("readr") != "1.4.0" ) {
+            install.packages("devtools")
+            devtools::install_version("readr", version = "1.4.0")
+          }
           install.packages("remotes")
           remotes::update_packages(c("rcmdcheck", "mockr"), upgrade="always")
           dependency_list <- remotes::dev_package_deps(dependencies=TRUE)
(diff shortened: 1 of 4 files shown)


chinandrew:
Just so I understand, does the cache in CI not have the new major version yet, since I don't see a pinned version anywhere? If so, how did that happen if it's currently set to always update? Certainly there was some CI run in the last day that would've triggered it.

nmdefries:
> does the cache in CI not have the new major version yet

Nope, the cache has the old version of `readr` (1.4.0). A new version of the cache is only saved once a month (more on this below). When the cache is loaded, any packages in it that are out of date are updated to their newest versions. However, those updates only apply to the current run. Once a cache with a given name (key) is created, it can't be overwritten, so package updates aren't saved anywhere.

For clarity, I'll explicitly install the 1.4.0 version.

On monthly caching: thinking about it more, it makes more sense to move this to daily versioning. I was initially trying to minimize the number of times we would have to run the full install process (since it takes a long time), but with restore keys cache updates are incremental. I'll make a separate PR for that.

chinandrew (approved):
:shipit: :shipit: :shipit:

PR #1132, opened 2021-07-20
Allow contingency runs to fetch sparse data if none provided

nmdefries:
### Description
Introduce logic to automatically set desired input files to a sparse list (every fourth day of data + the last day in a period), plus associated fixes and improvements.

These changes were originally introduced in #988.

### Changelog
- Create function to filter filenames down to only those for every fourth day of data (since there is overlap between data issues).
- Fix how end date is defined for each file. The first date in the name is actually the issue date; we want the third date in the filename.
- Various updates to get tests working: changes to filename end date specification; missing expectations; remove list of contingency input files during integration teardown.
- If params end_date is missing, use today's date.
- Bonus: `contingency-combine.R` (script for combining output files across different time periods) expects exactly 2 args.
- Bonus: alphabetically sort grouping variables for output filenames.


krivard (approved):
👍  nice work!

PR #1133, opened 2021-07-20
Switch survey CI to use daily caching

nmdefries:
### Description
Switch survey CI to use daily caching.

### Changelog
- Use full date ("YYYYMMDD") in cache name instead of month.

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -35,15 +35,15 @@ jobs:
       - name: Install linux dependencies
         run: |
             sudo apt-get install libcurl4-openssl-dev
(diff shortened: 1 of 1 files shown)


chinandrew:
This means one run a day will take a while right?

nmdefries:
No, it should only take a few minutes (the one here took 4 min total). Each new daily run will load the cache from the previous day, which already has most of the required packages, so we only have to run incremental updates, we don't have to reinstall everything. And then the old packages + updates get saved to the new cache.

chinandrew (approved):
:shipit::shipit::shipit:

PR #1121, opened 2021-07-14
Fix too many statements linting error in CI

qx-teo:
### Description
Fixes #1120.
Created separate function to create reference and geo_sig dfs

### Changelog
Itemize code/test/documentation changes and files added/removed.
- dynamic.py

PR #1134, opened 2021-07-21
Add max_expected_lag check to check_missing_geo_sig_date_combo

qx-teo:
Checks to see if the checking date is within than max_lag days away, if yes then ignore and don't throw an error

### Description
Re-adding the max_expected_lag fix that got removed accidentally when create_dfs() function was added:
Used to exist in #1105 under line 161 [here](https://github.com/cmu-delphi/covidcast-indicators/pull/1105/files#), but got replaced by #1121 under line 249 [here](https://github.com/cmu-delphi/covidcast-indicators/pull/1121/files).

### Changelog
- dynamic.py

### Fixes 
- Added code back
- Changed direction of sign to match with the logic (for another example, can consider line 180 in dynamic.py under check_min_max_date()

### DIFF
--- a/_delphi_utils_python/delphi_utils/validator/dynamic.py
+++ b/_delphi_utils_python/delphi_utils/validator/dynamic.py
@@ -246,15 +246,18 @@ class DynamicValidator:
         report.increment_total_checks()
 
         if recent_df.empty:
-            report.add_raised_error(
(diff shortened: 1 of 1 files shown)


nmdefries (approved):
:+1:

PR #1148, opened 2021-07-28
Add bash initializer needed for temp environment

krivard:
### Description
Dangling change to Makefile so that we get the right Python when we run in the temporary production environment. Most systems won't need this. In tmp-prod, the rc file contains the following content:

```
scl enable rh-python38 bash
```

### Changelog
- Makefile -- source an rc file when constructing the shell
- bash-init.sh.template -- empty; copy this to bash-init.sh to use

### DIFF
--- a/facebook/Makefile
+++ b/facebook/Makefile
@@ -1,4 +1,4 @@
-SHELL:=/bin/bash
+SHELL:=/bin/bash --rcfile bash-init.sh
 
 # dry-run mode: generate all files, but do not post them anywhere, and disable all emails to outside parties.
 DRY:=yes


nmdefries (commented):
My understanding is that this change, if `bash-init.sh` is not empty, runs all make targets in a new bash instance with some specified version of Python rather than using the version of Python already installed. And we need Python 3.8 specifically because of this line `python3.8 -m venv env` in the make `install-python` target.

Since we're using a docker container, couldn't we explicitly install Python 3.8 there instead? Of course, the approach implemented here is more flexible, since it doesn't require a docker container.

krivard:
Clarification: this rc file is needed for the current execution environment on bigchunk-dev-02; it permits running a necessary setup command _before_ `source env/bin/activate`. Ultimately, the env python executable is still used, we're not trying to swap in a non-indicator-approved python version here. RedHat just needs us to give the OS a heads-up before we run a version of python other than the system version.

More information: https://developers.redhat.com/blog/2018/08/13/install-python3-rhel#software_collection_tips

krivard:
(I anticipate the final containerized version won't need this rc file)

nmdefries (approved):
Got it and looks good!

PR #1136, opened 2021-07-24
Remove 7dav from cumulative signals

chinandrew:
### Description
These signals are nonsensical and can be removed.

### Changelog
Itemize code/test/documentation changes and files added/removed.
- Remove 7dav signals for usafacts, jhu,  and combined cumulative signals by  skipping them in the variants loop
- Update test + add test for combo indicator (basically copy pasted from jhu)

### Fixes 
- Fixes #1131

PR #1149, opened 2021-07-28
Remove cumulative 7dav signals from consideration by SirCAL

krivard:
### Description
Followup from #1136; now that we're no longer producing these signals, we shouldn't have Sir Complains-a-lot check them for freshness...

### Changelog
- dev and prod params for sir_complainsalot


jingjtang (approved):
Lgtm

PR #1154, opened 2021-07-29
[fb-package] Clean up test warnings

nmdefries:
### Description
Clean up noisy warnings and notes in tests, largely due to non-standard evaluation in `dplyr` and `data.table`.

### Changelog
- Explicitly import stats and dplyr functions.
- Use `.data` pronoun to prevent column names from being flagged as unrecognized global variables.
- Swap `data.table` `setindex` and `setkey` for `setindexv` and `setkeyv` so column names can be provided as strings.


capnrefsmmat (approved):
this looks like fun!

PR #1159, opened 2021-07-30
Geo type mapping

### LINKED ISSUE #1135: Geo Mapper: Return all contained/child regions
Given a geo id (e.g. "ca" for California, a state) and an enclosed geo type (e.g. "county"), return a set of geo ids of the specified type that lie within the specified geo id (e.g. all counties within California)

We don't need to support all possible pairs. For now, we just need:

* all states within a nation
* all counties within a state
* all states within an hhs region

We should _not_ support the following pairs, because they do not nest neatly (some regions overlap more than one region of the larger geo type):

* all msas within a state
* all zips within a county
* possibly others, extend this list as you find them

Related to: https://github.com/cmu-delphi/delphi-epidata/issues/542

zhuoran-Cheng16:
### Description
add Geo type Mapper function in GeoMapper class: Return all contained/child regions

### Changelog
- add **get_geo_mapper** function 
- add unit test of **get_geo_mapper**

### Fixes 
- Fixes #1135


_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -595,3 +595,77 @@ def get_geo_values(self, geo_type):
             crosswalk = pd.read_csv(stream, dtype=str)
             self.geo_lists[geo_type] = set(crosswalk[geo_type])
             return self.geo_lists[geo_type]
+
+    def get_geo_mapper(self, geo_id, geo_type,geo_id_type):
```
krivard, line 599: Name change:
```suggestion
    def get_geos_within(self, geo_id, geo_type,geo_id_type):
```

krivard (approved):
👍  in broad strokes, looping in Dmitry to check the details

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -595,3 +595,77 @@ def get_geo_values(self, geo_type):
...
+                    index=list(crosswalk2["fips"]).index(fips_id)
+                    state_id=crosswalk2["state_id"][index]
+                    if state_id not in self.geo_lists["state_id"]:
+                        self.geo_lists["state_id"].add(state_id)
+                geo_id_set=self.geo_lists["state_id"]
+        elif geo_type=="county" and geo_id_type=="state":
+            stream = pkg_resources.resource_stream(
+                __name__, self.crosswalk_filepaths["fips"]["state"]
+            )
+            crosswalk = pd.read_csv(stream, dtype=str)
```
dshemetov, line 662: Same comment about `_load_crosswalk_from_file` here.

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -595,3 +595,77 @@ def get_geo_values(self, geo_type):
...
+            self.geo_lists["fips"]=set()
+            for i in range(1,len(crosswalk["state_id"])):
+                if crosswalk["state_id"][i]==geo_id:
+                    self.geo_lists["fips"].add(crosswalk["fips"][i])
```
dshemetov, line 666: Same comment about `self.geo_lists` here.

dshemetov (changes requested):
Have some comments about things that could reuse existing functions and spotted a possible bug with using the internal cache variable. Otherwise looks good 👍 .

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -595,3 +595,59 @@ def get_geo_values(self, geo_type):
...
+        Set of geo ids of the specified type that lie within the specified container geocode,
+        all in string format.
+        """
+        if child_geocode_type not in ("county","state"):
+            raise ValueError("child_geocode_type must be one of state, nation and hhs")
+        get_geo_values=set()
+        if child_geocode_type=="state":
+            if container_geocode_type=="nation" and container_geocode=="us":
+                crosswalk = self._load_crosswalk_from_file("state", "state")
+                get_geo_values=set(crosswalk["state_id"])
```
dshemetov, line 633: Ah so `get_geo_values` sounds like a function name (verb_noun), I would suggest renaming it to just `geo_values` (noun). That's the typical style I try to keep consistent, so that functions and variables can be easily distinguished.

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -595,3 +595,59 @@ def get_geo_values(self, geo_type):
...
+            if container_geocode_type=="hhs":
+                crosswalk_fips_to_hhs = self._load_crosswalk_from_file("fips", "hhs")
+                crosswalk_fips_to_state = self._load_crosswalk_from_file("fips", "state")
+                fips_hhs=[]
+                for i in range(1,len(crosswalk_fips_to_hhs["hhs"])):
+                    if crosswalk_fips_to_hhs["hhs"][i]==container_geocode:
+                        fips_hhs.append(crosswalk_fips_to_hhs["fips"][i])
```
dshemetov, line 640: This for-loop gets the job done and I think it's correct! However, I want to show you a really cool thing in Pandas that lets you do this kind of selection in a really slick way (I think you're gonna like how clean this is):
`crosswalk_fips_to_hhs[crosswalk_fips_to_hss["hhs"] == container_geocode]`

I'll walk through what's happening here:
1. Since `crosswalk_fips_to_hhs` is a `pd.DataFrame`, the result of `crosswalk_fips_to_hss["hhs"] == container_geocode` is a Pandas boolean Series with the same index as `crosswalk_fips_to_hhs` and True/False values if the original row had an "hhs" column equal to `container_geocode`.
2. Now the really cool thing is that you can feed `boolean_series = (crosswalk_fips_to_hss["hhs"] == container_geocode)` back into `crosswalk_fips_to_hss` and just get exactly the rows that were True.

[Check out Panda's more general guide](https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#boolean-indexing) for more examples on these sorts of things.

Try refactoring these lines using these methods!

PS. Also once you get a dataframe of all the fips codes that have a particular hhs, you will need to use [`pd.DataFrame.isin`](https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#indexing-with-isin) to select states with fips codes in that dataframe.

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -595,3 +595,59 @@ def get_geo_values(self, geo_type):
...
+                for i,fips_id in enumerate(fips_hhs):
+                    index=list(crosswalk_fips_to_state["fips"]).index(fips_id)
+                    state_id=crosswalk_fips_to_state["state_id"][index]
+                    if state_id not in get_geo_values:
+                        get_geo_values.add(state_id)
+        elif child_geocode_type=="county" and container_geocode_type=="state":
```
dshemetov, line 646: You know what, can we change `child_geocode_type` to `contained_geocode_type`? I just realized that we're mixing metaphors here (parent, child) <-> (container, contained) and its slightly confusing.

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -595,3 +595,59 @@ def get_geo_values(self, geo_type):
...
+            crosswalk = self._load_crosswalk_from_file("fips", "state")
+            for i in range(1,len(crosswalk["state_id"])):
```
dshemetov, line 648: Try doing the same refactoring with Pandas dataframes here!

dshemetov (changes requested):

dshemetov:
Found a refactoring suggestion for you using some cool Pandas tricks and a few minor nits. After that we'll be golden!

_delphi_utils_python/delphi_utils/geomap.py
zhuoran-Cheng16: Gotcha!

_delphi_utils_python/delphi_utils/geomap.py
zhuoran-Cheng16: That's really cool. Thanks for the explanation!

dshemetov:
:shipit: :shipit: :shipit:

dshemetov (approved):

Issue #1082, opened 2021-06-28
Permit Sir Complains-a-lot params to specify geo levels which have been retired for some signals


krivard:
The response rate for surveys has been dropping steadily since January as public interest in the pandemic wanes. This has caused many signals to no longer be able to meet minimum sample size thresholds for some geo levels, such as county or hrr.

Sir Complains-a-lot currently assumes all active signals are available at all geo levels present in the metadata. [The params file](https://github.com/cmu-delphi/covidcast-indicators/blob/main/ansible/templates/sir_complainsalot-params-prod.json.j2) lets you retire a signal entirely, but we'd like to keep checking whatever geo levels are still available for these signals during their decline.

Create a way for the params file to specify signal+geo pairs which have been retired. Ideally this should take inspiration from [the way we specify suppressed errors for the validator](https://github.com/cmu-delphi/covidcast-indicators/tree/main/_delphi_utils_python/delphi_utils/validator#customization).

PR #1174, opened 2021-08-05
Add support for retiring specific geo types to SirCAL

chinandrew:
### Description
SirCAL currently can only retire entire signals but not individual geo types. this PR adds support for retiring a subset of geos for a given signal. It is backwards compatible with the existing params files so they do not need to be altered.

### Changelog
- Add support for lists in retired-signals specifying multiple geo types to remove. Syntax determined via discussion with @krivard 
- Add tests for new function for determining retired signals
- Enable CI for sircal
- Fix linting errors to pass CI.

### Fixes 
- #1082

### DIFF
--- a/sir_complainsalot/delphi_sir_complainsalot/check_source.py
+++ b/sir_complainsalot/delphi_sir_complainsalot/check_source.py
@@ -1,16 +1,18 @@
+"""Evaluate data source for problems."""
+
 from dataclasses import dataclass
+from datetime import datetime, timedelta
 from typing import List
-from delphi_utils import get_structured_logger
 
 import covidcast
-from datetime import datetime, timedelta
-import numpy as np
 import pandas as pd
 
 covidcast.covidcast._ASYNC_CALL = True  # pylint: disable=protected-access
 
 @dataclass
 class Complaint:
+    """Container for a SirCAL complaint."""
+
     message: str
     data_source: str
     signal: str
@@ -20,19 +22,18 @@ class Complaint:
 
     def __str__(self):
         """Plain text string form of complaint."""
-
         return "{source}::{signal} ({geos}) {message}; last updated {updated}".format(
             source=self.data_source, signal=self.signal, geos=", ".join(self.geo_types),
             message=self.message, updated=self.last_updated.strftime("%Y-%m-%d"))
 
     def to_md(self):
         """Markdown formatted form of complaint."""
-
         return "*{source}* `{signal}` ({geos}) {message}; last updated {updated}.".format(
             source=self.data_source, signal=self.signal, geos=", ".join(self.geo_types),
             message=self.message, updated=self.last_updated.strftime("%Y-%m-%d"))
 
-def check_source(data_source, meta, params, grace, logger):
+
+def check_source(data_source, meta, params, grace, logger):  # pylint: disable=too-many-locals
     """Iterate over all signals from a source and check for problems.
 
     Possible problems:
@@ -50,7 +51,6 @@ def check_source(data_source, meta, params, grace, logger):
     10 days means we check the most recent 10 days of data. Defaults to 7.
 
     """
-
     source_config = params[data_source]
     gap_window = pd.Timedelta(days=source_config.get("gap_window", 7))
     max_allowed_gap = source_config.get("max_gap", 1)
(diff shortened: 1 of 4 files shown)


sir_complainsalot/delphi_sir_complainsalot/check_source.py
```diff
@@ -63,8 +63,8 @@ def check_source(data_source, meta, params, grace, logger):
     gap_complaints = {}
 
     for _, row in signals.iterrows():
-        if "retired-signals" in source_config and \
-           row["signal"] in source_config["retired-signals"]:
+        retired_signals = source_config.get("retired-signals")
```
krivard, line 66: probably should move this outside the loop -- there are like 600 rows in meta for some of the sources

sir_complainsalot/delphi_sir_complainsalot/check_source.py
```diff
@@ -100,7 +100,8 @@ def check_source(data_source, meta, params, grace, logger):
         if current_lag_in_days > source_config["max_age"] + grace:
             if row["signal"] not in age_complaints:
                 age_complaints[row["signal"]] = Complaint(
-                    "is {current_lag_in_days} days old".format(current_lag_in_days=current_lag_in_days),
+                    "is {current_lag_in_days} days old".format(
+                        current_lag_in_days=current_lag_in_days),
```
krivard, line 104:
```suggestion
                    f"is {current_lag_in_days} days old",
```

?

sir_complainsalot/delphi_sir_complainsalot/check_source.py
```diff
@@ -156,3 +157,15 @@ def check_source(data_source, meta, params, grace, logger):
...
+
+def _is_retired(row, retired_signals):
+    """Determine if a row of the metadata belongs to a retired signal according to the params."""
+    if not retired_signals:
+        return False
+    for signal in retired_signals:
+        if row["signal"] == signal:
+            return True
+        if isinstance(signal, list) and {row["signal"], row["geo_type"]}.issubset(set(signal)):
+            return True
```
krivard, line 170: this function gets run about 1500 times, so it might be worth worrying about speed here

krivard: (current sirCAL running time is about a half hour)

krivard (approved):
good tests, minor complexity & style questions but this will work as implemented so up to you

sir_complainsalot/delphi_sir_complainsalot/check_source.py
chinandrew: hmm I think it should be ok for now since the retired_signals list isn't very long. fb-survey is length 54 at the moment, but the others have ~4 signals.
```
In [20]: retired_signals = [list(range(5)) for i in range(100)]

In [21]: meta = covidcast.metadata()

In [22]: meta.shape
Out[22]: (1703, 15)

In [23]: %%time
    ...: for _, row in meta.iterrows():
    ...:     if _is_retired(row, retired_signals):
    ...:         continue
    ...: 
CPU times: user 2.26 s, sys: 0 ns, total: 2.26 s
Wall time: 2.26 s

In [24]: retired_signals = [list(range(5)) for i in range(10)]

In [25]: %%time
    ...: for _, row in meta.iterrows():
    ...:     if _is_retired(row, retired_signals):
    ...:         continue
    ...: 
CPU times: user 390 ms, sys: 7.86 ms, total: 398 ms
Wall time: 397 ms
```

I tried speeding it up with pandas apply, but that doesn't change much
```
In [72]: retired_signals = [list(range(5)) for i in range(100)]

In [73]: %time meta.apply(_is_retired, axis=1, retired_signals=retired_signals)
CPU times: user 2.27 s, sys: 0 ns, total: 2.27 s
Wall time: 2.27 s

In [74]: retired_signals = [list(range(5)) for i in range(10)]

In [75]: %time meta.apply(_is_retired, axis=1, retired_signals=retired_signals)
CPU times: user 273 ms, sys: 3.96 ms, total: 277 ms
Wall time: 276 ms

```

However, it is a bit neater and so I've put that line together with the data source filtering.

krivard: 👍

krivard (approved):
👍  tidier loops ftw

PR #1153, opened 2021-07-28
[fb package] Add a bunch more Wave 11 indicators to API

nmdefries:
### Description
Add lower-priority Wave 11 indicators to API.

### Changelog
- Added new `code_news_and_info` function.
- Added new indicators.

PR #1175, opened 2021-08-06
[fb-survey] Expand contingency with lower-priority Wave 11 indicators

nmdefries:
### Description
Contingency table variant of #1153. Adds a bunch more indicators.

### Changelog
- Add more indicators to `contingency_indicators.R`
- Define supporting variables in `contingency_variables.R`
- Add function `code_addl_activities` to create masking while doing activities.


facebook/delphiFacebook/R/contingency_indicators.R
```diff
@@ -68,14 +68,64 @@ get_aggs <- function() {
...
+    "pct_symp_diff_breathing", "symp_diff_breathing", compute_binary, jeffreys_binary,
+    "pct_symp_fatigue", "symp_fatigue", compute_binary, jeffreys_binary,
+    "pct_symp_nasal_congestion", "symp_nasal_congestion", compute_binary, jeffreys_binary,
+    "pct_symp_runny_nose", "symp_runny_nose", compute_binary, jeffreys_binary,
+    "pct_symp_aches", "symp_aches", compute_binary, jeffreys_binary,
+    "pct_symp_sore_throat", "symp_sore_throat", compute_binary, jeffreys_binary,
+    "pct_symp_chest_pain", "symp_chest_pain", compute_binary, jeffreys_binary,
+    "pct_symp_nausea", "symp_nausea", compute_binary, jeffreys_binary,
+    "pct_symp_diarrhea", "symp_diarrhea", compute_binary, jeffreys_binary,
+    "pct_anosmia", "symp_loss_smell_taste", compute_binary, jeffreys_binary,
```
capnrefsmmat, line 96: Does this one not match the naming of the others simply because it predates the better naming scheme?

facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -951,3 +1059,51 @@ code_behaviors <- function(input_data, wave) {
...
+  } else {
+    input_data$a_mask_work_outside_home_indoors_1d <- NA
+    input_data$a_mask_shop_indoors_1d <- NA
+    input_data$a_mask_restaurant_indoors_1d <- NA
+    input_data$a_mask_spent_time_indoors_1d <- NA
+    input_data$a_mask_large_event_indoors_1d <- NA
+  }
+  
+  return(input_data)
+}
```
capnrefsmmat, line 1108: I don't think these indicators measure something useful. Consider C13b and C13c on Wave 11. Let's say a respondent says, in C13b, that they went to an indoor market, grocery store, or pharmacy (option 2) in the past 24 hours. C13c will then ask if they wore a mask *during only that activity*, i.e. option 2 will be the only listed option. But they will count as a `FALSE` for whether they wore a mask on public transit or attending an indoor event, despite doing neither.

If we want "what % of people conducting this activity are wearing masks?", which I assume is what Facebook is thinking of, then you'll need to account for what they selected in C13/C13b when evaluating C13a/C13c.

capnrefsmmat (changes requested):
See comment about C13; I think those indicators won't do what we want

facebook/delphiFacebook/R/contingency_indicators.R
nmdefries: That's correct. I can mirror `anosmia` as `symp_loss_smell_taste` (either by running the aggregation procedure twice, or by copying the results over explicitly at some point), or rename it and discontinue `anosmia`. Early on in the contingency tables we renamed some other signals so in the docs those old names are listed as "formerly known as".

facebook/delphiFacebook/R/contingency_variables.R
nmdefries: Thanks, good catch.

facebook/delphiFacebook/R/contingency_indicators.R
capnrefsmmat: I think the right decision here would be to check with UMD on their naming, to ensure consistency; I don't think it matters too much what the exact name is

nmdefries: There's a mix of names on the UMD side. They offer this signal as `anosmia` *and* `symp_loss_smell_taste`, but a related indicator (based on B2c, which symptoms are unusual) only as `anosmia_unusual`. I'll leave ours as `anosmia` for better consistency between all the related signals.

PR #1176, opened 2021-08-06
Use archiver.run function in runner

qx-teo:
### Description
archiver.archive() isn't actually a function, we use run() instead which was implemented here: https://github.com/cmu-delphi/covidcast-indicators/commit/ead53c8af78c833e578967a1a11c53f87d55cfaf#diff-817ffd086856eae7a95f784aee9b38cb94e4ef48e9d9307e5dc6930f93729225

Also fixed tests

### Changelog
- runner.py
- archiver.py

### Fixes 
- changed all calls of archiver.archive() to archiver.run()

### DIFF
--- a/_delphi_utils_python/delphi_utils/runner.py
+++ b/_delphi_utils_python/delphi_utils/runner.py
@@ -45,7 +45,7 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
             name = indicator_fn.__module__,
             filename=params["common"].get("log_filename", None)))
     if archiver and (not validator or validation_report.success()):
-        archiver.archive()
+        archiver.run()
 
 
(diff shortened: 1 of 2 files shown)


krivard:
Additional context for future travelers: this means we've been running quidel_covidtest and google_symptoms all this time without any archive diffing

qx-teo:
Also worth noting - at this point in time there aren't archive params in ansible for quidel_covidtest or google_symptoms, so there hasn't been archiver running anyway - see [this](https://github.com/cmu-delphi/covidcast-indicators/blob/f83033786a912c475e53654a2bc58c2d067e2727/_delphi_utils_python/delphi_utils/archive.py#L132).

krivard (approved):
👍

PR #1177, opened 2021-08-06
Archiver params

qx-teo:
### Description
Adding indicator-prefix to params.json.template as well as ansible so that we can call archiver from params directly.
Haven't previously noticed this issue since only the runner uses the from_params() method of calling the archiver. 
Prior to #1146, all indicators using the runner did not contain archiver params anyway and so the function was not called. The addition of chng archiver params did not cause an issue since there correctly was the indicator_prefix param.

### Changelog
- param.json.template and respective ansible params for USAfacts, JHU and CAN.

### Fixes 
- Added indicator_prefix param, based off what is currently used in individual indicator module run functions.


krivard (approved):
Verified matching with S3 names

PR #1180, opened 2021-08-06
Remove ArchiveDiffer from USAFacts run()

qx-teo:
### Description
ArchiveDiffer to be ran in the runner itself.
Should only be ran in conjunction with changing automation/production commands.

### Changelog
- run.py

### Fixes 
- Remove archiver-related code to consolidate in the runner.


krivard (commented):
👍

PR #1183, opened 2021-08-09
[fb-package] Add feature to produce microdata with race/ethnicity field

nmdefries:
### Description
Allow pipeline to produce microdata with combo race-ethnicity field if params flag is turned on.

### Changelog
- Introduce `produce_individual_raceeth` and `individual_raceeth_dir` params. Set default values in template params.
- Move race/ethnicity coding function to `variables.R`. Inclusion in API run conditional on `produce_individual_raceeth`.
- Add race/ethnicity combo field to list of microdata output fields.
- Light refactor of `write_individual`, adding supporting function `write_individual_day` and step to save raceeth microdata if flag is on.
- Update tests.
- Bonus: make sure template params points to new home of test files and weights in `integration-tests`.


facebook/params.json.template
```diff
@@ -8,9 +8,11 @@
   "cache_dir": "./cache",
   "archive_dir": "./archive",
   "individual_dir": "./individual",
-  "weights_in_dir": "./delphiFacebook/tests/testthat/weights_full",
+  "individual_raceeth_dir": "./individual_raceeth",
+  "produce_individual_raceeth": true,
```
capnrefsmmat, line 12:
```suggestion
  "produce_individual_raceeth": false,
```
default to false to prevent accidents setting up new runs

facebook/delphiFacebook/R/responses.R
```diff
@@ -535,8 +539,8 @@ create_complete_responses <- function(input_data, county_crosswalk)
     "G1", "G2", "G3", "H1", "H2", "H3", "I1", "I2", "I3", "I4", "I5",
     "I6_1", "I6_2", "I6_3", "I6_4", "I6_5", "I6_6", "I6_7", "I6_8",
     "I7", "K1", "K2", "V11a", "V12a", "V15a", "V15b", "V16", "V3a", "module", # added in Wave 11
-
-    "token", "wave", "UserLanguage",
+    
+    "raceethnicity", "token", "wave", "UserLanguage",
```
capnrefsmmat, line 543: This will cause the count of `!is.na()` responses in `filter_complete_responses` to be wrong, since we're adding a new column that's not from Qualtrics; to ensure we continue to filter the same way, we'll need to fix that

(and we should add a test with raceethnicity data that ensures responses with ≥2 answers are kept and those with < 2 are dropped)

capnrefsmmat (changes requested):
One bug and a couple nitpicks, and then I think this should be ready to go

capnrefsmmat (approved):
Looking good

facebook/delphiFacebook/R/variables.R
```diff
@@ -764,3 +764,53 @@ code_beliefs <- function(input_data, wave) {
...
+  if ( "hispanic" %in% names(input_data) &&
+       "race" %in% names(input_data) ) {
+    input_data$raceethnicity <- case_when(
+      input_data$hispanic ~ "Hispanic",
+      !input_data$hispanic & input_data$race == "AmericanIndianAlaskaNative" ~ "NonHispanicAmericanIndianAlaskaNative",
+      !input_data$hispanic & input_data$race == "Asian" ~ "NonHispanicAsian",
+      !input_data$hispanic & input_data$race == "BlackAfricanAmerican" ~ "NonHispanicBlackAfricanAmerican",
+      !input_data$hispanic & input_data$race == "NativeHawaiianPacificIslander" ~ "NonHispanicNativeHawaiianPacificIslander",
+      !input_data$hispanic & input_data$race == "White" ~ "NonHispanicWhite",
+      !input_data$hispanic & input_data$race == "Other" ~ "NonHispanicMultipleOther",
```
capnrefsmmat, line 808:
```suggestion
      !input_data$hispanic & input_data$race == "MultipleOther" ~ "NonHispanicMultipleOther",
```
This prevented the microdata from having any multiracial/other respondents, and also affected all the prior contingency tables, I suspect. I'm currently testing out the fix to ensure it generates the right data

capnrefsmmat (changes requested):
Noticed a bug upon trying to build data. We may need to rebuild past contingency tables with raceethnicity variables to fix this (to include the MultipleOther category when applicable).

capnrefsmmat (approved):
That fix worked. We'll want to merge this soon so you can rerun the old contingency tables with it (ugh :()

PR #1191, opened 2021-08-12
Increase GS amount of data produced per run

### LINKED ISSUE #1079: Update Google Symptoms num_export_days logic
Currently GS does the following

```
    if num_export_days is None:
        # Get number of days based on what's missing from the API.
        metadata = covidcast.metadata()
        gs_metadata = metadata[(metadata.data_source == "google-symptoms")]
        num_export_days = max(gs_metadata.min_lag)
```

 which may lead to problem if we haven't made an issue in a long time (but those issues had low lags). Changing it to something like `datetime.today() - min(gs_metadata.max_time)` should get around this.

### LINKED ISSUE #1190: Increase google-symptoms date range for pulling data
**Description**
Google-symptoms currently only considers the maximum of the min-lags observed when doing pulls from BigQuery, this unfortunately results in 0-2 days of data for smoothed signals and 6-8 days of data for raw signals. This leads to a lot of validator problems, namely check_missing_dates, check_missing_geo_sig, and check_missing_geo_sig_date.

**Solution**
Increase the pull range to generate (span_length) number of days for the validator, which would mean (span_length+6) for raw signals.

Other solutions discussed include rewriting the validator or suppressing errors, but these were deemed less elegant.

nmdefries:
### Description
Increase min amount of data pulled to improve compatibility with validator. Improve robustness to data source outages.

### Changelog
- Use min(max_time) instead of max(min_lag) in `num_export_days` default definition to prevent skipped dates in the case that GS hasn't issued data in a long time but the most recent issues had low lags.
- Produce at least enough days of data by default to prevent validator from complaining about missing dates. Uses `span_length` validation setting so stays up-to-date with validation.

### Fixes 
- Closes #1079
- Closes #1190

### DIFF
--- a/google_symptoms/delphi_google_symptoms/run.py
+++ b/google_symptoms/delphi_google_symptoms/run.py
@@ -64,13 +64,12 @@ def run_module(params):
         # Calculate number of days based on what validator expects.
         max_expected_lag = params["validation"]["common"].get("max_expected_lag", {"default": 4})
         global_max_expected_lag = max(map(int, list(max_expected_lag.values()) ))
-        span_length = params["validation"]["common"].get("span_length", 14)
 
         # Select the larger number of days. Prevents validator from complaining about missing dates
         # and backfills in case of an outage.
         num_export_days = max(
             (datetime.today() - to_datetime(min(gs_metadata.max_time))).days + 1,
-            span_length + global_max_expected_lag
+            params["validation"]["common"].get("span_length", 14) + global_max_expected_lag
             )
 
     logger = get_structured_logger(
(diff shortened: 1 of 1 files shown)


qx-teo:
Might need to use lag_converter() from utils on max_expected_lag, in case max lag comes in the form of "sundays" -
But since this is only for GS, and it doesn't update weekly, it should be fine?

nmdefries:
> But since this is only for GS, and it doesn't update weekly / use the sundays format, it should be fine?

Yeah, this is what I was thinking, but it's also easy enough to add `lag_converter` in. I'll take a look tomorrow.

qx-teo (approved):
lgtm!

PR #1192, opened 2021-08-13
Tell sirCAL to ignore full and parttime schooling survey indicators

nmdefries:
### Description
The item backing the schooling indicators in CTIS has been removed for the summer (aiming to add it back in Wave 12). Tell sirCAL to ignore.


krivard (approved):
👍

PR #1202, opened 2021-08-17
Configure survey indicator for deployment

korlaxxalrok:
### Description
This preps the survey indicator to be deployed to our runtime system. We will
need the indicator skeleton to be in place when we run the containerized version.

### Changelog
- Adds the indicator the Jenkins processing list
- Adds a production params template file
- Adds encrypted secrets and vars file pointers


krivard (approved):
👍

PR #1205, opened 2021-08-17
Fix: Make the facebook indicator actually build correctly

korlaxxalrok:
### Description
Fix

I forgot to include a basic setup.py which caused Jenkins to fail. 

### Changelog
- Adds setup.py

### DIFF
--- a/facebook/setup.py
+++ b/facebook/setup.py
@@ -1,3 +1,7 @@
+# this file is used only as a hookup for Jenkins. 
+# if you are looking for the setup.py file associated with the 
+# delphi_facebook python package, see `delphiFacebook/python/setup.py`
+
 from setuptools import setup
 from setuptools import find_packages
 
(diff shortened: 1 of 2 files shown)


facebook/setup.py
```diff
@@ -0,0 +1,20 @@
+from setuptools import setup
```
krivard, line 1:
```suggestion
# this file is used only as a hookup for Jenkins. 
# if you are looking for the setup.py file associated with the 
# delphi_facebook python package, see `delphiFacebook/python/setup.py`

from setuptools import setup
```

krivard (changes requested):
Something munged up your commit record -- to fix that conflict and avoid adding duplicate commits to main, try one of these:

* `git rebase main`
* `git rebase -i main` & drop those first two from QX
* branch from main, cherry-pick the two commits you want, drop the bad branch, rename the good branch

& force-push. If you wind up unrecoverably mired, lmk and I can poke it tomorrow.

korlaxxalrok:
@krivard Ah, oops. I had rebased my branch to fixup a couple of test commits and I think I went a bit too far back in the history and included QX's unintentionally. Appreciate the 👀 

I think I have fixed it now, and have also accepted your changes 👍

krivard (approved):
👍

PR #1212, opened 2021-08-19
Switch to general-purpose release workflows

krivard:
### Description
Two new workflows for general-purpose releases. These auto detect when a separate delphi-utils release is needed, and note in the PR text what is being done. Once merged, auto-detect again to determine whether to upload a new version of delphi-utils to PyPI, and tag accordingly.

This will mean we double up on tags whenever we need to do a delphi-utils release, but I'm not super bothered by that.

These have been crudely tested in [a fork](https://github.com/krivard/covidcast-indicators/actions).

### Changelog
- new on:workflow-dispatch workflow: create-release.yml
- new on:push workflow: publish-release.yml
- new bumpversion config to cover the entire indicators repo
- Annotations to old workflows so we keep them around just in case but make it easy to know which ones to use

### Fixes 
- Addresses confusion from Katie's vacation last week


.github/workflows/create-release.yml
```diff
@@ -0,0 +1,73 @@
...
+        id: utils-changed
+        if: steps.changes.outputs.utils == 'true'
+        working-directory: ./_delphi_utils_python
+        run: |
+          echo -n "::set-output name=version::"
+          bump2version --list ${{ github.event.inputs.versionName }} | grep ^new_version | sed -r s,"^.*=",,
+          echo -e "\n::set-output name=msg::(*new*)"
+      - name: Detect delphi-utils version
+        id: utils-unchanged
+        if: steps.changes.outputs.utils == 'false'
```
sgratzl, line 51: isn't there a working-directory missing?

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,123 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install bump2version
+        run: python -m pip install bump2version
+      - name: Check for delphi-utils changes
+        uses: dorny/paths-filter@v2
+        id: changes
+        with:
+          base: 'prod'
```
sgratzl, line 34: I don't understand base and ref are the same. how can be there changes?

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,73 @@
+name: Create Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      versionName:
```
sgratzl, line 6: so this version might be used for both delphi_utils and the indicators? if I specify patch then indicators will be patched in any case and delphi_utils might

.github/workflows/publish-release.yml
krivard: [From the action README](https://github.com/dorny/paths-filter): When `base` and `ref` are the same, changes are detected against the most recent commit before the push. I tried using `prod` and `prod~1` first, but that doesn't work

image

.github/workflows/create-release.yml
krivard: That's correct. The indicators version helps us track when particular changes went into production, and that includes changes to utils.

Intended use is to deploy whenever a change either in utils or indicators needs to go out to prod. Initial version selection (patch/minor/etc) would be made based on that change. If we're trying to put out a patch change to indicators but it turns out there's a change to utils as well, the PR text alerts the change. If the utils change is also a patch, fine; if it merits a minor version bump, then the PR can be closed and the workflow re-run as minor.

.github/workflows/create-release.yml
krivard, line 51:
```suggestion
        if: steps.changes.outputs.utils == 'false'
        working-directory: ./_delphi_utils_python
```

.github/workflows/publish-release.yml
krivard, line 34:
```suggestion
          # when base and ref are the same, changes are detected against the most recent commit before the push.
          # see https://github.com/dorny/paths-filter#usage for details.
          base: 'prod'
```

sgratzl (approved):
👍

PR #1214, opened 2021-08-20
Validator doc

qx-teo:
### Description
Fixed bug about delphi_validator module name - since validator has been shifted under utils
Added new params(dry_run)
Updated new params (min/max expected lag, end date, ref window size)
Removed unnecessary pip install since validator is now housed under delphi_utils
Updated PLANS (may need second set of eyes)

### Changelog
- PLANS.md
- README.md

### Fixes 
- Fixes #(issue)


_delphi_utils_python/delphi_utils/validator/README.md
```diff
@@ -53,7 +52,10 @@ Please update the follow settings:
...
+   * `max_expected_lag` (default: 10 for all unspecified signals): dictionary of signal name-string pairs specifying the maximum number of days of expected lag (time between event occurrence and when data about that event was published) for that signal. Values are either numeric or `sunday+{num},{num}`. `sunday+{num},{num}` is used for indicators that update data on a regular weekly basis. The first number denotes the day of the week (with Monday = 1 and so on, and Sunday taking either 0 or 7). The second number denotes the expected lag for the data upon upload date. In other words, if the signal updates weekly on Wednesday's for the past week's data up til Sunday, the correct parameter would be something like `sunday+3,3`.
+   * `min_expected_lag` (default: 1 for all unspecified signals): dictionary of signal name-string pairs specifying the minimum number of days of expected lag (time between event occurrence and when data about that event was published) for that signal. See `max_expected_lag` for further details.
```
nmdefries, line 58: For these two, you should describe how to set it globally as well.

_delphi_utils_python/delphi_utils/validator/README.md
```diff
@@ -53,7 +51,10 @@ Please update the follow settings:
 
 * `common`: global validation settings
    * `data_source`: should match the [formatting](https://cmu-delphi.github.io/delphi-epidata/api/covidcast_signals.html) as used in COVIDcast API calls
-   * `end_date`: specifies the last date to be checked; this can be specified as `YYYY-MM-DD`, `today`, or `today-{num}`.  The latter is interpretted as `num` days before the current date.
+   * `dry_run`: boolean; `true` prevent system exit with error and ensures that the success() method of the ValidationReport always returns true.
+   * `end_date`: specifies the last date to be checked; this can be specified as `YYYY-MM-DD`, `today`, `today-{num}`, or `sunday+{num}`.  `today-num` is interpretted as `num` days before the current date. `sunday+{num}` sets the end date as the most recent day of the week prior to today (as specified by the user). The numeric input represents the day of the week (`sunday+1` denotes Monday, and so on, with 0 or 7 both indicating Sunday). Defaults to the global minimum of `min_expected_lag` days behind today.
```
nmdefries, line 55: spelling

```suggestion
   * `end_date`: specifies the last date to be checked; this can be specified as `YYYY-MM-DD`, `today`, `today-{num}`, or `sunday+{num}`.  `today-num` is interpreted as `num` days before the current date. `sunday+{num}` sets the end date as the most recent day of the week prior to today (as specified by the user). The numeric input represents the day of the week (`sunday+1` denotes Monday, and so on, with 0 or 7 both indicating Sunday). Defaults to the global minimum of `min_expected_lag` days behind today.
```

nmdefries:
I removed from the plans doc some more feature requests that I think you've dealt with. You should (formally via review or informally) check that those changes are reasonable.

We could add something to the plans doc about using individual z-score thresholds instead of hardcoded ones in `check_avg_val_vs_reference`. Not necessarily for all signals, but at least for those we suppress because they aren't described well by the current hardcoded thresholds (e.g. monotonic signals, like `pcr_positivity_rate`). Also useful because we don't know how much signal behavior varies over long time periods (multiple months-years). Your proposed approach was "caching past z-scores and somehow using that to get signal specific thresholds", so say calculate 95%ile on the fly using past data.

Smaller TODOs can go on here to, any needs that have come up that you're planning on working on.

nmdefries (approved):
Looks good!

PR #1223, opened 2021-08-24
Make convex solver more robust for weekday correction

rumackaaron:
### Description
For the doctor visits indicator, we use cvxpy to solve an optimization problem that provides a weekday correction. This solver is not always robust, occasionally it will fail if the magnitude of the objective function is too large. A quick and dirty way to make it more robust is to divide the objective function by some large number.

### Changelog
`weekday.py`: Try up to four times to rescale the objective function, up from two times.

### Fixes 
- No issue linked.

### DIFF
--- a/doctor_visits/delphi_doctor_visits/weekday.py
+++ b/doctor_visits/delphi_doctor_visits/weekday.py
@@ -82,14 +82,16 @@ class Weekday:
             penalty = (
                 lmbda * cp.norm(cp.diff(b[6:], 3), 1) / (X.shape[0] - 2)
             )  # L-1 Norm of third differences, rewards smoothness
-            try:
-                prob = cp.Problem(cp.Minimize(-ll + lmbda * penalty))
-                _ = prob.solve()
-            except:
-                # If the magnitude of the objective function is too large, an error is
-                # thrown; Rescale the objective function
-                prob = cp.Problem(cp.Minimize((-ll + lmbda * penalty) / 1e5))
-                _ = prob.solve()
+            scales = [1, 1e5, 1e10, 1e15]
+            for scale in scales:
+                try:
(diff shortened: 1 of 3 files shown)


doctor_visits/delphi_doctor_visits/weekday.py
```diff
@@ -82,14 +82,16 @@ def get_params(data):
...
+            scales = [1, 1e5, 1e10, 1e15]
+            for scale in scales:
+                try:
+                    prob = cp.Problem(cp.Minimize((-ll + lmbda * penalty) / scale))
+                    _ = prob.solve()
+                    break
+                except:
+                    # If the magnitude of the objective function is too large, an error is
+                    # thrown; Rescale the objective function by going through loop
+                    pass
```
krivard, line 94: Consider an `else` block here to throw a useful error if none of the scale values work.

[more info on for/else](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops)

rumackaaron: I'm not sure what the ideal behavior is. It's now set up that if the solver fails, the indicator code won't crash, but unless someone is looking through the error logs, this is a pretty silent error.

krivard: it should either crash or skip outputting the files for the adjusted sensors. At least that way we'll get a sirCAL alert.

rumackaaron: How does it look now?

krivard (approved):
Great! @korlaxxalrok can you use your cosmic powers to figure out how we can update the code in Maria's directory on bigchunk-dev-03?

PR #1226, opened 2021-08-26
Remove US territories from valid zips list

nmdefries:
### Description
Zip codes belonging to unweighted US territories are dropped from the list of valid zip codes. As a result, we will not include survey respondents reporting that they live in one of these non-valid zip codes in CID lists, microdata, or aggregate calculations.

### Changelog
- Drop entries in `zip_metadata` in `geo.R::produce_zip_metadata` corresponding to US territories (Puerto Rico, Guam, American Samoa, the Virgin Islands, and the Northern Mariana Islands).

### Fixes 
FB does not advertise our survey to users outside of the US proper. If a respondent is listed on FB as living in the US, but in the survey say they are currently living in a US territory, the weights assigned to them are incorrect since weights are based on demographics of their FB-reported location.

### DIFF
--- a/facebook/delphiFacebook/R/geo.R
+++ b/facebook/delphiFacebook/R/geo.R
@@ -1,9 +1,12 @@
 #' Returns metadata about each U.S. Zip Code
+#' 
+#' Drops information about US territories, which FB does not provide weights for.
 #'
 #' @param static_dir     local directory containing the file "02_20_uszips.csv"
 #'
 #' @importFrom stringi stri_trans_tolower
 #' @importFrom readr read_csv cols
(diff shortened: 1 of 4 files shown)


capnrefsmmat:
Our sample size in Puerto Rico is already too low to even provide CLI estimates in the API, so are we actually using it in any of our models?

capnrefsmmat:
Specifically, the last day we have API estimates for Puerto Rico was August 12, 2020, so over a year ago.

capnrefsmmat (approved):

krivard:
Whoa, I had no idea. Objection retracted!

PR #1227, opened 2021-09-01
Runtime improvements to survey pipelines

nmdefries:
### Description
Runtime improvements to survey API and contingency pipelines. Minor memory improvements (unmeasured).

### Changelog
- Simplify contingency pipeline.
    - Drop support for custom aggs (#977). No longer doing custom aggs.
    - Drop support for running weekly and monthly pipelines at the same time.
    - Drop derived field processing based on multiple choice name prefix "mc_". No longer using that naming convention.
- Make `is_selected` faster. Implemented C++ version and faster R version that operates on only unique answer sets. Create/update a lot of associated files to support use of C++ code.
- Switch `summarize_indicators_day` to store initial output in list of lists instead of tibble/dataframe to avoid copy-on-modify inefficiency. Only generate indicator list once. Create `dfs_out` in single step instead of modifying in a loop.
- Switch `summarize_aggregations_group` to only generate indicator tibble once and create `dfs_out` in single step instead of modifying in a loop.
- Read weight files in with `fread`.

### Result
API pipeline is faster: 24 min vs 32 min single threaded, generating all signals for a single day from a late-June Wave 11 input file (on my personal computer, so actual times here won't transfer).

Contingency pipeline is faster: 1h45m vs 4h25m generating tables for the week of Aug 22.


facebook/delphiFacebook/R/variables.R
```diff
@@ -40,11 +60,14 @@ is_selected <- function(vec, selection) {
         # selection ("") or missing (NA) as missing, for generality.
         NA
       } else {
-        selection %in% resp
+        any(resp == selection)
```
capnrefsmmat, line 63: I see you've replaced `%in%` in two spots; is this known to be faster for some reason?

facebook/delphiFacebook/src/is_selected_cpp.cpp
```diff
@@ -0,0 +1,32 @@
...
+      out[i] = NA_LOGICAL;
+      continue;
+    }
+
+    for (int j = 0; j < response.size(); ++j ) {
+      if(StringVector::is_na(response[j])) {
+        out[i] = NA_LOGICAL;
+        break;
+      }
+      if(response[j] == target) {
```
capnrefsmmat, line 25:
```suggestion
      if (StringVector::is_na(response[j])) {
        out[i] = NA_LOGICAL;
        break;
      }
      if (response[j] == target) {
```
style nitpick

facebook/contingency_tables.R
```diff
@@ -1,4 +1,6 @@
 library(delphiFacebook)
 
+#Rprof(interval = 0.5)
```
capnrefsmmat, line 3: leftover from testing?

capnrefsmmat (commented):
A couple minor nitpicks, but it looks good. But we probably shouldn't commit `is_selected_cpp.o`, right? Because that's platform-dependent and you want Rcpp to compile it on package install on each machine

facebook/delphiFacebook/R/variables.R
capnrefsmmat: I think `%in%` calls `match`:

```r
> `%in%`
function (x, table) 
match(x, table, nomatch = 0L) > 0L
```

Perhaps the extra bookkeeping done by `match` to find the index of the first occurrence? Dunno

capnrefsmmat (approved):

capnrefsmmat:
also I should add that "1h45m vs 4h25m generating tables" is super nice; excellent optimization result

PR #1231, opened 2021-09-03
Generalize weights file read procedure

nmdefries:
### Description
Make weights file read procedure more explicit to work with small number of weights files that are missing headers.

### Changelog
- Specify field types by position, not name.
- Name fields by position on read.

### Fixes
Because the new weight files read procedure (#1227) specifies field types by name, it revealed two weights files that don't use those names. `2020-05-04_covid19_dap_adults_finish_step_1_weights.csv` and `2020-05-05_covid19_dap_adults_finish_step_1_weights.csv` don't use the "cid" and "weight" names or have any headers (presumably a bug in the early FB weights pipeline).

The pre-`fread` implementation of the weights file read procedure didn't specify names when setting field types, so they would have been assigned default names (X1, X2 in `read_csv`) on read. `bind_rows` thus would have created a dataframe containing four columns, "cid", "weight", "X1", and "X2", with "X1" and "X2" entirely ignored in later steps, e.g. joining onto the survey data.

### Implications
Due to the missing weights, data products from these dates were not generated correctly. For example, nation-level `raw_wcli` has no observations for 2020-05-04 or 2020-05-05 despite high sample size (~70k) on surrounding days (see `covidcast_signal("fb-survey", "raw_wcli", geo_type="nation", start_day = "2020-04-20", end_day="2020-05-20")`). We will need to reissue data products that include these dates.

### DIFF
--- a/facebook/delphiFacebook/R/weights.R
+++ b/facebook/delphiFacebook/R/weights.R
@@ -61,10 +61,15 @@ join_weights <- function(data, params, weights = c("step1", "full"))
 
   weights_files <- dir(params$weights_in_dir, pattern = pattern, full.names = TRUE)
   weights_files <- sort(weights_files)
+  
+  col_types <- c("character", "double")
+  col_names <- c("cid", "weight")
   agg_weights <- bind_rows(lapply(
     weights_files,
     fread,
-    colClasses = c(cid="character", weight="double"))
+    colClasses = col_types,
+    col.names = col_names
+    )
   )
   agg_weights <- agg_weights[!duplicated(cid),]
   data <- left_join(data, agg_weights, by = c("token" = "cid"))


capnrefsmmat:
Interesting bug, and good catch. Will the new code issue a warning or error if a data file comes in with different column names or types than expected, e.g. if something goes wrong with the weights pipeline in the future and we get a corrupted file? It'd be good to detect that early so we don't have missing days again.

nmdefries:
So the old code `fread(filepath, colClasses = c(cid="character", weight="double"))` is stricter. It raises a warning if the named columns aren't in the file, and it reads in whatever columns there are with names from the header or, if there is no header, auto-assigned names (V1, V2, etc) (which we then ignore).

The new code isn't as strict; it just names whatever columns exist by position with no particular requirement for name-type mappings, like the previous `read_csv` procedure.

On types in `fread`, "[i]f type coercion results in an error, introduces NAs, or would result in loss of accuracy, the coercion attempt is aborted for that column with warning and the column's type is left unchanged" ([`fread` documentation](https://www.rdocumentation.org/packages/data.table/versions/1.14.0/topics/fread)).

I can switch this back to the old `fread` code and add in a `tryCatch` to fail for any failed type coercion or missing column warnings. Any weight files that have nonstandard format will have to be fixed or, if we decide the contents are valid despite the formatting, like these two, read in with special format settings.

capnrefsmmat:
Hmm, so we'd at least get a warning if, say, they swapped the order of the columns. That's good.

I wonder if ultimately the right thing to do would be to add a separate check to see whether we've successfully loaded weights for all days in an interval, or whether there are gaps in the weights indicating a missing file. Is that plausible to check?

nmdefries:
When we load the weights files, it doesn't include any date info. We would have to check for dates missing any weights after the weights are joined on to the data -- is that what you were imagining? Or we could add weights filenames or dates based on filenames to the weights df and check before joining.

I was leaning towards checking format, etc, up front so that we know the exact mapping between each file and set of weights. Maybe when we load each weights file or using a filename/filename-derived date field.

capnrefsmmat:
Hmm, okay. So I think this PR is definitely a good step and we should go ahead with it. It sounds like auto-detecting missing weights on specific days is a bit harder because the data files themselves don't include dates; the dates are only by convention in the file name, and if we're trying to be robust to files delivered with wrong names (like sparked this PR) then relying in on the file name is out. So we'd have to do some check to the final data frame, except that's not perfect either because there will always be some submissions (by people following a wrong link or whatever) without any weights, so we'd need some heuristic.

Possibly a more generic approach would be to simply issue a warning if there's a day with low sample size in the middle of the output range, because that suggests problems with the weights, Qualtrics input files, or something else entirely.

capnrefsmmat (approved):

PR #1228, opened 2021-09-03
Geomapper docstring and error cleanup

alexcoda:
### Description
Update out-of-date docstrings, some formatting, and error handling. Contains no changes to functionality.

### Changelog
- Update inaccurate docstring parameters.
- Simplify logic for error handling in `get_geos_within` method.
- Minor formatting changes

PR #1232, opened 2021-09-07
Pre load geomap data files

### LINKED ISSUE #282: Refactor the geocode utility to use a disk caching utility
We load crosswalk files on demand in the geocode utility because they can be large and slow to load. This functionality can likely be outsourced to a memoize-to-disk library like [joblib](https://joblib.readthedocs.io/en/latest/memory.html) and remove code complexity, such as manually tracking the filenames for the crosswalk files. 

This functionality can simplify extensions to fast crosswalks: for example, to convert from jhu_uid to zip, we currently need to crosswalk jhu_uid to fips and then fips to zip, whereas we could store the jhu_uid to zip mapping directly after computing it once.

alexcoda:
### Description
Update the Geomapper class to pre-load the crosswalk data on initialization.

This change was suggested by @chinandrew. It closes #282 by loading all of the data when the class is initialized. This loading is quick enough to not significantly slow down class creation:
37556AD6-3553-4F7A-A268-61A56053D314

### Changelog
- Change crosswalk and geo value sets to load when the class is initialized
- Add a public method `get_crosswalk` for directly accessing the crosswalk tables. This method is now used within combo_cases, quidel, and google_symptoms.
- Use a fixture in the tests to prevent from loading the data for each individual test

Note that this PR is branched off of #1228, to see just the diff for this PR, click [here](https://github.com/cmu-delphi/covidcast-indicators/pull/1232/commits/6a4480530e4b21e9af4961ff89e8f29fb3b3d9db).

### DIFF
--- a/_delphi_utils_python/delphi_utils/geomap.py
+++ b/_delphi_utils_python/delphi_utils/geomap.py
@@ -44,8 +44,8 @@ CROSSWALK_FILEPATHS = {
         "pop": join(DATA_PATH, "state_pop.csv")
     },
     "jhu_uid": {"fips": join(DATA_PATH, "jhu_uid_fips_table.csv")},
-    "hhs": {"pop": join(DATA_PATH, "hhs_pop.csv"),},
-    "nation": {"pop": join(DATA_PATH, "nation_pop.csv"),},
+    "hhs": {"pop": join(DATA_PATH, "hhs_pop.csv")},
+    "nation": {"pop": join(DATA_PATH, "nation_pop.csv")},
 }
 
 
@@ -195,10 +195,10 @@ class GeoMapper:  # pylint: disable=too-many-public-methods
             }
         elif (from_code, to_code) == ("fips", "state"):
             dtype = {
-                    "fips": str,
-                    "state_code": str,
-                    "state_id": str,
-                    "state_name": str,
+                "fips": str,
+                "state_code": str,
+                "state_id": str,
+                "state_name": str,
             }
 
         # Population tables
@@ -556,6 +556,7 @@ class GeoMapper:  # pylint: disable=too-many-public-methods
         if geo_type == "county":
             return "fips"
         return geo_type
+
     def get_geo_values(self, geo_type):
         """
         Return a set of all values for a given geography type.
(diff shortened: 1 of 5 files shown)


chinandrew:
For additional context, our thinking was that this could lower complexity for public functions and make geomapper stateless.

dshemetov:
It takes 185ms to load all the crosswalk files? That's pretty fast!

krivard:
tagging @dshemetov to spread the knowledge around

krivard (approved):
lgtm. when we deploy this we'll need to put some extra eyes on the following affected pipelines:
* google_symptoms
* quidel_covidtest
* combo cases/deaths

_delphi_utils_python/delphi_utils/geomap.py
```diff
@@ -124,29 +120,21 @@ def __init__(self):
...
-            f"No crosswalk files for {from_code}; try {'; '.join(self.crosswalk_filepaths.keys())}"
-        assert to_code in self.crosswalk_filepaths[from_code], \
-            f"No crosswalk file from {from_code} to {to_code}; try " \
-            f"{'; '.join(self.crosswalk_filepaths[from_code].keys())}"
-
-        if self.crosswalks[from_code][to_code] is None:
-            self.crosswalks[from_code][to_code] = self._load_crosswalk_from_file(from_code, to_code)
-        return self.crosswalks[from_code][to_code]
-
-    def _load_crosswalk_from_file(self, from_code, to_code):
```
dshemetov, line 151: While we're refactoring this utility for clarity, I think we can get rid of most of the logic in this function by writing 
```
def _load_crosswalk_from_file(self, from_code, to_code, data_path):
    stream = pkg_resources.resource_stream(__name__, data_path)
    dtype = {
        from_code: str,
        to_code: str,
        "fips": str,
        "zip": str,
        "hrr": str,
        "hhs": str,
        "msa": str,
        "state_code": str,
        "state_id": str,
        "state_name": str,
        "pop": int,
        "weight": float
    }
    usecols = [from_code, "pop"] if to_code == "pop" else None
    return pd.read_csv(stream, dtype=dtype, usecols=usecols)
```
since `pd.read_csv` will just ignore irrelevant columns and there's no clash in the dtypes otherwise.

dshemetov (changes requested):
One suggestion, otherwise lgtm!

dshemetov (approved):
🚢 !

PR #1233, opened 2021-09-07
sirCAL to ignore vaccine_barrier_*_tried for MSA

nmdefries:
### Description
sirCAL to ignore `vaccine_barrier_*_tried` indicators at the MSA level.

### Changelog
- Add vaccine barriers to sirCAL production template using new geo-specific format.


krivard (approved):
👍

PR #1236, opened 2021-09-07
Use recoded values for response choice codes in survey codebook

### LINKED ISSUE #1206: Apparent miscoding of B10b answer choice in survey codebook
As reported by @jacobbien:

> I think I've come across a mistake in the codebook with question B10b (my favorite question... reason for being tested).  At wave 11, choice 5 changes from being about attending a large event to being about international travel.  Now, my first thought was that it was too bad that this wasn't coded with a new number since this is more than just a light edit to wording.  However, later on I noticed that no one has ever chosen option 5 in wave 11.  Looking back, I now see that there is an option 9 in the survey response data and yet in the codebook the choices end at option 8.  So my guess is that the codebook has the following error (but would like confirmation):
>
> For wave 11, question B10b, in the choices column, the 5 should be a 9.  That is, the international travel reason should be option 9.
>
> I just confirmed my hunch by looking at the pdf:
Screen Shot 2021-08-17 at 3 22 10 PM

nmdefries:
### Description
Use recoded values for response choice codes in survey codebook. Qualtrics auto-assigns response codes for all questions. For a subset, we assign our own codes that override the Qualtrics ones. Our codes are used in the survey PDF, response data, and pipeline so we'd like the codebook to match.

### Changelog
- Change names (integer codes, later keys in the JSON format) of answer choices to those specified in `RecodeValues` section of QSFs.

### Fixes 
Fixes unexpected answer choice coding that disagrees with survey PDF. Closes #1206.

### DIFF
--- a/facebook/qsf-tools/generate-codebook.R
+++ b/facebook/qsf-tools/generate-codebook.R
@@ -169,11 +169,25 @@ process_qsf <- function(path_to_qsf,
       map(.x, "Operator"),
       map(.x, "RightOperand")
     )) %>% 
+    # Remove empty logic
     map(~ gsub("  Is  ", "", .x)) %>% 
     map(~ gsub("NULL NULL Is NULL NULL", "", .x)) %>%
     map(~ gsub(" ?NULL ?", "", .x)) %>%
+    # Remove QID flag
     map(~ gsub("q://", "", .x)) %>%
+    # Recode choice numbers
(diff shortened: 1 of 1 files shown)


capnrefsmmat (approved):
Looks fine, presuming you've tried it on the QSFs we've had. Can you regenerate the public codebook so I can update it as well?

PR #1238, opened 2021-09-08
First pass of the CDC Vaccination Indicator

Ananya-Joshi:
### Description
First pass on the CDC Indicator with tests


Ananya-Joshi:
Missing HI and TX data by county.

cdc_vaccines/delphi_cdc_vaccines/pull.py
```diff
@@ -0,0 +1,162 @@
...
+    df.columns = ["fips",
+                  "cumulative_counts_tot_vaccine",
+                  "cumulative_counts_tot_vaccine_12P",
+                  "cumulative_counts_tot_vaccine_18P",
+                  "cumulative_counts_tot_vaccine_65P",
+                  "cumulative_counts_part_vaccine",
+                  "cumulative_counts_part_vaccine_12P",
+                  "cumulative_counts_part_vaccine_18P",
+                  "cumulative_counts_part_vaccine_65P",
+                  "timestamp"]
```
krivard, line 95: this longhand can stay; all the ways I can think of to abstract it out are worse

cdc_vaccines/delphi_cdc_vaccines/pull.py
```diff
@@ -0,0 +1,162 @@
...
+    df["counts_part_vaccine"] = df["cumulative_counts_part_vaccine"].diff()
+    df["counts_part_vaccine_12P"] = df["cumulative_counts_part_vaccine_12P"].diff()
+    df["counts_part_vaccine_18P"] = df["cumulative_counts_part_vaccine_18P"].diff()
+    df["counts_part_vaccine_65P"] = df["cumulative_counts_part_vaccine_65P"].diff()
+
+    rem_list = [ x for x in list(df.columns) if x not in ['timestamp', 'fips'] ]
+    # Handle edge cases where we diffed across fips
+    mask = df["fips"] != df["fips"].shift(1)
+    df.loc[mask, rem_list] = np.nan
+    print(rem_list)
```
krivard, line 128: was this for debugging or should it be logged?

cdc_vaccines/delphi_cdc_vaccines/pull.py
```diff
@@ -0,0 +1,162 @@
...
+    # Obtain new_counts
+    df.sort_values(["fips", "timestamp"], inplace=True)
+    df["counts_tot_vaccine"] = df["cumulative_counts_tot_vaccine"].diff()  # 1st discrete difference
+    df["counts_tot_vaccine_12P"] = df["cumulative_counts_tot_vaccine_12P"].diff()
+    df["counts_tot_vaccine_18P"] = df["cumulative_counts_tot_vaccine_18P"].diff()
+    df["counts_tot_vaccine_65P"] = df["cumulative_counts_tot_vaccine_65P"].diff()
+    df["counts_part_vaccine"] = df["cumulative_counts_part_vaccine"].diff()
+    df["counts_part_vaccine_12P"] = df["cumulative_counts_part_vaccine_12P"].diff()
+    df["counts_part_vaccine_18P"] = df["cumulative_counts_part_vaccine_18P"].diff()
+    df["counts_part_vaccine_65P"] = df["cumulative_counts_part_vaccine_65P"].diff()
```
krivard, line 122: leverage regularity here, eg add to constants:
```
DIFFERENCE_MAPPING = {
    f"{INCIDENCE}_counts_{completeness}_vaccine{age}": f"{CUMULATIVE}_counts_{completeness}_vaccine{age}"
    for completeness, age in product(COMPLETENESS, AGE)
}
SIGNALS = list(DIFFERENCE_MAPPING.keys()) + list(DIFFERENCE_MAPPING.values())
```

then this section becomes
```
for to, from in DIFFERENCE_MAPPING.items():
    df[to] = df[from].diff()
```

cdc_vaccines/delphi_cdc_vaccines/pull.py
```diff
@@ -0,0 +1,162 @@
...
+            "counts_tot_vaccine_65P",
+            "cumulative_counts_part_vaccine",
+            "counts_part_vaccine",
+            "cumulative_counts_part_vaccine_12P",
+            "counts_part_vaccine_12P",
+            "cumulative_counts_part_vaccine_18P",
+            "counts_part_vaccine_18P",
+            "cumulative_counts_part_vaccine_65P",
+            "counts_part_vaccine_65P"
+        ],
```
krivard, line 161: would this work?
```suggestion
        # Reorder
        ["fips", "timestamp"] + SIGNALS,
```

cdc_vaccines/delphi_cdc_vaccines/run.py
```diff
@@ -0,0 +1,91 @@
...
+    if not os.path.exists(params["common"]["export_dir"]):
+        os.makedirs(params["common"]["export_dir"])
+
+    for (sensor, smoother, geo) in product(SIGNALS, SMOOTHERS, GEOS):
+        logger.info("Running on ",
+            sensor=sensor,
+            smoother=smoother,
+            geo=geo)
+        df = GeoMapper().replace_geocode(
+            all_data[['timestamp','fips', sensor]],from_col='fips',
```
krivard, line 58: style nit
```suggestion
            all_data[['timestamp','fips', sensor]],
            from_col='fips',
```

cdc_vaccines/delphi_cdc_vaccines/constants.py
```diff
@@ -0,0 +1,54 @@
...
+	PART_VAC_12P     ,
+	PART_VAC_SUM_18P ,
+	PART_VAC_18P     ,
+	PART_VAC_SUM_65P ,
+	PART_VAC_65P
+]
+
+GEOS = [
+    "nation",
+    "state_code",
```
krivard, line 46: if this is just being used as a geomapper input, it's fine, but if it is also getting used as an input to `export_csvs` then it has to be `state` or the API will reject the file. I think we typically use `state` here in the constants file and conditionally use `state_code` only when we pass it to geomapper.

cdc_vaccines/README.md
```diff
@@ -0,0 +1,70 @@
+# CDC Vaccinations
+
+This indicator provides the number official vaccinations in the US. We export the county-level 
+daily vaccination rates data as-is, and publishes the result as a COVIDcast signal. 
+We also aggregate the data to the HHS, State, and Nation levels.
```
krivard, line 5:
```suggestion
We also aggregate the data to the MSA, HRR, State, HHS Region, and Nation levels. 
```

cdc_vaccines/README.md
```diff
@@ -0,0 +1,70 @@
+# CDC Vaccinations
+
+This indicator provides the number official vaccinations in the US. We export the county-level 
+daily vaccination rates data as-is, and publishes the result as a COVIDcast signal.
```
krivard, line 4:
```suggestion
daily vaccination rates data as-is, and publish the result as a COVIDcast signal. 
```

cdc_vaccines/README.md
```diff
@@ -0,0 +1,70 @@
+# CDC Vaccinations
+
+This indicator provides the number official vaccinations in the US. We export the county-level
```
krivard, line 3:
```suggestion
This indicator provides the official vaccination counts in the US. We export the county-level 
```

cdc_vaccines/README.md
```diff
@@ -0,0 +1,70 @@
+# CDC Vaccinations
+
+This indicator provides the number official vaccinations in the US. We export the county-level 
+daily vaccination rates data as-is, and publishes the result as a COVIDcast signal. 
+We also aggregate the data to the HHS, State, and Nation levels. 
+For detailed information see the files DETAILS.md contained in this directory.
+
+Note that individuals could be vaccinated outside of the US. Additionally, 
+there is no county level data for counties in Texas and Hawaii. There are unknown 
+counties in each state and a row for unknown county and unknown state.
```
krivard, line 10: "There are unknown counties in each state and a row for unknown county and unknown state." 

Can you elaborate on this?

cdc_vaccines/params.json.template
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "archive": {
+    "aws_credentials": {
+      "aws_access_key_id": "",
+      "aws_secret_access_key": ""
+    },
+    "bucket_name": "",
+    "indicator_prefix": "usafacts",
+    "cache_dir": "./cache"
+  }
```
krivard, line 34: We typically don't put archiving parameters in the developer params template, and we _definitely_ don't want to overwrite the usafacts archive cache.
```suggestion
```

krivard (changes requested):
Excellent start!

cdc_vaccines/README.md
Ananya-Joshi: Sorry it wasn't clear! Here is an image from the raw data that better shows what is happening: 
Screen Shot 2021-09-10 at 3 06 59 PM
These are the unknown counties in each state, and the last row here is an unknown county in an unknown state. They have non-zero values for vaccination.

cdc_vaccines/delphi_cdc_vaccines/pull.py
Ananya-Joshi: Removed!

cdc_vaccines/delphi_cdc_vaccines/pull.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Note that the raw dataset gives the `cumulative` metrics, from which
+    we compute `counts` by taking first differences.  Hence, `counts`
+    may be negative.  This is wholly dependent on the quality of the raw
+    dataset.
+
+    We filter the data such that we only keep rows with valid FIPS, or "FIPS"
+    codes defined under the exceptions of the README.  The current  exceptions
+    include:
+    # - 0: statewise unallocated
```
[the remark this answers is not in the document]
dshemetov: Sure I'll take a look on Monday!

cdc_vaccines/delphi_cdc_vaccines/constants.py
```diff
@@ -0,0 +1,54 @@
...
+	FULL_VAC_65P     ,
+	PART_VAC_SUM     ,
+	PART_VAC         ,
+	PART_VAC_SUM_12P ,
+	PART_VAC_12P     ,
+	PART_VAC_SUM_18P ,
+	PART_VAC_18P     ,
+	PART_VAC_SUM_65P ,
+	PART_VAC_65P
+]
```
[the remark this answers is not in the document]
Ananya-Joshi: Changed! Thank you for the suggestion!

Ananya-Joshi:
The changes requested have been made. We should talk about the two open issues still - the unknown states/counties and the megacounty thread.

cdc_vaccines/README.md
krivard:
```suggestion
there is no county level data for counties in Texas and Hawaii. Each state has some vaccination counts assigned to "unknown county". Some vaccination counts are assigned to "unknown state, unknown county".
```

cdc_vaccines/delphi_cdc_vaccines/pull.py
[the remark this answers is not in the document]
dshemetov: Alright, looking through our chats on prop signals, this is what can happen: if megacounties are used to keep track of unassigned counts, then we must be careful not to add the megacounty population counts when doing prop signals, otherwise when aggregating to larger geos we will double count the state population. I don't see any prop signals atm, so this seems fine.

cdc_vaccines/delphi_cdc_vaccines/pull.py
```diff
@@ -0,0 +1,136 @@
...
+                  "cumulative_counts_tot_vaccine_18P",
+                  "cumulative_counts_tot_vaccine_65P",
+                  "cumulative_counts_part_vaccine",
+                  "cumulative_counts_part_vaccine_12P",
+                  "cumulative_counts_part_vaccine_18P",
+                  "cumulative_counts_part_vaccine_65P",
+                  "timestamp"]
+    df_dummy = df.loc[(df["fips"]!='00000') & (df["timestamp"] == min(df["timestamp"]))].copy()
+    #handle fips 00000 separately
+    df_oth = df.loc[((df["fips"]=='00000') &
```
dshemetov, line 97: What does the 00000 FIPS code represent? Not assigned to any county or state?

cdc_vaccines/delphi_cdc_vaccines/pull.py
```diff
@@ -0,0 +1,136 @@
...
+                    "cumulative_counts_part_vaccine_12P",
+                    "cumulative_counts_part_vaccine_18P",
+                    "cumulative_counts_part_vaccine_65P",
+                    ]] = 0
+
+    df =pd.concat([df_dummy, df])
+    # Obtain new_counts
+    df.sort_values(["fips", "timestamp"], inplace=True)
+    for to, from_d in DIFFERENCE_MAPPING.items():
+        df[to] = df[from_d].diff()
```
dshemetov, line 115: Btw, you might like [this](https://github.com/cmu-delphi/covidcast-indicators/blob/b4c9b5573232b89c45cf3ce8ea6d0bf61ef53133/jhu/delphi_jhu/pull.py#L27) version of taking diffs, grouped by geos.

dshemetov: I think the `.isna()` in that code should probably be replaced with `.loc[min_time_value, :]` or something, but the gist is there.

cdc_vaccines/delphi_cdc_vaccines/run.py
```diff
@@ -0,0 +1,97 @@
...
+            new_col="geo_id",
+            new_code=geo_map,
+            date_col="timestamp")
+        df["val"] = df[["geo_id", sensor]].groupby("geo_id")[sensor].transform(
+            smoother[0].smooth
+        )
+        df["se"] = np.nan
+        df["sample_size"] = np.nan
+        sensor_name = sensor + smoother[1]
+        print(sensor_name)
```
dshemetov, line 75: Also, if you anticipate any nans in the dataset that will be ingested into the db, consider annotating them with nancodes in the missing columns, like [here](https://github.com/dshemetov/covidcast-indicators/blob/263485b8d4d94a36ebd9ed6292d29b62b69efc64/jhu/delphi_jhu/run.py#L67).

cdc_vaccines/delphi_cdc_vaccines/pull.py
Ananya-Joshi: I think it refers to when the state and the county are both listed as "UNK". So yes, it only gets used at aggregate at the nation level. What do you think?

cdc_vaccines/delphi_cdc_vaccines/run.py
Ananya-Joshi: Should I remove these columns? I kept them because they were in the USAFacts indicator, but I can remove them if needed.

cdc_vaccines/delphi_cdc_vaccines/pull.py
Ananya-Joshi: I wonder if we can keep this method for now, but then later I'll look to fix the method and use it.

cdc_vaccines/delphi_cdc_vaccines/pull.py
dshemetov: That makes sense, yea. Just curious about the data! Not much to do here.

cdc_vaccines/delphi_cdc_vaccines/pull.py
dshemetov: Sure, that's alright. A possible refactor for later.

cdc_vaccines/delphi_cdc_vaccines/run.py
dshemetov: Oh no, I mean that we now support three extra columns `missing_value`, `missing_stderr`, and `missing_sample_size`, where we write nancodes for why a value is missing. See the link for an example. The possible nancodes are [here](https://github.com/dshemetov/covidcast-indicators/blob/263485b8d4d94a36ebd9ed6292d29b62b69efc64/_delphi_utils_python/delphi_utils/nancodes.py). So for example, this indicator is not a sample-based indicator so stderr and sample_size just do not apply, so we would write `Nans.NOT_APPLICABLE` in the corresponding columns.

krivard: Since this is a new indicator, we should definitely be creating it with the correct Nans codes in the missing_* columns as needed. It sounds like this indicator should use:
* `missing_stderr = Nans.NOT_APPLICABLE`
* `missing_sample_size = Nans.NOT_APPLICABLE`

cdc_vaccines/delphi_cdc_vaccines/pull.py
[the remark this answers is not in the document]
Ananya-Joshi: Okay! So from what I understand, I'll leave as in for now!

cdc_vaccines/delphi_cdc_vaccines/run.py
```diff
@@ -69,8 +68,8 @@ def run_module(params):
         df["val"] = df[["geo_id", sensor]].groupby("geo_id")[sensor].transform(
             smoother[0].smooth
         )
-        df["se"] = np.nan
-        df["sample_size"] = np.nan
+        df["se"] = nancodes.Nans.NOT_APPLICABLE
+        df["sample_size"] = nancodes.Nans.NOT_APPLICABLE
```
dshemetov, line 72: This is more what I meant
```suggestion
        df["se"] = np.nan
        df["sample_size"] = np.nan
        df["missing_se"] = nancodes.Nans.NOT_APPLICABLE
        df["missing_sample_size"] = nancodes.Nans.NOT_APPLICABLE
```

Ananya-Joshi: Sorry! I messed up version control. Thanks for the catch - it should all be taken care of now :)

Ananya-Joshi: It looks like create_export_csv in delphi_utils needs the 'se' and 'sample_size' columns - not the missing columns - should I change the delphi utils function? 

../../_delphi_utils_python/delphi_utils/export.py:73: in create_export_csv
    export_df = df[df["timestamp"] == date][["geo_id", "val", "se", "sample_size",]]

Ananya-Joshi: Okay, I'll wait until that gets merged in.

Ananya-Joshi:
@krivard @dshemetov , do we need to do anything else for this indicator?

cdc_vaccines/delphi_cdc_vaccines/run.py
dshemetov: Now that #1252 has been merged, my suggestion should work

dshemetov:
@Ananya-Joshi just the Nancodes annotation in the thread above and it should be good to go 👍

Ananya-Joshi:
> @Ananya-Joshi just the Nancodes annotation in the thread above and it should be good to go 👍
As far as I can see on my local, they're changed! Is this not the case for what you can see?

dshemetov:
Oh, now we're missing the actual "se" and "sample_size" columns. I will make a commit with what I mean.

dshemetov (approved):
LGTM

krivard:
This looks like rebase leavings:
image

Indeed, the gitref for "Remove test for old export func" is supposed to be 5fa63149d and not ca09586. @Ananya-Joshi can you clarify what was done here? I want to make sure when we merge it we won't wind up with duplicate copies of all of those dual-author commits.

PR #1239, opened 2021-09-08
Add R and apt packages to container image

korlaxxalrok:
### Description

Add some necessary packages

### Changelog

- Dockerfile: add packages

### DIFF
--- a/facebook/Dockerfile
+++ b/facebook/Dockerfile
@@ -17,7 +17,7 @@ RUN apt-get update && apt-get install -qq -y \
     python3-venv \
     sshpass \
     openssh-server \
-    sendmail
+    ssmtp
 
 ADD ./delphiFacebook /facebook/delphiFacebook/
 ADD ./static /facebook/static/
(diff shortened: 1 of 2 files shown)


nmdefries (commented):
Everything seems to install correctly, but `sendmail` hangs if I try to call it. Is there additional configuration we need to do for `sendmail`?

korlaxxalrok:
Ah, good catch. I'll check.

korlaxxalrok:
@nmdefries I think this will work.

We'll add ssmtp and a simple config to the container that will relay mail to the container host.

We'll have to run the container with the host networking option (`--net=host`) so that we have simple access to the host via `localhost`, and then the host will relay again through a specific server based on the email domain of the recipient (automagically configured as best I can tell, but this appears to be the mechanism on our VMs owned by CMU). I've tested this on a bigchunk with success to both a CMU and Google account.

This will also require a simple change to the postfix config on the container host.

It will be hard to test this locally, but if you have access to a bigchunk (or potentially other system that is part of the `delphi.cmu.edu` domain) you can build the container there and it _should_ work 🤞 

This introduces another dependency on the container host, but later on we can find a way to use an external mail relay. That would be one of the best ways to decouple.

nmdefries (approved):
Looks good. I'm not able to verify that sendmail works (as noted above, doesn't work on any old machine) but it no longer hangs.

PR #1243, opened 2021-09-14
Add NCHS mortality geo aggregation at the HHS and nation levels

### LINKED ISSUE #1041: Make NCHS data available at HHS, nation level
The NCHS mortality data is currently only available at the state level. It seems like it should be possible to aggregate it to the nation and HHS levels. (If it's *not* possible for some reason, we should document that so nobody tries to use state and aggregate themselves.) Having all our signals consistently available at HHS and nation when possible would make it easy to compare things.

There's no pressing need for this that I know of; I just noticed the inconsistency and think it'd be nice to fix.

alexcoda:
### Description
Add NCHS mortality geo aggregation at the HHS and nation levels.

Note that because one of the columns being aggregated is a percentage, we're doing a weighted sum with the geomapper

### Fixes 
- Closes #1041 
- Closes #1213 which this PR was based off of (because I'm on a fork, I couldn't push directly to that open PR)


alexcoda:
@krivard - ready for review!

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -82,30 +95,35 @@ def run_module(params: Dict[str, Any]):
             for sensor in SENSORS:
                 print(metric, sensor)
                 df = df_pull.copy()
+                df["se"] = np.nan
+                df["sample_size"] = np.nan
```
krivard, line 99: this can be pulled out of the for loop

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -82,30 +95,35 @@ def run_module(params: Dict[str, Any]):
...
+                sensor_name = "_".join([SENSOR_NAME_MAP[metric], sensor])
+                if geo in ["hhs", "nation"]:
+                    df = gmpr.replace_geocode(
+                        df, "state_id", "state_code", from_col="geo_id", date_col="timestamp")
+                    df = gmpr.replace_geocode(
+                        df, "state_code", geo, date_col="timestamp").rename(columns={geo: "geo_id"})
```
krivard, line 105: can we reduce some of the duplication with `percent_of_expected_deaths` here at all?

nchs_mortality/delphi_nchs_mortality/run.py
krivard: actually it can probably be pulled out of the if block as well

krivard (commented):
As a bonus, resolving the code-duplication questions may get the linter to hush. 
I could be convinced otherwise though if I've misread what's going on, wow that is a monster.

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -18,7 +19,7 @@
 from .pull import pull_nchs_mortality_data
 
 
-def run_module(params: Dict[str, Any]):
+def run_module(params: Dict[str, Any]):  # pylint: disable=too-many-branches, too-many-statements
```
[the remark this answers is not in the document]
alexcoda: Definitely not, just pushed some changes to pull out some helper methods!

nchs_mortality/.pylintrc
```diff
@@ -4,6 +4,7 @@
 disable=logging-format-interpolation,
     too-many-locals,
     too-many-arguments,
+    fixme,
```
alexcoda, line 7: Added this for a TODO I left in the code. Let me know if this is something we'd want to keep consistent across all indicators and I can add it to others.

PR #1247, opened 2021-09-14
Un-retire schooling indicators from sirCAL

nmdefries:
### Description
Revert #1192. Schooling indicators are now turned back on.

### Changelog
- Remove `*_school_fulltime` and `*_school_parttime` from sirCAL retired signals for fb-survey.


krivard:
This is difficult to review due to the length of the list. Could we take this opportunity to add some whitespace for readability?

Something like

```
"retired-signals": [
    "item1", "witem1", "item2", "witem1", ...[just out to 80 characters]
    "itemN", "witemN"
],
```

krivard (approved):
👍  excellent, thanks!

PR #1249, opened 2021-09-15
Support experimental survey Wave 12

nmdefries:
### Description
Support experimental version of survey Wave 12. This adds qualifying data to the microdata but does not include it in any aggregates. [FB-defined specs are here](https://docs.google.com/document/d/16og11mGQjiMkLuVt7M-C1FPusVrAkp3nJSjRirEpU2M/edit?usp=sharing).

New variable and indicator definitions are not needed here since experimental Wave 12 data is not being included in any aggregates. We will need to add those for post-experimental Wave 12.

### Changelog
- Create function `experimental_arm_assignment` to determine experimental version displayed based on assigned random number. Arm logic is defined in [FB's specs](https://docs.google.com/document/d/16og11mGQjiMkLuVt7M-C1FPusVrAkp3nJSjRirEpU2M/edit?usp=sharing).
- Call experimental Wave 12 `12.5` (numeric). The finalized post-experimental Wave 12 is planned to have a different survey ID and be called `12` (numeric) for easier coding and to avoid date-based logic. Generally speaking, we expect `wave` to be an integer, although no logic explicitly depends on that. I've opted to code experimental Wave 12 as numeric (rather than string, which would have a clearer meaning) so that the microdata output `wave` field won't change type.
- Drop wave "12.5" from data for aggregation; we don't want to include experimental data.
- Exclude `w12_assignment` and `module`, along with `raceethnicity`, from being used in `filter_complete_responses`.
- Update tests.


facebook/delphiFacebook/R/responses.R
```diff
@@ -361,7 +367,8 @@ filter_data_for_aggregation <- function(df, params, lead_days = 12L)
                dplyr::between(.data$hh_number_sick, 0L, 30L),
                dplyr::between(.data$hh_number_total, 1L, 30L),
                .data$hh_number_sick <= .data$hh_number_total,
-               .data$day >= (as.Date(params$start_date) - lead_days)
+               .data$day >= (as.Date(params$start_date) - lead_days),
+               !(.data$wave %in% c(12.5)) # Ignore experimental Wave 12 data
```
capnrefsmmat, line 371: Why not just `.data$wave != 12.5`?

facebook/delphiFacebook/R/responses.R
```diff
@@ -500,6 +507,28 @@ module_assignment <- function(input_data, wave) {
...
+#' 
+#' @return data frame with new `module` column
+#' @importFrom dplyr case_when
+experimental_arm_assignment <- function(input_data, wave) {
+  if (wave == 12.5) {
+    assert( "random_num_exp" %in% names(input_data) )
+    input_data$w12_treatment <- case_when(
+      input_data$random_num_exp >= 0.6666 ~ 1, # demographics placed after symptom items
+      input_data$random_num_exp >= 0.3333 ~ 2, # demographics placed after vaccine items
+      input_data$random_num_exp < 0.3333 ~ 3, # alternative wording to V1
```
capnrefsmmat, line 524: Have we verified with Adrianne that the Qualtrics setup will be using the same number of decimal places and the same choice of `>=` and `<`? Wouldn't want to have goofy issues where 0.1% of people are misclassified because of boundary cases

capnrefsmmat (commented):
A couple questions, but looking good so far

facebook/delphiFacebook/R/responses.R
nmdefries: Yup, this logic is taken directly from Adrianne's specifications.

facebook/delphiFacebook/R/responses.R
```diff
@@ -667,9 +701,11 @@ filter_complete_responses <- function(data_full, params)
...
 
   # 9 includes StartDatetime, EndDatetime, Date, token, wave, geo_id,
-  # UserLanguage + two questions (ignore raceethnicity field which may or may
-  # not exist, depending on params)
-  valid_row_filter <- rowSums( !is.na(data_full[, names(data_full) != "raceethnicity"]) ) >= 9
+  # UserLanguage + two questions (ignore raceethnicity, module, and
+  # w12_assignment fields which may or may not exist, depending on params and
+  # survey version)
+  ignore_cols <- c("raceethnicity", "w12_assignment", "module")
+  valid_row_filter <- rowSums( !is.na(data_full[, !(names(data_full) %in% ignore_cols)]) ) >= 9
```
capnrefsmmat, line 708: Hm, you're adding both `w12_assignment` and `module` here, but I thought `module` was also in Wave 11. Were we applying the inclusion criteria incorrectly for Wave 11 before this?

facebook/delphiFacebook/R/responses.R
```diff
@@ -511,11 +539,13 @@ module_assignment <- function(input_data, wave) {
...
 #' @param input_data data frame of responses
 #' @param county_crosswalk crosswalk mapping ZIP5 to counties
+#' @param params list containing `produce_individual_raceeth`, indicating
+#'   whether or not to issue microdata with race-ethnicity field
 #' @importFrom stringi stri_trim stri_replace_all
 #' @importFrom dplyr left_join group_by filter ungroup select rename
 #'
 #' @export
-create_complete_responses <- function(input_data, county_crosswalk)
+create_complete_responses <- function(input_data, county_crosswalk, params)
```
capnrefsmmat, line 548: ...how did this work before?

capnrefsmmat (approved):
Looks good; just some questions

facebook/delphiFacebook/R/responses.R
nmdefries: Yeah, with the previous version of the filter Wave 11 responses would have only required 1 answered question, so we'll need to reissue those microdata (or run the existing files through just this corrected filter?).

facebook/delphiFacebook/R/responses.R
nmdefries: Something something environments. This does work in actual runs -- `params` is in the global environment and thus accessible to `create_complete_responses` -- but not in the integration tests, since `params` isn't defined globally there (the params are passed to `run_facebook` as an unnamed object).

nmdefries:

capnrefsmmat:
@krivard FYI, the first possible day for Wave 12 to come out is tomorrow, Sept 30, though we may have to delay due to issues with Qualtrics affecting the international instrument.

krivard:
We now deploy surveys through the same github mechanism as all the other indicators 🎉 

I will merge this code and deploy to prod. When we're ready to begin pulling in qualtrics files for wave 12, we should update [the production params file](https://github.com/cmu-delphi/covidcast-indicators/blob/main/ansible/templates/facebook-params-prod.json.j2).

PR #1251, opened 2021-09-15
Update setup.py files to "Python :: 3.8" annotation

dshemetov:
### Description
Small annotation in setup.py files for each indicator for consistency with Makefiles.

### Changelog
Change Python :: 3.7 to Python :: 3.8.


qx-teo (approved):
LGTM!

krivard (approved):
Verified that we are in fact using Python 3.8 on primary, so this is a low-risk change even though it touches all our indicators at once.

PR #1253, opened 2021-09-16
Make survey tests robust to not having a restore cache

nmdefries:
### Description
Make survey tests not fail when cache containing installed R packages from previous test run is not loaded.

### Changelog
- Check if readr is installed before checking its version, to avoid "there is no package called readr" errors.

### Fixes 
When no cache is loaded, the call `packageVersion("readr")` errors out since the `readr` package doesn't exist yet.


dshemetov (approved):
LGTM

PR #1259, opened 2021-09-20
Set E2 to integer on read

nmdefries:
### Description
Set read type for E2 schooling question. In some wave 11 input files, `read_csv` is trying to read as boolean.

### Changelog
- List E2_1 and E2_2 as integer in `read_csv` specs


krivard:
This has now been deployed.

PR #1263, opened 2021-09-24
Route validator to validator_dir

qx-teo:
### Description
Make validator (and indicator functions) run in ./receiving, only transfer files to export_dir if validation is passed.
We ensure that changes aren't too drastic by only making one indicator (google_symptoms) send data to ./receiving, the rest directly upload files to export_dir and skip validation.

### Changelog
Add code for transferring files between folders:
- runner.py
- utils.py
Make validation run in validation_dir:
- validate.py
- datafetcher.py
Change indicator function to use validation_dir:
- run.py (in google_symptoms)
- params.json.template (in google_symptoms, ansible)


qx-teo:
I've some weird failure of geomap locally, just pushing it here to see if it's the same behavior (I doubt so!)

PR #1265, opened 2021-09-26
Change geomapper class to use "timestamp" as the default name for date_col's

alexcoda:
### Description
Change the default date_col name to "timestamp" from "date". The `export_csv` function expects "timestamp", and most indicators were specifying it as this anyways.

Note that there were two indicators (changehc, and claims_hosp) which were using the column named like "date" and this has been changed to "timestamp".


dshemetov:
LGTM

PR #1266, opened 2021-09-26
Minor clean up of changehc sensor updating

alexcoda:
### Description
Some minor cleanup, no functional changes

### Changelog
- Remove redundant start/end date setting within changehc. This statement was never executed because we always specify the start and end dates.
- Simplify logic for handling signals, since only 1 signal is being processed within `CHCSensorUpdate.update_sensor`.
- Remove unnecessary uses of .loc throughout all indicators.


changehc/delphi_changehc/update_sensor.py
```diff
@@ -113,13 +109,13 @@ def __init__(self,
```
dshemetov, line 109: V minor: can we also change `CDCSensorUpdator` to `CDCSensorUpdater`? Updator makes me think of giant robots, for some reason 😄

chinandrew: > Updator makes me think of giant robots, for some reason

You act like this is a problem :thinking: lol just changed this in a new commit

dshemetov: Haha, you're not wrong! Thanks!

PR #1273, opened 2021-09-28
Implement second pass at validation gating

qx-teo:
### Description
- Parsing new "delivery" params, which houses delivery_dir
- If delivery exists, and validation passes, files to be transferred from export_dir to delivery_dir after archiving

To add an indicator to the new gating procedure, we would add a new "delivery_dir" to its params and change the current export_dir to ./receiving. This means:
- validation/archiving can be done within ./receiving
- Delivery transfers it to /common/covidcast for acquisition
- If validation fails, we perform the gating step by keeping the data within ./receiving, until validation passes (either by the source correcting inaccuracies raised by validation, or by us adding exceptions to validation errors)

### Changelog
- utils.py (adding in delivery function)
- runner.py (adding delivery to the pipeline)


nmdefries (approved):
Looks good!

Issue #1269, opened 2021-09-26
Move weekday.py into shared utils


chinandrew:
The weekday.py file used in CHNG/DV/Claims_hosp is almost identical between them (CHNG and claims_hosp _are_ identical, while DV just handles multiple columns in a loop), so moving into a shared util will reduce code duplication.

PR #1282, opened 2021-10-01
[DRAFT] Standardize weekday calc_adjustment() function

chinandrew:
### Description
As first step for #1269, refactor DV, CHNG, and Claims_hosp to use the same calc_adjustment function. CHNG and claims_hosp were previously using the same one while DV had a multi-column version.  This means all 3 indicators have the exact same function, and later we'll move them out to a single shared file

### Changelog
- Add column argument to calc_adjustment and have it loop through each one.
- Update index management outside function 
- Reshape returns weekday params to be 2D arrays universally (further refactoring to unify get_params() is next)

### Fixes 
- Partially addresses #1269


chinandrew:
@mariajahja @rumackaaron For :eyes:  on this draft.

PR #1283, opened 2021-10-02
Possible minor archiver update

dshemetov:
### Description
Attempting to fix the current CHNG outage, grasping for bugs without much of a lead. Found what I thought was an issue, improved test coverage, but then the improved tests pass even without these archiver changes, so 🤷. Keeping this here, since it's probably good to integrate in for future clarity.

If we do find an issue with the archiver, I'll just use this PR to make that hotfix.

### Changelog
* do not set missing columns for deleted rows, if `after_df` does not contain them
* improve test coverage for the cases of deleted rows without missing columns


dshemetov:
Less of a review request and more of ping for awareness.

krivard (approved):
👍  definitely in favor of better test coverage even if it didn't find anything wrong

krivard:
(fix your lint)

PR #1284, opened 2021-10-04
Ignore existing *.tgz files when creating tidy archives

korlaxxalrok:
### Description
Fix:

Recent changes to the survey indicator's Makefile mean that we now end up storing the tidy .tgz archive files in ./tidy. We need to exclude these archive files so we don't have runaway storage issues.

### Changelog
- Update Makefile


krivard (approved):
🎉

PR #1296, opened 2021-10-06
Combine `summarize_*` filter in surveys aggregation

nmdefries:
### Description
Combine column and row filters in `summarize_indicators_day`/`summarize_aggregations_group` function for small memory improvement (on my machine, minus a few GB at peak memory usage).

### DIFF
--- a/facebook/delphiFacebook/R/aggregate.R
+++ b/facebook/delphiFacebook/R/aggregate.R
@@ -181,6 +181,9 @@ summarize_indicators_day <- function(day_df, indicators, target_day, geo_level, 
 
       # Copy only columns we're using.
       select_cols <- c(metric, var_weight, "weight_in_location")
+      # This filter uses `x[, cols, with=FALSE]` rather than the newer
+      # recommended `x[, ..cols]` format to take advantage of better
+      # performance. Switch to using `..` if `with` is deprecated in the future.
       ind_df <- sub_df[!is.na(sub_df[[var_weight]]) & !is.na(sub_df[[metric]]), select_cols, with=FALSE]
       
       if (nrow(ind_df) > 0)
--- a/facebook/delphiFacebook/R/aggregate.R
+++ b/facebook/delphiFacebook/R/aggregate.R
@@ -181,7 +181,7 @@ summarize_indicators_day <- function(day_df, indicators, target_day, geo_level, 
 
       # Copy only columns we're using.
       select_cols <- c(metric, var_weight, "weight_in_location")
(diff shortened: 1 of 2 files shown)


facebook/delphiFacebook/R/aggregate.R
```diff
@@ -181,7 +181,7 @@ summarize_indicators_day <- function(day_df, indicators, target_day, geo_level,
 
       # Copy only columns we're using.
       select_cols <- c(metric, var_weight, "weight_in_location")
-      ind_df <- sub_df[, select_cols, with=FALSE][!is.na(sub_df[[var_weight]]) & !is.na(sub_df[[metric]]), ]
+      ind_df <- sub_df[!is.na(sub_df[[var_weight]]) & !is.na(sub_df[[metric]]), select_cols, with=FALSE]
```
capnrefsmmat, line 184:
```suggestion
      ind_df <- sub_df[!is.na(sub_df[[var_weight]]) & !is.na(sub_df[[metric]]), ..select_cols]
```

I was wondering what `with=FALSE` does and I think you can do this instead, according to data.table's docs: https://rdatatable.gitlab.io/data.table/reference/data.table.html

It even says that "Note that `x[, cols]` is equivalent to `x[, ..cols]` and to x`[, cols, with=FALSE]` and to `x[, .SD, .SDcols=cols]`", so maybe `..` isn't needed either. This appears to be true as of data.table [v1.11.0](https://rdatatable.gitlab.io/data.table/news/index.html#data-table-v1-11-0-01-may-2018-2018-05-01) (see item 18 under New Features).

capnrefsmmat (changes requested):
Good catch (and lol several GB from a single line). I think it can be simplified further though, if it matters

facebook/delphiFacebook/R/aggregate.R
nmdefries: From a small benchmarking experiment, the `with` format appears to be faster (~16%) than `..`. The New Features note does say they intend to deprecate `with` in "the next few years" (as of 2018) in favor of the dots, so maybe now's the time to switch?

capnrefsmmat: I was definitely not expecting a performance difference there...

Reviewing the data.table release notes, they generally seem to communicate deprecations well in advance. First new versions will issue a warning to say the feature is deprecated, and a year or two later they'll remove it. So we're probably safe to benefit from the performance gain for now. Maybe it'd be worth adding a comment to the code so if they do deprecate it a year from now, we remember why we used `with`.

capnrefsmmat (approved):

PR #1298, opened 2021-10-07
Add vaccination status breakdown to contingency tables

nmdefries:
### Description
Report all indicators for vaccinated/accepting/hesitant groupings.

### Changelog
- `contingency_variables.R`: define vaccination status var and function
- `responses.R`: add func to `load_response_one`
- `contingency_indicators.R`: add vaccination status var to list of breakdowns

### DIFF
--- a/facebook/delphiFacebook/R/contingency_variables.R
+++ b/facebook/delphiFacebook/R/contingency_variables.R
@@ -245,8 +245,7 @@ code_vaccinated_breakdown <- function(input_data, wave) {
       input_data$V3a == 4 ~ "Hesitant",
       TRUE ~ NA_character_
     )
-  }
-  else if (all(c("V1", "V3") %in% names(input_data))) {
+  } else if (all(c("V1", "V3") %in% names(input_data))) {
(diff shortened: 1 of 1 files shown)


facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -227,6 +227,45 @@ code_health <- function(input_data, wave) {
...
 
+#' Vaccinated summary variable for making table cuts
+#'
+#' @param input_data input data frame of raw survey data
+#' @param wave integer indicating survey version
+#' 
+#' @return augmented data frame
+code_vaccinated_breakdown <- function(input_data, wave) {
+  # grouping variable - vaccination status
+  if (any(c("V1", "V3", "V3a", "V11a") %in% names(input_data))){
```
capnrefsmmat, line 238: Is this outer `if` really necessary? Feels like you could just have the inner `if` and `else if`, and then you only need one `else` clause.

capnrefsmmat (changes requested):
Minor simplification, but looks right otherwise

PR #1302, opened 2021-10-07
Fix hanging survey CI tests

nmdefries:
### Description
Import `Rcpp` more explicitly [following this suggestion](https://lists.r-forge.r-project.org/pipermail/rcpp-devel/2014-February/007120.html). Seems like we were implicitly using `Rcpp` as loaded by `lubridate`, but after update 1.7.10 -> 1.8.0, `lubridate` no longer uses `Rcpp` and our package begins raising weird errors and hanging when trying to run C++ code.

### Changelog
- Add specific Rcpp import to the code calling the C++ function so that Rcpp gets imported in `NAMESPACE`.


capnrefsmmat (approved):
weird

PR #1303, opened 2021-10-08
Survey changes for the finalized version of Wave 12

nmdefries:
### Description
Support Wave 12 in survey package.

### Changelog
- Update fields based on item V15a. Wave 12 renamed to V15c, adding new response options and slightly clarifying the wording.
- Update fields based on items V5abc. Wave 12 added new response options.
- Update fields based on item B13. Wave 12 renamed to B13a, slightly clarifying the wording.
- Redefine full time/part time indoor schooling items for oldest child only to be based on P5.
- Define new `vaccinate_children` item for oldest child only to be based on P3.

~TODO: update V1-dependent logic if finalized version of Wave 12 switches to using V1alt (asks who respondent knows who've been vaccinated).~


nmdefries:
Wave 12 is not going to change to using V1alt, so this is ready to go.

PR #1324, opened 2021-10-19
Suppress rapid_change_num_rows

qx-teo:
### Description
Test constantly failing for covid_ag_raw_pct_positive since not all geo_ids have data every day, we suppress errors for this since it's a false positive

### Changelog
- params.json.template for both quidel folder and ansible.

### DIFF
--- a/ansible/templates/quidel_covidtest-params-prod.json.j2
+++ b/ansible/templates/quidel_covidtest-params-prod.json.j2
@@ -28,7 +28,14 @@
       "dry_run": true,
       "suppressed_errors": [
         {"check_name": "check_rapid_change_num_rows",
-        "signal": "covid_ag_raw_pct_positive"}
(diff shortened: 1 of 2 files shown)


ansible/templates/quidel_covidtest-params-prod.json.j2
```diff
@@ -26,7 +26,10 @@
       "min_expected_lag": {"all": "5"},
       "max_expected_lag": {"all": "5"},
       "dry_run": true,
-      "suppressed_errors": []
+      "suppressed_errors": [
+        {"check_name": "check_rapid_change_num_rows",
+        "signal": "covid_ag_raw_pct_positive"}
```
nmdefries, line 31: Does this need to be suppressed for all geo types?

nmdefries (commented):
Waiting to verify that very different number of geos per day is expected behavior.

PR #1335, opened 2021-10-25
Discontinue unweighted survey signals

nmdefries:
### Description
We currently report in the API weighted and unweighted versions of every survey signal. This roughly doubles the resources required to produce and store the survey data. However, due to reduced urgency at this point in the pandemic, unweighted signals that are available slightly sooner don't add much value when more-accurate weighted signals are also available.

[FB Asana task with more info](https://app.asana.com/0/1199704294865989/1201238136692629/f)

### Changelog
- Remove all unweighted signals (those using uniform weights) from indicator lists in `binary.R` and `count.R`.
- Suppress removed signals in sirCal ansible and package params templates.

Note: This is waiting for feedback from API users to see if anyone still has a strong need for particular unweighted signals. Aiming to deploy Nov 30.

PR #1344, opened 2021-10-28
[survey] Produce CID lists first

nmdefries:
### Description
Produce and post CID lists first so they get to FB on time, regardless of how long the rest of the pipeline takes.

### Changelog
- `params.json` target to output only `cids`.
- Create new `update-params.json` target to change output to all other output types (`covidalert`, `community`, etc)
- Change order of dependencies for `pipeline` target.


nmdefries:
Holding off on this for now since survey API runs are currently taking <1 hour and [deprecating unweighted signals](https://github.com/cmu-delphi/covidcast-indicators/pull/1335) will reduce that even more.

nmdefries:
Survey has ended. This was never used.

PR #1353, opened 2021-11-03
[survey] Stop with useful message if any input files fail on load/processing

### LINKED ISSUE #1349: Surveys pipeline should report file processing errors more sensibly
[Slack thread context](https://delphi-org.slack.com/archives/C0130CSQRN3/p1635515085013700)

The surveys pipeline does this:
https://github.com/cmu-delphi/covidcast-indicators/blob/cb3721b21872560b4077d842af46cfd728653db9/facebook/delphiFacebook/R/responses.R#L22-L28

If one of the input files has a processing error, it shows up in the `input_data` list as a `try-error` instead of a `data.frame`like. Here's an example from a recover session:

```
Browse[1]> lapply(input_data, class)
[[1]]
[1] "tbl_df"     "tbl"        "data.frame"

[[2]]
[1] "try-error"

[[3]]
[1] "spec_tbl_df" "tbl_df"      "tbl"         "data.frame" 

[[4]]
[1] "try-error"
```

Then `bind_rows` errors out with the somewhat-obtuse "Argument N must have names".

Instead, we'd like the pipeline to report the `try-error`s and the filenames they go with, to help make repairs go quicker.

nmdefries:
### Description
Stop with useful message if any input files fail on initial load and processing in `responses.R::load_response_one`.

### Changelog
- `responses.R`
    - Before binding input file results together, check output type. If any are of type `try-error`, exit with message specifying problematic file names and error message.

### Fixes 
Closes #1349.

### DIFF
--- a/facebook/delphiFacebook/R/responses.R
+++ b/facebook/delphiFacebook/R/responses.R
@@ -38,7 +38,7 @@ load_responses_all <- function(params, contingency_run = FALSE) {
   which_errors <- unlist(lapply(input_data, inherits, "try-error"))
   if (any( which_errors )) {
     errored_filenames <- paste(params$input[which_errors], collapse=", ")
-    stop(paste("ingestion and field creation failed for input data file(s)", errored_filenames))
+    stop("ingestion and field creation failed for input data file(s)", errored_filenames)
   }
   
   input_data <- bind_rows(input_data)
--- a/facebook/delphiFacebook/R/responses.R
+++ b/facebook/delphiFacebook/R/responses.R
@@ -18,13 +18,29 @@
 #' @export
 load_responses_all <- function(params, contingency_run = FALSE) {
   msg_plain(paste0("Loading ", length(params$input), " CSVs"))
-  
-  map_fn <- if (params$parallel) { mclapply } else { lapply }
+
+  # To prevent errors from one file from "contaminating" output for other files
+  # that happen to be in the same `mclapply` group, turn `mc.preschedule` off.
+  # This makes debugging problematic input files easier.
+  # See https://stackoverflow.com/questions/18330274/an-error-in-one-job-contaminates-others-with-mclapply
+  # for details.
+  mclapply_no_preschedule <- function(...) {
+    mclapply(..., mc.preschedule = FALSE)
+  }
+
+  map_fn <- if (params$parallel) { mclapply_no_preschedule } else { lapply }
   input_data <- map_fn(seq_along(params$input), function(i) {
     load_response_one(params$input[i], params, contingency_run)
   })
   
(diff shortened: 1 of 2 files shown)


facebook/delphiFacebook/R/responses.R
```diff
@@ -18,13 +18,29 @@
...
   input_data <- map_fn(seq_along(params$input), function(i) {
     load_response_one(params$input[i], params, contingency_run)
   })
   
   msg_plain(paste0("Finished loading CSVs"))
+  
+  which_errors <- unlist(lapply(input_data, inherits, "try-error"))
+  if (any( which_errors )) {
+    errored_filenames <- paste(params$input[which_errors], collapse=", ")
+    stop(paste("ingestion and field creation failed for input data file(s)", errored_filenames))
```
capnrefsmmat, line 41:
```suggestion
    stop("ingestion and field creation failed for input data file(s)", errored_filenames)
```

`stop()` automatically pastes its arguments anyway

facebook/delphiFacebook/R/responses.R
```diff
@@ -18,13 +18,29 @@
...
-  
-  map_fn <- if (params$parallel) { mclapply } else { lapply }
+
+  # To prevent errors from one file from "contaminating" output for other files
+  # that happen to be in the same `mclapply` group, turn `mc.preschedule` off.
+  # This makes debugging problematic input files easier.
+  # See https://stackoverflow.com/questions/18330274/an-error-in-one-job-contaminates-others-with-mclapply
+  # for details.
+  mclapply_no_preschedule <- function(...) {
+    mclapply(..., mc.preschedule = FALSE)
```
capnrefsmmat, line 28: Hmm. With prescheduling turned off, `mclapply` will apparently fork one child process per data file to read (though only at most `mc.cores` at a time). That means there will be m ore time spent starting up new processes. Did you try this out on any test data to see if it has a significant performance cost?

I'm hoping R's startup time is negligible compared to the time of reading in big huge CSVs, but I dunno.

capnrefsmmat (commented):
One question and one tiny suggestion, but it looks good otherwise.

facebook/delphiFacebook/R/responses.R
nmdefries: Good question, I'll run some tests.

nmdefries: Total CPU time of loading 46 input files 427m with prescheduling vs 467m without prescheduling. So loading input files in parallel is ~9% slower without prescheduling. That's not insignificant, so I'll switch prescheduling back on and make the error message clearer.

capnrefsmmat (approved):

PR #1355, opened 2021-11-08
[survey] Add race-ethnicity microdata support to monthly rollup script

nmdefries:
### Description
Add race-ethnicity microdata support to the `monthly-archive.sh` script that calls the R script to do the actual combining of daily CSVs.

### Changelog
- `monthly-archive.sh`
    - Create function to do the rolling up and posting of the resulting files. Takes output directory name as an argument.
    - Call function for both public results (normal microdata) and race-ethnicity microdata.

### DIFF
--- a/facebook/micro/monthly-archive.sh
+++ b/facebook/micro/monthly-archive.sh
@@ -8,13 +8,27 @@ else
 fi
 echo ${MONTH}
 R_MONTH=${MONTH#*_}; R_MONTH=${R_MONTH#0}
-BATCH="cd fb-public-results\nls -1 cvid_responses_${MONTH}*.gz"
-sftp -b <(echo -e "${BATCH}") -P 2222 fb-automation@ftp.delphi.cmu.edu 2>/dev/null | \
-    grep "^cvid" | \
-    awk -F_ 'BEGIN{print "cd fb-public-results"} {key=$3 $4 $5; if (key!=last && last!="") {print record} last=key; record=$0} END{print record}' | \
-    sed '/^cvid/ s/^/get /' >fetch.sftp
-sftp -b fetch.sftp -P 2222 fb-automation@ftp.delphi.cmu.edu
-OUT=${MONTH/_/-}
-Rscript ../monthly-files.R ${MONTH%_*} ${R_MONTH} . >${OUT}.csv
-gzip ${OUT}.csv
-sftp -b <(echo -e "cd fb-public-results\nput ${OUT}.csv.gz") -P 2222 fb-automation@ftp.delphi.cmu.edu
+
+# Save script's arguments to reference by name in perform_rollup_and_post() so the script's
+# positional arguments and perform_rollup_and_post's positional arguments don't get confusing.
+arg0=$0
+arg3=$3
+arg4=$4
+arg5=$5
+
+perform_rollup_and_post ()
+{
+    BATCH="cd $1\nls -1 cvid_responses_${MONTH}*.gz"
+    sftp -b <(echo -e "${BATCH}") -P 2222 fb-automation@ftp.delphi.cmu.edu 2>/dev/null | \
+        grep "^cvid" | \
+        awk -F_ 'BEGIN{print "cd $1"} {key=$arg3 $arg4 $arg5; if (key!=last && last!="") {print record} last=key; record=$arg0} END{print record}' | \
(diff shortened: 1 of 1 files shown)


facebook/micro/monthly-archive.sh
```diff
@@ -8,13 +8,27 @@ else
...
-Rscript ../monthly-files.R ${MONTH%_*} ${R_MONTH} . >${OUT}.csv
-gzip ${OUT}.csv
-sftp -b <(echo -e "cd fb-public-results\nput ${OUT}.csv.gz") -P 2222 fb-automation@ftp.delphi.cmu.edu
+
+# Save script's arguments to reference by name in perform_rollup_and_post() so the script's
+# positional arguments and perform_rollup_and_post's positional arguments don't get confusing.
+arg0=$0
+arg3=$3
+arg4=$4
+arg5=$5
```
krivard, line 17: ah, monthly-archive.sh doesn't have any positional arguments. The $3 $4 $5 in the awk script refer to the third, fourth, and fifth words in the line being parsed from the output of `ls -1 covid_responses...`. The single quote prevents the shell from substituting the $-expressions at execution time, instead preserving them as-is for awk to interpret.

facebook/micro/monthly-archive.sh
```diff
@@ -8,13 +8,20 @@ else
...
-Rscript ../monthly-files.R ${MONTH%_*} ${R_MONTH} . >${OUT}.csv
-gzip ${OUT}.csv
-sftp -b <(echo -e "cd fb-public-results\nput ${OUT}.csv.gz") -P 2222 fb-automation@ftp.delphi.cmu.edu
+
+perform_rollup_and_post ()
+{
+    BATCH="cd $1\nls -1 cvid_responses_${MONTH}*.gz"
+    sftp -b <(echo -e "${BATCH}") -P 2222 fb-automation@ftp.delphi.cmu.edu 2>/dev/null | \
+        grep "^cvid" | \
+        awk -F_ 'BEGIN{print "cd '"$1"'"} {key=$3 $4 $5; if (key!=last && last!="") {print record} last=key; record=$0} END{print record}' | \
```
nmdefries, line 17: On the topic of single quotes changing variable behavior, I want to highlight this change:

```
awk -F_ 'BEGIN{print "cd '"$1"'"}...'
```

was previously

```
awk -F_ 'BEGIN{print "cd fb-public-results"}...'
```

, which requires no variable expansion. To allow expansion of positional argument 1, I placed it it its own double-quoted string chunk to get concatenated with the rest of the `awk` processing command. A small test example seems to work correctly:

```
myvar="fb-public-results"
awk -F_ 'BEGIN{print "cd '"$myvar"'"}'
```

krivard: Thanks for raising. The canonical way to get environment variables into an awk program is this:
```suggestion
        awk -F_ -vDIR="$1"  'BEGIN{print "cd " DIR} {key=$3 $4 $5; if (key!=last && last!="") {print record} last=key; record=$0} END{print record}' | \
```

You can test with
```
thing="foo"; awk -vDIR="$thing" 'BEGIN{print "cd " DIR}'
# cd foo
```

facebook/micro/monthly-archive.sh
```diff
@@ -8,13 +8,20 @@ else
...
+        sed '/^cvid/ s/^/get /' >fetch.sftp
+    sftp -b fetch.sftp -P 2222 fb-automation@ftp.delphi.cmu.edu
+    OUT=${MONTH/_/-}
+    Rscript ../monthly-files.R ${MONTH%_*} ${R_MONTH} . >${OUT}.csv
+    gzip ${OUT}.csv
+    sftp -b <(echo -e "cd $1\nput ${OUT}.csv.gz") -P 2222 fb-automation@ftp.delphi.cmu.edu
+}
```
krivard, line 24: This script was originally designed to leave all the .gz files in place, making it easy to debug if something went wrong. It cleans up all .gz files at the beginning of the script. Since we're running it twice, and presumably want `monthly-files.R` to see a different set of files for each run, _but_ we still want to retain the .gz files for debugging, we need to clean up after ourselves instead:
```suggestion
    rm -rf $1
    mkdir $1
    mv *.gz $1/
}
```

krivard (changes requested):

krivard (approved):
👍  & thank you for letting me be picky 🙂

PR #1325, opened 2021-10-19
Delphi_utils geomapper: update to fips 2020 population estimates

dshemetov:
### Description
See #1319.

### Changelog
- Update source FIPS population source file.
- Update the dependent crosswalks.
- Add a sanity check notebook and remove an outdated one.

### Fixes 
- Closes #1319, #1320

PR #1356, opened 2021-11-09
Add 02261 to valid geo-codes for jhu

qx-teo:
### Description
Fix check_geo_id for validator: data from 02261 was split into 02263 and 02266 in this [commit](https://github.com/cmu-delphi/covidcast-indicators/commit/9ff0f46d603d8fcae5b2a8d08278cd7729e1e945#diff-68ed56b7a741f7fc5187a7701877e602c6d5c842f19923eca9b3159f125736b5) (#1325), but there's still rows coming from jhu for 02261, which is now not a valid geo_id. We add this as a valid geo_id to go about this.

### Changelog
- params.json.template in /jhu/ and ansible


nmdefries (approved):
Looks good

Issue #1360, opened 2021-11-09
MA has N/A values for HHS


qx-teo:
**Actual Behavior:**

HHS seems to have N/A values at state level for Influenza signals sporadically: this seems to happen every week on Tuesdays.  The past few dates with N/A values are: 

- 8/3/2021
- 8/10/2021
- 8/17/2021
- 8/24/2021
- 8/31/2021
- 9/7/2021
- 9/14/2021
- 9/21/2021
- 9/28/2021
- 10/5/2021
- 10/19/2021

**Next Steps**
Add a filtering step in delphi_hhs that removes rows with geo_id == "ma", timestamp > "08-01-2021" and NA values.

elray1:
Possibly related question -- is there some outlier suppression happening in the daily signal that is not present in the 7 day average signal? I am curious about what's going on in this plot: 
![image](https://user-images.githubusercontent.com/5132208/142445150-67164a25-6585-4908-8aa4-80531b174867.png)

Produced by this code:
```

ma_dat <- covidcast::covidcast_signals(
  data_source = "hhs",
  signal = c("confirmed_admissions_influenza_1d", "confirmed_admissions_influenza_1d_7dav"),
  geo_type = "state",
  geo_values = c("ma"))
ma_dat <- dplyr::bind_rows(ma_dat)

ggplot(data = ma_dat) +
  geom_line(mapping = aes(x = time_value, y = value, color = signal)) +
  facet_wrap( ~ geo_value) +
  theme_bw()
```

krivard:
No, we don't do any outlier suppression, how alarming. Let me look into that and maybe open up a separate issue.

PR #1361, opened 2021-11-09
Remove N/A Values for MA at state level

qx-teo:
### Description
MA seems to always give N/A values on Tuesdays, we remove these rows from the dataframe. See #1360 for more details.
In particular, we apply the following filter to HHS:
- Only for confirmed_admissions_influenza_1d (including prop / 7dav variants)
- Dates after 1 August, 2021
- Value is N/A
- Geo_type is state, geo_id is "ma"

### Changelog
- run.py for delphi_hhs

### DIFF
--- a/hhs_hosp/delphi_hhs/run.py
+++ b/hhs_hosp/delphi_hhs/run.py
@@ -117,6 +117,12 @@ def run_module(params):
             df=pop_proportion(df, geo_mapper)
         df = make_geo(df, geo, geo_mapper)
         df = smooth_values(df, smoother[0])
+        # Fix N/A MA values, see issue #1360
+        if geo == "state" and sensor.startswith(CONFIRMED_FLU):
+            print(df.shape, geo, smoother, sensor)
+            df = df[~np.logical_and(np.logical_and(df.val.isna(), df.geo_id == "ma"),
+                                    df.timestamp > "08-01-2021")]
+            print(df.shape)
         if df.empty:
             continue
         sensor_name = sensor + smoother[1]
--- a/hhs_hosp/delphi_hhs/run.py
+++ b/hhs_hosp/delphi_hhs/run.py
@@ -119,10 +119,9 @@ def run_module(params):
         df = smooth_values(df, smoother[0])
         # Fix N/A MA values, see issue #1360
         if geo == "state" and sensor.startswith(CONFIRMED_FLU):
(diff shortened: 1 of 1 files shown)


qx-teo:
@nmdefries @krivard Just thought of something: Maybe we should try to convert this error to a warning - but IIRC there's no way to do that right now for a particular combination of geo-sig-errortype (we only have the suppressed_error framework). 

Should we try to add that in? Or are we OK with just removing NAs for this particular filter for now? 

I'm also not too sure why having NA values is 'bad': I presume we want this for some sort of consistency, but I don't know if it messes with something by feeding in null values. The way I see it is:

1. Validator reports error, and leaves it in (current method with dry-run)
2. Validator reports error, prevents API ingestion (kinda bad given that there's NA data every week, would happen if dry-run was turned off)
3. Delphi_hhs removes rows, no error (This PR)
3a. Adding in a warning when rows are removed? Currently I removed rows under delphi_hhs.run, so the warning wouldn't show up in validator logs 
3b. We can also impute 0s for this particular filter, and give a warning.

Just checking! I know we originally decided not to impute 0s in general, but I don't know if we would want to do it if the scope was narrowed down as such.

krivard:
The NAs here are malformed data from the source dataset, and filtering them out (& not imputing them to zero) is part of the normal expected processing done by this indicator. An ongoing warning is not needed.

hhs_hosp/delphi_hhs/run.py
```diff
@@ -117,6 +117,12 @@ def run_module(params):
             df=pop_proportion(df, geo_mapper)
         df = make_geo(df, geo, geo_mapper)
         df = smooth_values(df, smoother[0])
+        # Fix N/A MA values, see issue #1360
+        if geo == "state" and sensor.startswith(CONFIRMED_FLU):
+            print(df.shape, geo, smoother, sensor)
+            df = df[~np.logical_and(np.logical_and(df.val.isna(), df.geo_id == "ma"),
+                                    df.timestamp > "08-01-2021")]
```
nmdefries, line 124: Two suggestions on this bit:

- I'd prefer to make this filter even narrower by only removing `NA`s occurring on Tuesdays. Seems to me that we should be alerted to unusual missing values that occur on other days of the week.
- `&` can be used to element-wise combine arrays so that the chained logic here is clearer.

nmdefries (changes requested):

PR #1364, opened 2021-11-15
Patch survey microdata rollup names

nmdefries:
### Description
Patch for #1355. Normal monthly microdata rollup files should be called `YYYY-MM.csv.gz` and race-eth rollup files should be called `YYYY-MM-race-ethnicity.csv.gz`. #1355 used the same name for both. Fix that.

### Changelog
- Add second arg to the `perform_rollup_and_post` function to append an additional string to the rollup file name.


facebook/micro/monthly-archive.sh
```diff
@@ -26,5 +26,5 @@ perform_rollup_and_post ()
     mv *.gz $1/
 }
 
-perform_rollup_and_post "fb-public-results"
-perform_rollup_and_post "protected-race-ethnicity-data"
+perform_rollup_and_post "fb-public-results" ""
```
krivard, line 29: the "" is not technically necessary but I'm okay with keeping it in for extra readability

PR #1367, opened 2021-11-18
Make FilesystemArchiveDiffer actually apply changes to cache

krivard:
### Description
FilesystemArchiveDiffer, which is used to build diff-based issues out of confirmation-based issues for data patches without disturbing the S3 archive, is supposed to update the local archive cache in-place. It has apparently _not_ been doing that, this whole time. This PR adds the update mechanism.

I have no idea why we ever thought this was working. Maybe diff_exports used to modify the local cache too, but that was moved in a recent refactor?

### Changelog
- archive.py:FilesystemArchiveDiffer - copy the exported files over to the local cache directory. Stole template code from GitArchiveDiffer for this.

### Fixes 
- Fixes 🤯  from building hhs+flu backissues

### DIFF
--- a/_delphi_utils_python/delphi_utils/archive.py
+++ b/_delphi_utils_python/delphi_utils/archive.py
@@ -641,8 +641,24 @@ class FilesystemArchiveDiffer(ArchiveDiffer):
             successes: All files from input
             fails: Empty list
         """
+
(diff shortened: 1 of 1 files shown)


nmdefries (approved):
Looks good

PR #1372, opened 2021-11-18
Mix respondent weights for all indicators

### LINKED ISSUE #764: Align approaches to mixing for symptom survey aggregates
The old pipeline for community responses did not apply mixing. To reproduce it, we currently ignore the mixed weights for some aggregates. Once a better mixing/weighting scheme is chosen, all signals should use it. Once a new approach is chosen, we may need to reissue old signals or update public-facing documentation about the change.

nmdefries:
### Description
Mix weights for all indicators, and remove the `skip_mixing` flag.

~Note: This change requires that we regenerate gold receiving files (since `val`s and thus `se`s are calculated differently for a subset of signals). Wait until #1065 is merged to finish this so we can regenerate gold receiving files with both sets of changes included.~

### Changelog
- `aggregate.R` and `contingency_aggregate.R`: remove logic that allows indicators to select whether or not to use mixing.
- `contingency_utils.R`, `binary.R`, and `count.R`: stop setting `skip_mixing` flag for indicators.
- Update tests.
- Regenerate gold receiving files, since `val`s and thus `se`s are calculated differently for a subset of signals.

### Fixes 
Closes #764.


capnrefsmmat:
Looks good; will officially approve once the tests are ready after the other merge.

nmdefries:
This change is going to be added to #1065 so the backfill includes both the standard error definition update and the weight-mixing update.

nmdefries:
@capnrefsmmat Test files are updated, so this is ready to look at again.

capnrefsmmat (approved):
Excellent

PR #1373, opened 2021-11-19
Stop sirCAL from complaining about missing survey `received_2_vaccine_doses`

nmdefries:
### Description
Item V2 was removed from the survey as of Nov 8, so stop sirCAL from checking if derived signals are up-to-date.

### Changelog
- sirCAL ansible template params and package template params.


krivard (approved):
👍

Issue #1374, opened 2021-11-22
CHNG Flu Signal


rumackaaron:
Analog of other chng signal using J09*-J11* (confirmed flu) outpatient counts as the numerator. Fulfills second bullet point of [Slack](https://delphi-org.slack.com/archives/C017FLA1G72/p1637273065011900) message.

**Data details**
 
Raw data comes from healthcare claims that Change processes and has use rights for. Change automatically runs several Delphi-specified SQL queries daily, including the number of outpatient claims with principal diagnosis code J09*-J11* per day per county. Counts between 1-3 are censored as "3 or less" which we treat as 1. The aggregated counts are placed on an SFTP server from which we can pull them.

The signal is the ratio of flu counts to total counts, with the same smoothing, weekday adjustments, and geographic pooling (megacounties) as the other chng signals.

PR #1375, opened 2021-11-24
CHNG Flu Signal

rumackaaron:
### Description
Add outpatient CHNG flu signal. This signal is analogous to the cli and covid signals already produced from CHNG data. The numerator is ICD codes J09*-J11* and we have already been receiving these aggregated counts from CHNG.

### Changelog
- `load_data.py` - Use `load_flu_data()` to create Pandas df with numerator specified as above and denominator as total outpatient visits.
- Other files were changed as necessary to create the signal but no modifications to the smoothing, adjustments, etc. were made.

### Fixes 
- Implements #1374

### Notes
The pseudocount addition creates a weekday problem. Over the summer, numerators are basically zero for several states while the denominators are smaller on weekends. Adding the pseudocount of 0.5 to the numerator means that there is a spike in the ratio on weekends because the signal value on a weekend is 0.5/(weekend total) and on a weekday is 0.5/(weekday total). This problem is present whether or not weekday adjustment is done, since the weekday adjustment only corrects the numerators which will remain at 0 in these states during the summer.


changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -93,3 +93,30 @@ def download_cli(filedate, out_path, ftp_conn):
     finally:
         if client:
             client.close()
+
+
+def download_flu(filedate, out_path, ftp_conn):
```
krivard, line 98: What's the difference between download_flu, download_covid, and download_cli? If they're identical I'd rather we not copy-paste.

krivard (changes requested):
Missing tests; also possible unnecessary code duplication

rumackaaron:
Unduplicated code and added tests

changehc/delphi_changehc/download_ftp_files.py
```diff
@@ -42,34 +42,7 @@ def get_files_from_dir(sftp, filedate, out_path):
 
 
 def download_covid(filedate, out_path, ftp_conn):
```
krivard, line 44:
```suggestion
def download_counts(filedate, out_path, ftp_conn):
```

changehc/delphi_changehc/run.py
```diff
@@ -14,8 +14,8 @@
 from delphi_utils import get_structured_logger
 
 # first party
-from .download_ftp_files import download_covid, download_cli
-from .load_data import load_combined_data, load_cli_data
+from .download_ftp_files import download_covid
```
krivard, line 17:
```suggestion
from .download_ftp_files import download_counts
```

changehc/delphi_changehc/run.py
```diff
@@ -26,10 +26,13 @@ def retrieve_files(params, filedate, logger):
...
-            download_covid(filedate, params["indicator"]["input_cache_dir"], params["indicator"]["ftp_conn"])
-        if "cli" in params["indicator"]["types"]:
-            download_cli(filedate, params["indicator"]["input_cache_dir"], params["indicator"]["ftp_conn"])
+        #if "covid" in params["indicator"]["types"]:
+        #    download_covid(filedate, params["indicator"]["input_cache_dir"], params["indicator"]["ftp_conn"])
+        #if "cli" in params["indicator"]["types"]:
+        #    download_cli(filedate, params["indicator"]["input_cache_dir"], params["indicator"]["ftp_conn"])
+        #if "flu" in params["indicator"]["types"]:
+        #    download_flu(filedate, params["indicator"]["input_cache_dir"], params["indicator"]["ftp_conn"])
+        download_covid(filedate, params["indicator"]["input_cache_dir"], params["indicator"]["ftp_conn"])
```
krivard, line 35:
```suggestion
        download_counts(filedate, params["indicator"]["input_cache_dir"], params["indicator"]["ftp_conn"])
```

krivard (approved):
👍 

I'm changing the function name but only because I'm insufferable

PR #1379, opened 2021-11-29
Suppress too-recent smoothed weighted indicators

nmdefries:
### Description
Prevent smoothed indicators from being reported if we don't have the most recent days of data. This prevents 7-day averages from being reported for only 6 days, since weights from FB are delayed a couple days.

### Changelog
- `aggregate.R`: skip calculation if most-recent weights are earlier than target date
- `weights.R`: return date of latest weights file along with the weights
- `run.R` and `contingency_run.R`: update to work with `weights.R` change.
- Update tests to work with `weights.R` change.


facebook/delphiFacebook/R/aggregate.R
```diff
@@ -179,6 +179,15 @@ summarize_indicators_day <- function(day_df, indicators, target_day, geo_level,
       var_weight <- indicators$var_weight[row]
       compute_fn <- indicators$compute_fn[[row]]
 
+      # Prevent smoothed weighted signals from being reported for dates after
+      # the latest available weight data.
+      if (target_day > params$latest_weight_date &&
+          indicators$smooth_days[row] == 6 &&
+          indicators$var_weight[row] == "weight") {
```
capnrefsmmat, line 186: Hmm, let me try to understand these conditions. If `indicators$smooth_days[row] == 1` and weights are unavailable for that day, there will be no observations with non-NA weight, and hence no data reported for that day. That's fine and good. If `indicators$smooth_days[row] > 1`, we could have ≥1 days of data without weights but still have at least some observations with weights, allowing us to report data.

It seems like more robust conditions would hence be

```suggestion
      if (target_day > params$latest_weight_date &&
          indicators$smooth_days[row] > 1 &&
          indicators$var_weight[row] != "weight_unif") {
```

to account for other types of smoothing and other names for weights. Right?

capnrefsmmat (approved):

PR #1380, opened 2021-11-29
Drop indicator-combination from sirCAL checks

krivard:
### Description
`indicator-combination` pipeline was stopped on November 18; this PR tells sirCAL to stop checking it.

### Changelog
- dev and production params for sirCAL: drop entry for indicator-combination


qx-teo (approved):
👍

PR #1395, opened 2021-12-06
Add dev and release processes

krivard:
### Description
I don't think we had this written down anywhere and we really should have it in case I run away and join the circus

### Changelog
- README: new sections for howto do dev and release


korlaxxalrok (approved):
Looks good 👍

PR #1401, opened 2021-12-08
CHNG ILI Signal

### LINKED ISSUE #1400: CHNG ILI Signal
Analog of other chng signals using J09*-J11* (confirmed flu) and {J22, B34.9} (flu like) outpatient counts as the numerator. See [Slack](https://delphi-org.slack.com/archives/C017FLA1G72/p1638925728023500) message.

Data details

Raw data comes from healthcare claims that Change processes and has use rights for. Change automatically runs several Delphi-specified SQL queries daily, including the number of outpatient claims with principal diagnosis code J09*-J11* per day per county and number of outpatient claims with principal diagnosis code in {J22, B34.9}. Counts between 1-3 are censored as "3 or less" which we treat as 1. The aggregated counts are placed on an SFTP server from which we can pull them.

The signal is the ratio of ili counts to total counts, with the same smoothing, weekday adjustments, and geographic pooling (megacounties) as the other chng signals.

rumackaaron:
### Description
Produces new signal as described in #1400 

### Changelog
- `load_data.py`: load ili data from Flu and Flu_Like columns
- Other small modifications as necessary

### Fixes 
- Fixes #1400

### DIFF
--- a/ansible/templates/changehc-params-prod.json.j2
+++ b/ansible/templates/changehc-params-prod.json.j2
@@ -22,7 +22,7 @@
     "parallel": false,
     "geos": ["state", "msa", "hrr", "county", "hhs", "nation"],
     "weekday": [true, false],
-    "types": ["covid","cli","flu"],
+    "types": ["covid","cli","flu","ili"],
     "wip_signal": "",
     "ftp_conn": {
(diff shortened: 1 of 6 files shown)


nolangormley (approved):
Looks good!

PR #1411, opened 2021-12-09
[CTIS] Fix `latest_weight_date` regex input so date isn't missing

nmdefries:
### Description
Apply regex that gets weight file date to weight filename only. The regex returns `NA` if given the whole path. Followup to #1379.

### Changelog
- `weights.R::join_weights()`: apply regex to filename only.

### Fixes
When `latest_weight_date` is missing, the pipeline generates too-recent smoothed weighted indicators since [the condition](https://github.com/cmu-delphi/covidcast-indicators/blob/a2004f38c3f70bdbdb65ddc5706cd8f9413a9ca6/facebook/delphiFacebook/R/aggregate.R#L184-L186) to suppress too-recent indicators never triggers.


krivard (approved):
👍

PR #1366, opened 2021-11-18
Extend covid-act-now max indicator lag to 9 in sir-CAL params

krivard:
### Description
CDC has switched their reporting of total test volume to lag (which we obtain via COVID Act Now) by an additional 3-4 days beyond the previous 5 day lag. This PR updates the Sir CAL config to match.

### Changelog
Itemize code/test/documentation changes and files added/removed.
- production params
- dev params

### Fixes 
- Fixes series of sirCAL alerts in #system-monitoring

PR #1415, opened 2021-12-10
Change CAN max lag to 9

qx-teo:
### Description
Following similar update in sirCAL (See #1366): adjusting covid-act-now max indicator lag to 9 since the CDC increases lag on their end

### Changelog
- params.json.template files


krivard (approved):
👍

PR #1419, opened 2021-12-11
Remove artificial day shift in labeling

mariajahja:
### Description
Match config behavior in [doctor-visits](https://github.com/cmu-delphi/covidcast-indicators/blob/main/doctor_visits/delphi_doctor_visits/config.py) where `DAY_SHIFT` is set to 0. 

### Changelog
- delphi_claims_hosp/config.py

### DIFF
--- a/claims_hosp/delphi_claims_hosp/config.py
+++ b/claims_hosp/delphi_claims_hosp/config.py
@@ -3,7 +3,7 @@ This file contains configuration variables used to generate the claims-based Hos
 
 Author: Maria Jahja
 Created: 2020-06-01
-Modified: 2020-09-27
+Modified: 2021-12-11
 
 """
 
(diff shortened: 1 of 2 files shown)


claims_hosp/delphi_claims_hosp/config.py
```diff
@@ -3,7 +3,7 @@
 
 Author: Maria Jahja
 Created: 2020-06-01
-Modified: 2020-09-27
+Modified: 2021-12-11
```
chinandrew, line 6: is there a reason we're not relying on git to track this?

krivard (approved):
👍

claims_hosp/delphi_claims_hosp/config.py
krivard: Possible reason: production deployments are stripped of all .git metadata, so markup like this would be the only way to verify which changes are in effect. It's not a _good_ reason (perennially backburnered is having deployment add a VERSION file with the gitref, which would be better) but it's better than nothing

chinandrew (approved):
:shipit: :shipit: :shipit:

mariajahja:
@krivard Just a note that the server running the daily generation has been updated.

Issue #1406, opened 2021-12-09
National and state-level google-symptoms signals are on different scales


ryantibs:
The COVID Symptom Searches on Google signal appears to be on the scale of roughly 0-100 at the national level and 0-1 at the state level. This results in very weird looking plots on the dashboard (when looking at a state, the national curve is overlaid, and makes the state curve look like it's basically at zero for all time): 

![Screen Shot 2021-12-09 at 2 01 15 PM](https://user-images.githubusercontent.com/732541/145463405-af5c7198-0548-48f3-9d8f-79f40d7a2ff9.png)
![Screen Shot 2021-12-09 at 2 00 55 PM](https://user-images.githubusercontent.com/732541/145463411-4d300c2a-7d2a-4e32-b1b6-8d0a199d1d87.png)
 
@krivard @nmdefries @jingjtang: Is this just due to the way we're constructing the national signal (a bug)? Assuming so we should fix this ASAP because it's hurting the dashboard visualizations.

(Assuming not, we need to figure out a different way to visualize the national and state signals in tandem, and then I would transfer this issue over on the www-covidcast repo.).

krivard:
This is expected behavior: google symptoms is, roughly speaking, a count not a proportion. Nation is the sum of all states; each state is the sum of its counties, etc.

ryantibs:
It's not a count, because each time series is self-normalized.  From [their documentation](https://github.com/GoogleCloudPlatform/covid-19-open-data/blob/main/docs/table-search-trends.md#how-we-process-the-data):

-----
The data shows the relative popularity of symptoms in searches within a geographical region. To normalize and scale the daily and the weekly time series (processed separately), we do the following for each region:

1. First, we count the number of searches for each symptom in that region for that day/week.
2. Next, we divide this count by the total number of Search users in the region for that day/week to calculate relative popularity (which can be interpreted as the probability that a user in this region will search for the given symptom on that day/week). We refer to this ratio as the normalized popularity of a symptom.
3. We then find the maximum value of the normalized popularity across the entire published time range for that region, over all symptoms using the chosen time resolution (day/week). We scale this maximum value to 100. All the other values are mapped to proportionally smaller values (linear scaling) in the range 0-100.
4. Finally, we store the scaling factor and use it to scale values (for the same region and time resolution) in subsequent releases. In future updates, when a symptom popularity exceeds the previously-observed maximum value (found in step 3), the new scaled values may be larger than 100.

-----

[...]

PR #1423, opened 2021-12-13
Fix hhs national from google symptoms

jingjtang:
### Description
This PR is used to fix the daily report for HHS and National level.

The count values in Google Symptoms are not exactly counts, but have been pre-scaled by Google already. Such pre-processing makes the "counts" that we get from Google are in an arbitrary unit. Previously we have a simple aggregated (summation of the counts from states) from state level report to HHS and National level report which is incorrect. Now, we fix those mapping to make them match what we did for the geo mapping from county to HRR and MSA. 

Notice that, as what we did for the geo mapping from county to MSA and HRR, the finalized `counts` for higher geographical levels should be pretty small since we treat counties with no report as 0 report. 

The previous [PR ](https://github.com/cmu-delphi/covidcast-indicators/pull/1418) was mistakenly closed after renaming the branch. 

### Changelog
- `run.py` and `geo.py`
- `tests/test_geo.py`

### Fixes 
- Fixes [#1406](https://github.com/cmu-delphi/covidcast-indicators/issues/1406)


jingjtang:
Record the discussion related to either generating the HHS/national level report from state level or county level reports

```
krivard 1 hour ago
Do we want to use the state data to generate hhs and nation?
  
Member
@ryantibs ryantibs 1 hour ago
Thanks for catching this! I was about to raise the same issue 🙂

And yes---I don't see why we would generate it from counties provided that we have states in-hand (and even aggregating up from counties --> states is not going to be ideal since they won't recover states).
  
Member
Author
@jingjtang jingjtang 1 hour ago • 
edited 
No. We already know that at least for the current Google Symptoms (Anosmia and Ageusia only), we have only around ~100 counties available. Since we do a population weighted average, it should be more accurate to do the aggregation from county level. For example, we only have 2 counties available in state A, if we start from state level, we will add a weight based on the state population to the already normalized and scaled count from Google when doing the average.
  
Member
Author
@jingjtang jingjtang 1 hour ago
Or do you think the other way is more reasonable?
  
Member
@krivard krivard 1 hour ago
we only have two counties available in state A, but we have no idea how many counties go into Google's value for state A, right? if they're able to include more counties in their published state figures, shouldn't we make use of them?
 
👍
1
  
Member
Author
@jingjtang jingjtang 1 hour ago
Yeah that makes sense, I may make too far assumption about this. Since a lot of things are unknown to us, I think you and Ryan is correct. I will change those mappings from state level.
  
Member
@ryantibs ryantibs 44 minutes ago
I can see there are pros/cons either way, but I do think it's worth aggregating state signals to build national and HHS signals, for ultimately the following reason: state numbers are based on normalizing state counts, and there's a lot less censorship here than at the county level (where lots of counts are censored), so you'll be able to leverage a lot more data this way.
```

jingjtang:
Things to do for this fixing are:

[done by Sam] Change the visualization strategy for Google Symptoms. We need to carefully label that there should be no comparison between regions at the same geographical level and also between geographical levels.
[waiting for review] Fix the aggregation method (from simple summation to population-weighted average) for geo mapping from county to HHS and Nation.

google_symptoms/delphi_google_symptoms/geo.py
```diff
@@ -23,20 +23,30 @@ def generate_transition_matrix(geo_res):
...
+    if geo_res in ["hrr", "msa"]:
+        mapping_flag = "fips"
+        map_df = gmpr.get_crosswalk("fips", geo_res)
+        # Add population as weights
+        map_df = gmpr.add_population_column(map_df, "fips")
+    else:
+        mapping_flag = "state_id"
+        map_df = gmpr.get_crosswalk("state", "state")
+        map_df = gmpr.add_geocode(map_df, "state_code", "hhs")
+        map_df = gmpr.add_geocode(map_df, "state_code", "nation")
```
krivard, line 35: Would this work?
```suggestion
        map_df = gmpr.add_geocode(map_df, "state_code", geo_res)
```

krivard (approved):
This is tough to test since it depends on GeoMapper populations, but from what I can see it extends the HRR/MSA implementation but otherwise exactly matches it, which is what we wanted.

PR #1435, opened 2021-12-19
Fix pipeline when barrier column is all NAs part 2

nmdefries:
### Description
Followup to #1346. Set answer code 13 to NA *before* checking if col is all NA.

### Changelog
- `variables.R`

### Fixes 
In rare cases, `vaccine_barriers <- ifelse(vaccine_barriers == "13", NA, vaccine_barriers)` or `vaccine_barriers <- ifelse(input_data$V15b == "13", NA, input_data$V15b)` will make a column all NA when it wasn't before. This happens when the only non-NA responses have answer code "13".

### DIFF
--- a/facebook/delphiFacebook/R/variables.R
+++ b/facebook/delphiFacebook/R/variables.R
@@ -646,7 +646,9 @@ code_vaccines <- function(input_data, wave) {
     # If the entire column is NA, ifelse() results in a logical vector, not a
     # character vector, which confuses split_options; since the result should be
     # NA anyway
-    vaccine_barriers <- ifelse(vaccine_barriers == "13", NA_character_, vaccine_barriers)
+    vaccine_barriers <- as.character(
+      ifelse(vaccine_barriers == "13", NA_character_, vaccine_barriers)
+    )
     if (any(!is.na(vaccine_barriers))) {
       vaccine_barriers <- split_options(vaccine_barriers)
     }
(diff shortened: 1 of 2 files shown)


nmdefries:
@capnrefsmmat FYI. I'm thinking the more robust fix is to change `split_options` to return a list of `NA` if given a column of all `NA`. I don't think that messes up any of our other logic, but that'll need a more careful look and more code changes, so I'll do it in a separate PR.

facebook/delphiFacebook/R/variables.R
```diff
@@ -646,8 +646,8 @@ code_vaccines <- function(input_data, wave) {
     # If the entire column is NA, ifelse() results in a logical vector, not a
     # character vector, which confuses split_options; since the result should be
     # NA anyway
+    vaccine_barriers <- ifelse(vaccine_barriers == "13", NA_character_, vaccine_barriers)
     if (any(!is.na(vaccine_barriers))) {
-      vaccine_barriers <- ifelse(vaccine_barriers == "13", NA_character_, vaccine_barriers)
```
krivard, line 650: if the entire column is NA, and (thus) ifelse() results in a logical vector, how does that affect `any(!is.na(vaccine_barriers))`?

nmdefries: `is.na` appears to have the same behavior for logical `NA` and character missing (`NA_character_`), but I'll explicitly cast to character to be safe. The downstream `is_selected` also appears to work correctly with logical `NA` but expects strings.

krivard (approved):
👍

PR #1439, opened 2021-12-28
Combine V15a/c logic properly for survey Wave 12

nmdefries:
### Description
Since items V15a and V15c won't be available at the same time for a given data file, combine their `if` blocks into one so that only one of the conditions is triggered per data file.

### Changelog
- `variables.R`: Change `if` blocks that depend on item V15c (`if ( all(c("V15c", "V15b") %in% names(input_data)) )` and `if ( "V15c" %in% names(input_data) )`) to `else if`.

### Fixes 
error introduced in #1303.

When item V15a exists, as in Wave 11 data, `vaccine_barrier_*` and `vaccine_barrier_*_has` fields were created in an initial `if` block. A second `if` block checked for the existence of item V15c (replacing V15a as of Wave 12) and if not found, overwrote those fields with `NA`. Since the existence of V15a and V15c were checked separately, the V15c `else` condition was always triggered, resulting in the `vaccine_barrier_*` and `vaccine_barrier_*_has` fields always being completely missing for pre-Wave 12 survey versions and the relevant indicators not being reported.

We'll need to reissue the `smoothed_wvaccine_barrier_*` and `smoothed_wvaccine_barrier_*_has` indicators back to Dec 12.

### DIFF
--- a/facebook/delphiFacebook/R/variables.R
+++ b/facebook/delphiFacebook/R/variables.R
@@ -665,22 +665,10 @@ code_vaccines <- function(input_data, wave) {
     input_data$v_vaccine_barrier_time <- is_selected(vaccine_barriers, "10")
     input_data$v_vaccine_barrier_type <- is_selected(vaccine_barriers, "12")
     input_data$v_vaccine_barrier_none <- is_selected(vaccine_barriers, "11")
-  } else {
-    input_data$v_vaccine_barrier_eligible <- NA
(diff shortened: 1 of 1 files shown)


capnrefsmmat (approved):
Good catch. I think this makes sense now, although it may be good to add more explicit code comments to explain which items are in which waves:

| Wave | V15a | V15b | V15c |
| --- | --- | --- | --- |
| 11 | Yes | Yes | No|
| 12 | No | Yes | Yes |

Issue #1388, opened 2021-12-01
delphi-utils dual-year geocoding files are not being exported in the package


krivard:
While testing the jhu indicator on staging I discovered that it doesn't actually run:

```
[indicators@app-mono-dev-01 jhu]$ env/bin/python -m delphi_jhu
Traceback (most recent call last):
  File "/home/indicators/.pyenv/versions/3.8.2/lib/python3.8/runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/indicators/.pyenv/versions/3.8.2/lib/python3.8/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/mnt/data/indicators/runtime/jhu/delphi_jhu/__main__.py", line 12, in <module>
    run_module(read_params())  # pragma: no cover
  File "/mnt/data/indicators/runtime/jhu/delphi_jhu/run.py", line 91, in run_module
    gmpr = GeoMapper()
  File "/mnt/data/indicators/runtime/jhu/env/lib/python3.8/site-packages/delphi_utils/geomap.py", line 136, in __init__
    self._load_crosswalk_from_file(from_code,
  File "/mnt/data/indicators/runtime/jhu/env/lib/python3.8/site-packages/delphi_utils/geomap.py", line 145, in _load_crosswalk_from_file
    stream = pkg_resources.resource_stream(__name__, data_path)
  File "/mnt/data/indicators/runtime/jhu/env/lib/python3.8/site-packages/pkg_resources/__init__.py", line 1150, in resource_stream
    return get_provider(package_or_requirement).get_resource_stream(
  File "/mnt/data/indicators/runtime/jhu/env/lib/python3.8/site-packages/pkg_resources/__init__.py", line 1613, in get_resource_stream
    return open(self._fn(self.module_path, resource_name), 'rb')
FileNotFoundError: [Errno 2] No such file or directory: '/mnt/data/indicators/runtime/jhu/env/lib/python3.8/site-packages/delphi_utils/data/2020/zip_fips_table.csv'
```

However I couldn't replicate it in my dev environment until I swapped the editable-install command used by `make test` for the uneditable-install command used by [build-and-package.sh](https://github.com/cmu-delphi/covidcast-indicators/blob/main/jenkins/build-and-package.sh).

Looks like we didn't update the package spec when we added in the 2010/2020 year split:

[...]

PR #1444, opened 2022-01-03
Add make install-ci target to install without editable mode

### LINKED ISSUE #1443: Consider updating CI to use nonlocal delphi-utils installation (possibly through a separate make target) -- attn @chinandrew

chinandrew:
### Description
See #1388 and #1443

### Changelog
- Add a `make install-ci` target that installs everything in the standard, noneditable mode. Not sure `install-ci` is the best naming so open to suggestions.
- Fix nowcast directory which was missing inits
- Fix the find/grep statement in the makefile to only return 1 result. If subdirs are present in an indicator, it returns multiple results and causes issues with `make test`

### Fixes 
- Fixes #1443


chinandrew:
looks like i gotta fix some nocast linting which was previously missed, will do that in a separate pr

krivard (approved):
👍

PR #1445, opened 2022-01-03
Fix nowcast linting

chinandrew:
### Description
Nowcast folder wasnt being linted properly, which I discovered in #1444. This PR fixes that.

Blocking #1444
### Changelog
- Allow 1 latter capital variables since they are taken from mathematical notation
- Remove ground_truth.py file which is super old and needs to be rewritten anyway if we want to use it
- Remove unused imports/variables.

### Fixes 
- Fixes #(issue)


mariajahja (approved):
lgtm. Thanks!

PR #1447, opened 2022-01-05
[CTIS] Codebook improvements

nmdefries:
### Description
Since experimental wave 12 is referred to as "12.5", reading it as an integer causes a parsing problem that corrupts the codebook.

Tell `readr::write_excel_csv` to use minimal quotes so Wave 12.5 -> Wave 12 codebook diff is smaller.

Fix logic error in filtering `w12_treatment` field so that `w12_treatment` is included for only experimental Wave 12 ("12.5").

### Changelog
- `generate-codebook.R`


capnrefsmmat (approved):

nmdefries:

PR #1454, opened 2022-01-07
Convert safegraph_patterns to use the Places API instead of S3

### LINKED ISSUE #1451: Switch Safegraph indicator to use the Places API
Safegraph is shutting off S3 access to their data on 2022-01-31 in favor of their Places API.

The [weekly patterns endpoint in the Places API as-documented](https://docs.safegraph.com/reference/weekly-patterns-1) does not support our needs. There is a beta endpoint called `AcademicBulkWeeklyPatterns` that Safegraph have set up for us which may be a suitable replacement.

I have some draft code that uses it, but the downsides are:
* it is too slow to use single-threaded
* partitioning the placekey-space (for multi-threaded access) is both nonobvious (they're ordered but not evenly distributed) and nonstatic (the set of available placekeys changes monthly, substantially, but not entirely)
* using 5 threads was encountering 504 Gateway Timeout errors last October when I was testing. our safegraph contact was going to look into it, but never responded and then i got trapped in JHU-patch-town.

I will revive the draft and post it in a branch, but someone (probably not me) will need to complete the changes for deployment in advance of the 31st.

krivard:
### Description

First pass complete, needs finishing.

* [x] Multithreading based on a rough empirical partitioning of placekeys
* [x] Minimal refactor to achieve running status
* [ ] Still needs support code for periodically recomputing partition signposts
* [ ] Multiple-date handling has been disabled for now. May need to reinstate later

### Changelog
- new file pull.py
- refactored process.py to use new pull machinery
- refactored run.py to use new process machinery and remove S3 sync

### Fixes 
- Fixes #1451


ansible/templates/safegraph_patterns-params-prod.json.j2
```diff
@@ -7,6 +7,7 @@
     "static_file_dir": "./static",
     "raw_data_dir": "/common/safegraph",
     "n_core": 12,
+    "apikey": "{{ safegraph_apikey }}"
```
krivard, line 10: This should fix the failing JSON check
```suggestion
    "apikey": "{{ safegraph_apikey }}",
```

Issue #1457, opened 2022-01-10
[Validator] Extract source subdivisions from /covidcast/meta instead of hard-coding


krivard:
Validator currently hard-codes the mapping from sources to source subdivisions:

https://github.com/cmu-delphi/covidcast-indicators/blob/716f7330cad30d8fde5051e6ff779a2345539127/_delphi_utils_python/delphi_utils/validator/datafetcher.py#L114-L119

This means whenever we change the mapping in the spreadsheet (eg for https://github.com/cmu-delphi/delphi-epidata/issues/800), we would need to remember the extra step of updating this hardcoded copy in validator. That is not ideal.

Is this mapping exposed as part of `/covidcast/meta`? If so, we should have validator look it up there instead. If not, we should add an endpoint to delphi-epidata `/covidcast/*` to surface it.

krivard:
`mata` with no arguments gives you `db_source` and then `source` for all the member signals, which is exactly this mapping.

Issue #1456, opened 2022-01-10
Remove network call from validator test of `get_geo_signal_combos`


krivard:
Currently the validator function `get_geo_signal_combos` makes two network calls: one using `covidcast.metadata` to fetch from `/covidcast_meta` and one using `requests` to fetch from `/covidcast/meta` (which has additional usefulness).

The unit tests only mock out `covidcast.metadata`, which means that changes to the output of `/covidcast/meta` can cause tests to fail (eg [for this unrelated change to the surveys pipeline](https://github.com/cmu-delphi/covidcast-indicators/runs/4763242791?check_suite_focus=true), which doesn't use the validator but is nevertheless blocked by failing tests)

We should find a way to mock out the `requests` call as well.

PR #1458, opened 2022-01-10
Patch validator for chng subdivision mapping change

krivard:
### Description
Tests are currently failing for validator after changing the source subdivision mapping for the changehc indicator. This PR installs a quick patch but does not solve the underlying problems, which are described in more detail in #1456 and #1457.

### Changelog
- test_datafetcher.py: Add todo
- datafetcher.py: Drop hardcoded mappings for `chng` source and add todo

### Fixes 
- Fixes failing builds


qx-teo (approved):
Looks good!

Issue #1354, opened 2021-11-08
New indicator: CDC COVID-19 Community Profile Report


krivard:
The CDC publishes a community profile report which, among other things, includes total test volume. Since COVID Act Now will soon stop providing total test volume, this profile report seems to be a reasonable replacement.

**Data details**
 
[COVID-19 Community Profile Report](https://healthdata.gov/Health/COVID-19-Community-Profile-Report/gqxm-d9w9) on healthdata.gov

This report is updated *daily*. Most metrics count over a *rolling 7-day period*. This is suitable for producing covidcast signals which are 7-day averages, updated daily.

It includes the following metrics at the state level:
* FEMA region
* Population
* Cases and deaths: last 7 days (count, count per 100k population, % of national), % change from previous week, previous week (count, count per 100k population, % of national), cumulative (count, count per 100k population)
* Number of days of downward case trajectory
* Community transmission level (current week, previous week)
* Nucleic Acid Amplification Tests (NAAT; includes things like PCR): this week, previous week, and % or absolute change: test positivity (all and age-stratified), testing volume (count, count per 100k population), median test result latency, % of test results received in 3 or fewer days
* Emergency Department Visits: this week and absolute change from previous week: % of visits with diagnosed COVID-19 (all and age-stratified)
* Hospital utilization: this week and % or absolute change from previous week: COVID admissions (count, % of inpatient beds, and age-stratified count), % inpatient beds occupied (all and COVID-19), % staffed adult ICU beds occupied (all and COVID-19), % of ventilators in use (all and COVID-19)
* Hospital supply and staffing: this week and absolute change from previous week:  % hospitals unable to maintain X (N95 respirators, surgical masks, eye protection, single-use gowns, exam gloves), % hospitals with current staffing shortage
* Therapeutics: Monoclonal antibody courses administered (this week and % change from previous week), monoclonal antibody courses on hand
* Reporting and capacity: Number of CCNs, X among hospitals reporting this week (total inpatient beds, total staffed adult ICU beds, total ventilators), % CCNs reporting anything at least once this week, % CCNs reporting everything at least once this week, % CCNs ever reporting, % CCNs reporting COVID-19 admissions this week
* COVID-19 vaccinations: this week and % change from last week: Doses administered (count, count per 100k population), Booster doses administered
[...]

PR #1466, opened 2022-01-13
Add hospital admissions to CPR indicator

nmdefries:
### Description
Add hospital admissions (pediatric + adult) to the Community Profile Report indicator. Generalize some package logic. Add tests to CI.

### Changelog
- `constants.py` - add info for hosp admissions signal. Add more fields to `SIGNALS` subdicts.
- `pull.py`
    - Define regex for hosp header. Have `from_header` check if a match exists for each header regex.
    - Extend `retain_header` and `skip_overheader` to keep hosp header/overheader.
    - Allow changes to `DatasetTimes` attributes so can combine test and hosp `DatasetTimes` objects when necessary.
    - Switch `nation_from_state` to use `state_id` instead of `state_code`. CPR uses state appreviations not FIPS codes, so `state_code` was causing nation dfs to be empty.
- `test_pull.py`
- `.github/workflows/python-ci.yml`
- `ansible/templates/dsew_community_profile-prod.json.j2`

### Fixes
Partially addresses #1354.

### DIFF
--- a/dsew_community_profile/delphi_dsew_community_profile/pull.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/pull.py
@@ -244,6 +244,16 @@ class Dataset:
         ]
 
         for sig in SIGNALS:
+            # Hospital admissions not available at the county or CBSA level prior to Jan 8, 2021.
+            if (sheet.level == "msa" or sheet.level == "county") \
+                and self.publish_date < datetime.date(2021, 1, 8) \
+                and sig == "confirmed covid-19 admissions":
+                self.dfs[(sheet.level, sig)] = pd.DataFrame(
+                        columns = ["geo_id", "timestamp", "val", \
+                            "se", "sample_size", "publish_date"]
+                    )
+                continue
+
             sig_select = [s for s in select if s[-1].find(sig) >= 0]
             assert len(sig_select) > 0, \
                 f"No {sig} in any of {select}\n\nAll headers:\n{NEWLINE.join(list(df.columns))}"
@@ -349,7 +359,7 @@ def fetch_new_reports(params, logger=None):
     # collect like signals together, keeping most recent publish date
     ret = {}
     for sig, lst in datasets.items():
-        ret[sig] = pd.concat(
+        latest_sig_df = pd.concat(
             lst
         ).groupby(
             "timestamp"
@@ -359,8 +369,8 @@ def fetch_new_reports(params, logger=None):
             "publish_date", axis=1
         )
 
-        if ret[sig].index.names and ret[sig].index.names[0] == "timestamp":
-            ret[sig] = ret[sig].droplevel("timestamp")
+        if len(latest_sig_df.index) > 0:
+            ret[sig] = latest_sig_df.reset_index(drop=True)
 
     # add nation from state
     geomapper = GeoMapper()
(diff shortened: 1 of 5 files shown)


nmdefries:
I spot-checked nation and NY state values for "total", "positivity", and hosp admissions for Jan 3 2022, Jan 7 2022, and Jan 8 2022, respectively, against manual calculations using the Jan 10 2022 report. Nation "positivity" was not close (0.0195 from pipeline vs 0.295 in the report, with no population-weighting in my hand calculation) -- looking at that more closely;  all other values matched exactly.

nmdefries:
```
wm = lambda x: np.average(x, weights=df.loc[x.index, "weight"])
df.dropna(subset = ["val"]).groupby("timestamp").agg(direct_weighted_mean=("val", wm))
```
gives similar values to my hand calculation, 0.312 vs 0.295.

We're ending up with multiple entries per state (with different values) for some dates, presumably from different publish dates. I'm able to fix that by normalizing the population weights by timestamp instead of overall, but that means the dates with multiple entries per state will really be averages over different publishing dates, instead of reflecting the most up-to-date values.

"Total" and hosp admissions values will also be ~doubled for those dates. So we'd like to remove the extra entries before the state-to-nation aggregation step.

krivard:
> We're ending up with multiple entries per state (with different values) for some dates, **presumably from different publish dates.**

In theory each publish date (each xls file) encodes a single reference date for hospital admissions, so this should not be possible.

For testing, we get two reference dates per xls file; the estimate from the most-recent xls file is the more accurate one.

Maybe we need to refactor this indicator to only process one file at a time, so that one file = one issue?

nmdefries:
Fixed that issue by keeping track of publish date in each df and using it to remove older values when combining across files/publish dates.

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -261,13 +320,13 @@ def download_and_parse(listing, logger):
 
 def nation_from_state(df, sig, geomapper):
     """Compute nation level from state df."""
-    if SIGNALS[sig]: # true if sig is a rate
-        df = geomapper.add_population_column(df, "state_code") \
+    if SIGNALS[sig]["is_rate"]: # true if sig is a rate
+        df = geomapper.add_population_column(df, "state_id") \
                       .rename(columns={"population":"weight"})
-        df.weight = df.weight / df.weight.sum()
+        df.weight = df.weight / df.weight.sum() * len(df.timestamp.unique())
```
krivard, line 326: Is this still needed?

nmdefries: With regards to the issue caused by multiple publish dates? Yes, this was a separate issue causing incorrect nation "positivity" values. When processing multiple days of data at a time, just summing the weights makes the normalization denominator too big. Then the normalized weights and the resulting nation values are too small.

krivard: wouldn't we want to groupby timestamp and then sum in that case, instead of multiplying uniformly? or are we always guaranteed the same number of rows per unique timestamp?

nmdefries: I'll switch to a groupby approach to be more robust.

krivard (approved):
Looks good. I'm running now on all dates just to make sure we didn't miss any sneaky header changes, and will merge once that's successful

nmdefries:
That should be fixed now, and a run through all dates looks fine.

Issue #1452, opened 2022-01-07
Update Quidel Covid  pipeline


jingjtang:
We should have several updates to Quidel Covid Test data at the current stage of COVID. 
- Considering adding age group specific signals for Quidel Covid
- Add megaounty to Quidel Covid
- Add another minimum report threshold for Quidel Covid
- All the three items above can get historical reports especially the third one, we now have deletion support in our system
- Roni (maybe and I) will meet with Quidel next week? to see whether we can get variant information from Quidel (Like Flu)

krivard:
[Ryan, Slack, 2022-01-10](https://delphi-org.slack.com/archives/C011CFX7TBR/p1641831874004000):
> Metrics:
> * [ ] for each availability threshold: 50, 40, 30, 20. (one table per threshold)
> * [ ] for each age group, and also for all-ages signal
> * [ ] % of locations available: average over time, for each county/state levels
> * [ ] correlations: timewise, average over locations, for each county/state levels

krivard:
* [Live doc](https://docs.google.com/document/d/1WP4VJEois4D8qfeVeKCOJ9c4D7KHscDHzhQZgmuVvcA/edit#heading=h.kxccq395w8od)
* PDF version: [Quidel_AgeGroup_Signals.pdf](https://github.com/cmu-delphi/covidcast-indicators/files/7866184/Quidel_AgeGroup_Signals.pdf), generated 2022-01-13 16:40

Conclusions:
> * stick with the current 50 availability threshold for 1) consistency with what we have published 2) statistical consideration 3) not big difference if lower the threshold
> * Use the age group breakdown: 0-4, 5-17, 18-49, 50-64, 65+

krivard:
Deletions are complete, but checks are showing suspicious counts.

Here's what was done:
1. Grab `issue,source,time_value,geo_type,signal,geo_value,value,stderr,sample_size` for quidel `covid_ag_smoothed_pct_positive` with issue<20220208 from MariaDB and store in a CSV file
2. Convert CSV file to batch issue format as "old data"
3. Reset the Quidel input cache. For each issue in the old data, in order chronologically, run the v0.3.2 (omicron) quidel indicator into a new export directory for that issue. Run the filesystem archiver, using the equivalent issue of the old data as the archivediffer cache. Now the contents of the export directory contain deletion annotations for any regions that were in the old data but are no longer generated by the new pipeline.
[...]

Issue #458, opened 2020-11-07
Add rest-of-state estimates for Quidel


jingjtang:
In currently pipeline for Quidel, we do not report counties with sample size smaller than a certain threshold (50 for both raw and smoothed signals). As Ryan suggested, we can add rest-of-state estimates for Quidel using data from those counties that are not individually reported.

PR #1467, opened 2022-01-14
Update quidel covidtest (Add Age Groups Signals, Add rest-of-state reports)

jingjtang:
### Description
- Add age group specific signals for Quidel Covid
   - Use the age group breakdown: 0-4, 5-17, 18-49, 50-64, 65+, 0-17(supergroup)
   - Stick with the current 50 availability threshold for 1) consistency with what we have published 2) statistical consideration 3) not big difference if lower the threshold
- Add megacountyies to Quidel Covid
  - Megacounties are added before the availability threshold checking. 
  - For raw signals, counties with counties <= 50 will be merged together as megacounties
  - For smoothed signals, counties with counties <= 25 will be merged together as megacounties
  - After getting the megacounties, if the count for a megacounty does not pass the availability threshold, we will not report anything for this megacounty.
- Live docs [here](https://docs.google.com/document/d/1WP4VJEois4D8qfeVeKCOJ9c4D7KHscDHzhQZgmuVvcA/edit?usp=sharing) for availability checking and correlation analysis

### Changelog
- `constants.py`, `pull.py`, `run.py`, `geo_map.py`, `generate_sensor.py`
- `tests/test_constants.py`, `tests/test_pull.py`, `tests/test_run.py`, `tests/test_geo_map.py`, `tests/test_generate_sensor.py`, `tests/test_data`

### Fixes 
- Fixes [#1452](https://github.com/cmu-delphi/covidcast-indicators/issues/1452), [#458](https://github.com/cmu-delphi/covidcast-indicators/issues/458)


jingjtang:
We need new signals names for age-group specific siganls.
The current ones:
- `covid_ag_raw_pct_positive_age_0to4`
- `covid_ag_raw_pct_positive_age_5to17`
- `covid_ag_raw_pct_positive_age_18to49`
- `covid_ag_raw_pct_positive_age_50to64`
- `covid_ag_raw_pct_positive_age_65toOlder`
- `covid_ag_raw_pct_positive`
- `covid_ag_smoothed_pct_positive_age_0to4`
- `covid_ag_smoothed_pct_positive_age_5to17`
- `covid_ag_smoothed_pct_positive_age_18to49`
- `covid_ag_smoothed_pct_positive_age_50to64`
- `covid_ag_smoothed_pct_positive_age_65toOlder`
- `covid_ag_smoothed_pct_positive`

Need @RoniRos @ryantibs @krivard 's approval for the signal names.

krivard:
> We need new signals names for age-group specific siganls.
Let me find what we're using in covid_hosp; we should be consistent where possible

ryantibs:
Katie's last naming scheme: 👍 from me.

jingjtang:
As suggested by @RoniRos, we can add another two super age groups: 0-17; 18-64, so that we have information for those small age groups like 0-4 added but not bothered too much by the available threshold

RoniRos:
Katie’s last naming scheme LGTM, too.

RoniRos:
This suggestion was for our consideration.  By adding the super-groups of 0-17 and 18-64, we may have many counties for which Quidel age-specific signal is available.  Jingjing will calculate and share the # counties above threshold, so we can consider if it’s worth it.

jingjtang:
location_nums_over_time

The new available ability is shown here. I think those two super age groups are good to be added.

RoniRos:
I agree with Jingjing that these two supergroups could be useful, especially at the county level.
I’d like to know what Ryan and Kate think in terms of value vs. cost.

Btw, Jingjing, the State-level panels include also a coverage curve for “total” (all-ages), but the county-level panel does not.  Can you please add it at your convenience?  It’s a good indication of upper limit, and what is lost when breaking down by age-group.

ryantibs:
It's quite hard for me to tell based on the current plots whether the tradeoff is actually worth it. I suspect it may not be. The easiest & most direct way to tell would be to add up the availability of the subgroups within each combined group

Eg, just compare the availability of 0_4 + 5_17 to 0_17 (so just comparing two lines), and likewise for the case.

RoniRos:
Ryan: I’m not sure why you would want to add up the coverages.  I would think one would want to subtract the coverage of each sub-group from the coverage of its super-group, e.g. :


  *   How many counties that do not support a 0-4 signal, at least support a 0-17 signal.
And separately:

  *   How many counties that do not support a 5-17 signal, at least support a 0-17 signal.

ryantibs:
Yes sorry, my "+" was supposed to be an "OR".  So I think we are saying equivalent things.

RoniRos:
Sorry to be a pest, but I still don't understand.  Why would you want to do an OR, rather than add up the two differences I listed as an example?  If there is a county were neither 0-4 nor 5-17 is available, arguably it should count twice as much as a county where only one of them is unavailable.  No?

Anyway, what are the use cases you envision for these signals?

RoniRos:
One could argue that when, say 0-17 and 5-17 are available but 0-4 is not, one could derive an estimate of 0-4 from the other two, based on their proportion in the population (since we don't give them their proportion among the tests).  So the case when neither 0-4 nor 5-17 are available is by far the worst.

ryantibs:
Not a pest, I just wasn't reading/thinking carefully! Yes, you're right, your message is alluding to something different.

But we wouldn't be able to find out from your counts whether in a particular county we have what you call the "worst case" (neither 0-4 nor 5-17 are available), since it's just marginal counts, right? And yes I buy your argument. So I guess the following four numbers would be useful:

How many counties that do not support a 0-4 signal, support a 0-17 signal.
How many counties that do not support a 5-17 signal, support a 0-17 signal.
How many counties that do not support a 0-4 signal OR do not support a 5-17 signal, support a 0-17 signal.
How many counties that do not support a 0-4 signal AND do not support a 5-17 signal, support a 0-17 signal.

@jingjtang Can you do this, and do the same for the other category 18-64, so that we can get a sense of the value ad?

jingjtang:
@RoniRos We initially add availability checking for all-age groups [here](https://delphi-org.slack.com/archives/C011CFX7TBR/p1641421855006500?thread_ts=1641247387.001700&cid=C011CFX7TBR), but that curve for county only is removed later **ONLY** for a more clear looking at the bottom area for the age-specific signals.

RoniRos:
Ok.  
From the curves Jingjing published so far, it's clear there is a non-negligible number of counties "rescued" by the merging.  But we still don't have a clear formula to help us decide whether or not to include these merged groups.

I just browsed some of CDC's data and found many places were they present age breakdowns as 0-17,18-49,50-64,65+.  E.g. https://www.cdc.gov/coronavirus/2019-ncov/cases-updates/burden.html 

So my suggested is that we _at least_ include 0-17.

jingjtang:
@ryantibs
0_17 but not 0_4: Available for 0-17 but not available for 0-4
0_17 but not 5_17: Available for 0-17 but not available for 5-17
0_17 but not OR: Available for 0-17 but not available for either 0-4 or 5-17
0_17 but not AND: Available for 0-17 but not available for both 0-14 and 5-17

Similar ones for 18-64

comparison_with_supergroups

ryantibs:
@RoniRos @jingjtang Thanks!

I'm convinced we should add 0-17. We can just make it clear in our documentation that we also wanted to separate this out into two subcategories duet to the fact that 0-4 is unvaccinated population. This would be a clean explanation as to why there is overlap/nesting in the lowest 3 categories (0-4, 5-17, 0-17). 

I'm less convinced we should add 18-64. Rationale: if we rescue *any* counties here with this big bucket, then the signal we publish is probably going to be very similar to the all-ages signal. (Because I'm guessing this big bucket includes the most common ages from a marginal perspective, and if signals at 18-49 and 50-64 are each missing, then it must means we're just right below the threshold we set for availability, and 18-64 puts us right above).  

So I suggest we just publish 0-17, along with the existing breakdowns, and finish this off ASAP. However, not strongly opposed to also including 18-64, if Roni or somebody else feels differently. Thanks!

RoniRos:
I agree on all counts.

quidel_covidtest/delphi_quidel_covidtest/pull.py
```diff
@@ -186,6 +188,55 @@ def preprocess_new_data(start_date, end_date, params, test_mode, logger):
         overall_pos, on=["timestamp", "zip"], how="left"
         ).fillna(0).drop_duplicates()
 
+    # Compute Summary info for age groups
+    df["PatientAge"]= df["PatientAge"].fillna(-1)
```
krivard, line 192:
```suggestion
    df["PatientAge"] = df["PatientAge"].fillna(-1)
```

quidel_covidtest/delphi_quidel_covidtest/pull.py
```diff
@@ -186,6 +188,55 @@ def preprocess_new_data(start_date, end_date, params, test_mode, logger):
...
+    # Should match the suffixes of signal names
+    df["label"] = None
+    df.loc[df["PatientAge"] < 5, "label"] = "age_0_4"
+    df.loc[((df["PatientAge"] >= 5)) & (df["PatientAge"] < 18), "label"] = "age_5_17"
+    df.loc[((df["PatientAge"] >= 18)) & (df["PatientAge"] < 50), "label"] = "age_18_49"
+    df.loc[((df["PatientAge"] >= 50)) & (df["PatientAge"] < 65), "label"] = "age_50_64"
+    df.loc[(df["PatientAge"] >= 65), "label"] = "age_65plus"
+    df.loc[df["PatientAge"] == -1, "label"] = "NA"
+
+    for agegroup in AGE_GROUPS[1:]:
```
krivard, line 206:
```suggestion
    # exclude total
    for agegroup in AGE_GROUPS[1:]:
```

quidel_covidtest/tests/test_geo_maps.py
```diff
@@ -1,29 +1,52 @@
...
-        assert set(new_df["fips"].values) == set(['25027', '53011', '48439'])
+        assert set(new_df["fips"].values) == set(['25027', '25013', '25015', '53011', '48439'])
         assert set(new_df["timestamp"].values) == set(df["timestamp"].values)
-        assert set(new_df["totalTest"].values)  == set([150, 200, 950])
-        assert set(new_df["positiveTest"].values) == set([18, 15, 75])
+        assert set(new_df["totalTest_total"].values)  == set([26, 10, 15, 20, 950])
+        assert set(new_df["positiveTest_total"].values) == set([14, 6, 8, 5, 75])
+
+        assert set(new_df["totalTest_age_0_4"].values)  == set([5, 2, 2, 85, 4])
+        assert set(new_df["positiveTest_age_0_4"].values) == set([2, 1, 1, 12, 1])
```
krivard, line 38: add check for `0_17` here, since that one is done differently

krivard:
@dshemetov can i leave review of the tests with you?

quidel_covidtest/tests/test_geo_maps.py
```diff
@@ -1,29 +1,62 @@
...
+        
+        assert set(new_df["totalTest_age_0_17"].values)  == set([10, 4, 4, 170, 8])
+        assert set(new_df["positiveTest_age_0_17"].values) == set([4, 2, 2, 24, 2])
+        
+        # Test Megacounties
+        new_df["timestamp"] = [datetime.strptime(x, "%Y-%m-%d") for x in new_df["timestamp"]]
+        mega_df = add_megacounties(new_df, True)
```
dshemetov, line 49: I noticed that mega_df here looks something like 
```
(Pdb) mega_df[mega_df.columns[:5]]
   timestamp   fips state_id  totalTest_total  positiveTest_total
0 2020-06-15  25013       ma             10.0                 6.0
1 2020-06-15  25015       ma             15.0                 8.0
2 2020-06-15  25027       ma             26.0                14.0
3 2020-06-15  48439       tx            950.0                75.0
4 2020-06-15  53011       wa             20.0                 5.0
0 2020-06-15  25000      NaN             25.0                14.0
1 2020-06-15  53000      NaN             20.0                 5.0
```
Do we want the NAs in state_id there? Probably, since you don't want to double count if you aggregate over state_id later, right? Also, does the value 25.0 for 25000 look right to you? It looks like it just combined the first two FIPS codes together. Just want to make sure, because I'm not entirely sure about the functionality of megacounties here (looks like a form of incomplete aggregation? a bit confused).

dshemetov (approved):
Since this is building on an already deployed version of the code (e.g. `smoothed_tests_per_device`), I only reviewed the new changes. The new changes contained mostly data processing from S3 and some new geo aggregation with megacounties. The new geo aggregation mostly relies on the megacounties in the geomapper. Looking at the code in geomapper, I am a little confused as to the goal of applying a partial rolling sum over the geos without tracking what geos went into that sum. So that's main question is: do the megacounties as defined in geomapper give the functionality we want here? 

Otherwise, looks fine to me. Tests pass, lint passes, and code coverage looks fine too.

quidel_covidtest/delphi_quidel_covidtest/constants.py
```diff
@@ -39,3 +40,12 @@
 #    SMOOTHED_TEST_PER_DEVICE: (True, True),
```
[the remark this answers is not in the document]
jingjtang: Now we don't publish or use test per device signals.

quidel_covidtest/tests/test_geo_maps.py
jingjtang: Thanks Dmitry. I should double check the NAs here. 

As for 25.0, it is correct. We only want to merge the counties with total counts smaller than a certain threshold(25 for smoothed signals, 50 for raw signals) for megacounties

jingjtang: The NA problem is fixed. state_id will be used for geo aggregation in `generate_sensor_for_(non)parent_geo`, so it cannot be NAs, thank for noticing that.

dshemetov: No problem!

Just to double check: so where do we filter out the counties with a low threshold after we do add_megacounties?

jingjtang: In `generate_sensor_for_(non)parent_geo`, we have signal functions from `data_tools` where we have the availability threshold to be 50. All the counties(including the megacounties) do not have enough total counts will be filtered out.

For smoothed signals, we use `_geographical_pooling` where we make the borrow proportion to be 0 if the we borrow too much(more than what we currently have, which means, for smoothed signals, if we have less than 25 counts, we need to borrow more than 25 counts. At that time, `_geographical_pooling` will force the borrow proportion to be 0) fake counts from states. That results in less than 50 counts for all the locations with less than 25 counts. The counties that will be merged into megacounties will definitely be filtered out even aftering the geographical pooling step. That's why I set 25 to be the threshold for smoothed signals when adding megacounties.

dshemetov: Actually, do you think you could write a unit test that produces megacounties and:
1) makes sure that below thresholded counties are filtered out and aggregated
2) makes sure that counties above threshold are left alone and not aggregated?

jingjtang: For the second one, the current unit test `test_county` in `test_geo_maps` do check whether we have some of counties(counts < 25) merged but the others left (>=25). That's how df is designed.

jingjtang: For the first one, we already have unit tests (`test_raw_positive_prop`, `test__geographical_pooling`) for functions in `test_data_tools` to test their functionality including whether the borrow proportion is expected and whether the positive percentage with inadequate total counts is turned into nan.

My logical is to add megacounties and later on all the counties (including megacounties are treated the same way). So the current tests are enough.

quidel_covidtest/delphi_quidel_covidtest/run.py
```diff
@@ -80,65 +95,68 @@ def run_module(params: Dict[str, Any]):
...
+                    last_date=last_date, suffix=agegroup)
+                dates = create_export_csv(
+                    state_df,
+                    geo_res=geo_res,
+                    sensor=sensor_name,
+                    export_dir=export_dir,
+                    start_date=export_start_date,
+                    end_date=export_end_date)
+                if len(dates) > 0:
+                    stats.append((max(dates), len(dates)))
```
dshemetov, line 136: There's some hidden logic here that `geo_groups == geo_map("state", data).groupby("state_id")` . This should make it clearer

```suggestion
assert geo_res == "state"
```

quidel_covidtest/delphi_quidel_covidtest/generate_sensor.py
```diff
@@ -79,11 +82,15 @@ def generate_sensor_for_parent_geo(state_groups, data, res_key, smooth,
...
     Returns:
         df: pd.DataFrame
     """
     has_parent = True
     res_df = pd.DataFrame(columns=["geo_id", "val", "se", "sample_size"])
+    if res_key == "fips": # Add rest-of-state report for county level
+        data = add_megacounties(data, smooth)
     res_groups = data.groupby(res_key)
     loc_list = list(res_groups.groups.keys())
     for loc in loc_list:
```
dshemetov, line 96: The code chunk
```
    res_groups = data.groupby(res_key)
    loc_list = list(res_groups.groups.keys())
    for loc in loc_list:
        res_group = res_groups.get_group(loc)
```
can be written as
```
for loc, res_group in data.groupby(res_key):
```

dshemetov (changes requested):
A few minor nits, while I was re-reading the code.

quidel_covidtest/tests/test_geo_maps.py
dshemetov: Let me think about it a little more, I need to understand this a bit better.

dshemetov: Alright, after looking at it more I am more inclined to believe that this code does what it says, it just was not easy to follow all the threads of pooling and aggregation that live in many different functions. 

I _would_ still feel more comfortable with a test_run test that at least checks that below-threshold counties aren't output. So I wrote a test like this and ran into unexpected behavior. Could you take a look and see? It's in test_run.py and I left a comment about what I did. Maybe I'm misunderstanding something, but it seems like a county that should

quidel_covidtest/tests/test_run.py
```diff
@@ -68,6 +68,18 @@ def test_output_files(self, clean_receiving_dir):
...
 
+        # 7/18/20 in test_data/test_data.csv has:
+        # - zip 24526 -> fips 51019 with 21 positive counts and 3 negative counts
+        # - zip 22079 -> fips 51059 with many positive and negative counts
+        # - zip 24527 -> fips 51143 with 150 positive counts and 3 negative counts
+        # I think that the following output file should then have
+        # - fips 51059 for the many counts
+        # - fips 51000 for the censored 24526 counts
+        # - fips 51143 for the above-threshold 24527 counts
+        # But instead, 51019 shows up with the same value as 51143. Is this right or am I missing something?
```
krivard, line 79: @jingjtang is this a case where it's borrowing from the state?

jingjtang:
@ryantibs Doubled checked the current code and thanks for @dshemetov! Sorry I misunderstood the current geographical pooling and which I think is not that good. 

The current strategy is 

As for smoothed signals:
- For counts smaller than 50, we do geo shrinkage. As for the geo shrinkage, we don't not borrow more than the counts that `parent state` have. Yes, we do have an upper bound of the borrow proportion to be 1, but that is relative to the counts in the home state.  That means, for example, as for county c and its home state s.
   - if we have 2 counts for county c, and we have 30 counts for its home state s. We will borrow only 30 fake counts, since 2 is 48 away for 50, and we need to borrow 48/30(>1) fake states.
   - if we have 2 counts for county c and we have 50(no matter what number that is larger than 48) counts for its home state s, we will borrow 48 fake counts.
   - Notice that, the sample size (counts) we consider for smoothed signals is the rolling sum over 7 day windows.  **That means we decide whether to do the geographical pooling by checking the sum of the counts for the 7 consecutive days.** 
   - _At the current stage, it rarely happens when a state has less than 50 counts for 7 consecutive days. _**Thus, approximately, we can say that for smoothed signals, if a county c have less than 50 counts (but larger than 0, over the last 7 days), we will always borrow enough fake counts to make up the rest.**_
- For counts larger than 50, we don't borrow anything.

As for raw signals:
- For counts smaller than 50, we report nothing.
- For counts larger than 50, we report what we have.

We need to make a decision on the availability threshold. 
1) Keep it the same as how the current one is
2) Switch to what I described (borrow no more than what county c currently have. I vaguely remember this is my newest understanding to this strategy, but turns out it mismatches the current pipeline, sorry again for this.)
3) Come out another strategy for this (If we choose this option, we can consider to give up the geographical pooling strategy from now on.)

Changes in the code needed:
2) little changes, in `_geographical_pooling` function in `data_tools`
3) big changes. But that will make the pipeline for Quidel much easier to understand and update in the future.

ryantibs:
@jingjtang @dshemetov First, a big thanks to Dmitry for catching this. 

Jingjing: my current preference is 2, because I want to move fast and get these signals published. However, can you do a some checks and tell me (in both cases, I'm asking about the all-ages signal, but also about the age-group signals):

1. how much this would change signal availability, and
2. how much this would change signal values themselves.

jingjtang:
@ryantibs The difference between option 2 and option 1 is, we will lose those reported values if the sample sizes are less than 25. 

> 1. how much this would change signal availability, and

image
The counties in the third columns will be unavailable if switching from 1) to 2). But the counties corresponding to the left two columns will remain. 

And here is another histogram where we look into those counties with some of dates missing due to inadequate sample sizes(sum of counts for the last 7 days is in [25, 50]). Time range for the data is from 2020-05-26(the first date when Quidel Covid Test is available) to 2020-01-09, 603 days in total. It shows how many counties have only [25, 50] counts for X days for the smoothed signal.

image

> 2. how much this would change signal values themselves.

The signal values wouldn't change. We will get all the same values if the counts are in the [25, 50] range. We will lose those values if the counts are less than [25, 50]. 

Here are some examples for selected counties with some of dates unavailable for option 2 but more dates available for option 1.
(straight lines are the dates where there are no available report in between)
imageimage
image

jingjtang:
@ryantibs The pipeline is done(`data_tools.py`) with option2 and corresponding sanity checks(unit tests in `test_run.py`, `test_data_tools.py`). If you think option2 is more suitable, we can @dshemetov for another code review. Otherwise, we can switch back to option1.

ryantibs:
Thanks @jingjtang I think I am leaning towards option 2. 

Few questions first

1. The plot titles say "< 25 counts for every consecutive 7 days". I just want to check to understand what you're plotting. You're plotting the sum of counts in 7-day rolling windows, and then reporting how many times that's < 25, right?
2. In the above plots with Counties 21079 and 8107, the straight blue line is just a plotting artifact right? I assume the real signal value is missing there, and the plotting behavior is just to connect it by a straight line?

jingjtang:
@ryantibs 
> 1. The plot titles say "< 25 counts for every consecutive 7 days". I just want to check to understand what you're plotting. You're plotting the sum of counts in 7-day rolling windows, and then reporting how many times that's < 25, right?
Yes, you are correct. But to be more clear, the smoothed signals use the sum of counts in 7-day rolling windows to be sample size in our API, which is different from the raw signals. 

And for the time series plot, the left most one, I checked how many counties have < 25 counts for each date (and if we switch to option2, those counties will be unavailable at all). 

> 2. In the above plots with Counties 21079 and 8107, the straight blue line is just a plotting artifact right? I assume the real signal value is missing there, and the plotting behavior is just to connect it by a straight line?

Yes, it is just a plotting artifact. In fact, we only have values report at the two endpoints. All the dates in between are unavailable for that county.

ryantibs:
Thank you.  I think option 2 is the winner in my mind, but will have a decision for by 5pm (just to be sure, will check to if Roni agrees with me at team-leads meeting).

ryantibs:
Hi @jingjtang, confirming that we should go with option 2.  Thank you for working hard to compare the options.

jingjtang:
Thanks @ryantibs. @dshemetov The pipeline is ready for review with the new strategy. Described in details below for your reference:

- As for smoothed signals:

  - If counts smaller than 50, we do geo shrinkage.  We will borrow no more than the counts that we currently have. That is, for example, if county c has 2 counts, we can only borrow 2 counts at most as pseudo counts from its home state. 

  - If counts larger than 50, we don't borrow anything.

  - The counts for smoothed signals are the sum of counts in the 7-day moving windows. 
  - After the temporal pooling and geographical pooling, we still have an availability checking. If the number of counts is larger than 50, we report it; otherwise, we report nothing.


- As for raw signals:
  - For counts smaller than 50, we report nothing.
  - For counts larger than 50, we report what we have.
  - The counts for raw signals are just the counts for the corresponding date (different from the smoothed signals)

quidel_covidtest/delphi_quidel_covidtest/geo_maps.py
```diff
@@ -12,21 +18,52 @@
...
     return data, geo_key
 
+def add_megacounties(data, smooth=False):
+    """Add megacounties to county level report."""
+    assert "fips" in data.columns # Make sure the data is at county level
+
+    # For raw signals, the threshold is MIN_OBS
+    # For smoothed signals, the threshold is MIN_OBS/2
+    if smooth:
+        threshold_visits = MIN_OBS/2
```
dshemetov, line 48: What's the reason for this change in thresholding here?

quidel_covidtest/tests/test_run.py
```diff
@@ -68,6 +68,15 @@ def test_output_files(self, clean_receiving_dir):
         )
         assert (df.columns.values == ["geo_id", "val", "se", "sample_size"]).all()
 
+        df = pd.read_csv(join("./receiving", "20200718_county_covid_ag_raw_pct_positive.csv"))
+        # TODO: should megacounty 51000 be here (it currently is)?
+        assert set(df.geo_id) == set([51143, 51059, 51000])
```
dshemetov, line 73: @jingjtang I wanted to add some tests for the functionality you mentioned in your last comment.

> 
> As for raw signals:
> 
> For counts smaller than 50, we report nothing.
> For counts larger than 50, we report what we have.
> The counts for raw signals are just the counts for the corresponding date (different from the smoothed signals)

`df` here is

```
   geo_id        val        se  sample_size
0   51000  51.923077  5.693813         77.0
1   51059  40.163934  6.328838         60.0
2   51143  50.000000  6.250000         64.0
```

- ✔️ZIP 24534, FIPS 51083 has 4 counts, so it's below threshold and absent.
- ✔️ZIP 24529, FIPS 51117 has 24 counts, so it's below threshold and absent.
- ✔️ZIP 24526, FIPS 51019 has 49 counts, so it's below threshold and absent.
- ✔️MEGAFIPS 51000 has 77 = 49 + 24 + 4 counts, so it aggregates the two above.
- ✔️ZIP 24527, FIPS 51143 has 64 counts, so it's above threshold and present.
- ✔️ZIP 22079, FIPS 51059 has 60 counts, so it's above threshold and present.

I guess, I have no question here. This is mostly a comment to make sure I understood this, so feel free to ignore (we can remove that #TODO).

quidel_covidtest/tests/test_run.py
```diff
@@ -68,6 +68,15 @@ def test_output_files(self, clean_receiving_dir):
...
+
+        df = pd.read_csv(join("./receiving", "20200718_county_covid_ag_smoothed_pct_positive.csv"))
+        # TODO: should megacounty 51000 be here (it currently isn't)?
+        assert set(df.geo_id) == set([51019, 51143, 51059])
```
dshemetov, line 77: @jingjtang I wanted to do the same test for smoothing. `df` here looks like 

```
   geo_id        val        se  sample_size
0   51000  49.034241  7.069749         50.0
1   51019  52.897278  7.059187         50.0
2   51059  40.163934  6.328838         60.0
3   51143  50.000000  6.250000         64.0
```

- ✔️ZIP 24534, FIPS 51083 has 4 counts, so it's below threshold and absent.
- ✔️ZIP 24529, FIPS 51117 has 24 counts, so it's below threshold and absent.
- ❓ MEGAFIPS 51000 is present because of aggregating the two above 4 + 24 > 25 = MIN_OBS/2; however, sample_size of 50 seems incorrect - because the sample size was only 29. Probably a bug because it assumes sample_size = MIN_OBS?
- ✔️ZIP 24526, FIPS 51019 has 49 counts, so it's geopooled up to 50.
- ✔️ZIP 24527, FIPS 51143 has 64 counts, so it's above threshold and present.
- ✔️ZIP 22079, FIPS 51059 has 60 counts, so it's above threshold and present.

dshemetov (changes requested):
I think I found a small bug, see the comments. Otherwise looks good!

quidel_covidtest/tests/test_run.py
```diff
@@ -68,6 +69,34 @@ def test_output_files(self, clean_receiving_dir):
...
+        parent_pos = 2+12+26+32+24
+        #ZIP 24526, FIPS 51019 has 49 counts, 26 positives, borrow 1 pseudo counts from VA
+        #MEGAFIPS 51000, should have 4+24 = 28 counts, 2+12 = 14 positives, borrow 22 pseudo counts     
+        #ZIP 24527, FIPS 51143 has 64 counts, 32 positives, do not borrow
+        #ZIP 22079, FIPS 51059 has 60 counts, 24 positives, do not borrow
+        assert np.allclose(df.val.values,
+                           [(14+parent_pos*22/parent_test+0.5)/(50+1)*100, 
+                            (26+parent_pos*1/parent_test+0.5)/(50+1)*100,
+                            (24+0.5)/(60+1)*100,
+                            (32+0.5)/(64+1)*100], equal_nan=True)
```
dshemetov, line 98: This is excellent, thank you!

dshemetov (approved):
LGTM!

krivard:
Three questions:
- [ ] Will this change require us to reissue any past data?
- [ ] Does this change require any updates to the API documentation?
- [ ] Are there any special instructions for the first run of the pipeline once this change is released?

jingjtang:
@krivard 
> * [ ]  Will this change require us to reissue any past data?
Yes, but only needs deletion. We have have some counties for smoothed signals deleted. They will not be reported anymore. But for the counties that still pass the availability threshold, the values are unchanged.

> * [ ]  Does this change require any updates to the API documentation?
We can make an update to it to explain the current strategy more clearly. But the current one matches to the current code.

> * [ ]  Are there any special instructions for the first run of the pipeline once this change is released?
Yes. If we run all the data back to the very first date for Quidel(2020-05-26), the cache should be deleted before the first run.  


By the way, since it's a time to re-run the past data, I want to ask whether we are able to extend the export range for Quidel from 45 days to 175 days? (If storage allows) This means for the daily run, we generate reports for the dates from -180 days to -5 days instead of -50 days to -5 days. The backfill related pipeline will benefit from this.

ryantibs:
Sorry if this is contributing noise, but I just want to double check --- apart from the new stricter way of pooling, and the new signals --- 1. that we'll get megacounties, and 2. that the national signal (and all signals) will take values for references dates all the way back to the start of the time period. These are things I've asked for on other issues or messages.

jingjtang:
> Sorry if this is contributing noise, but I just want to double check --- apart from the new stricter way of pooling, and the new signals --- 1. that we'll get megacounties, and 2. that the national signal (and all signals) will take values for references dates all the way back to the start of the time period. These are things I've asked for on other issues or messages.

Yes, we will have
1) megacounties: same smoothing method for smoothed signals. And for raw signal, if the counts are less than 50, they are merged into a megacounty; for smoothed signal, if the counts are less than 25, they are merged into a megacounty.
2) age-specific signals: 0-4, 5-17, 0-17, 18-49, 50-64, 65+
3) If the engineering side allows, we will at least have one time run to generate all the reports back to the start (2020-05-26), and maybe not all the issues(@krivard @korlaxxalrok)
4) new availability threshold for smoothed signals:  
    considering the sample size after temporal pooling
  - <25: report nothing (different from the previous strategy)
  - [25, 50): geographical pooling
  - [50, +infinity): report what they are

for this PR.

krivard:
> Will this change require us to reissue any past data? -> Yes, but only needs deletion.

Cool. To do this we will need the full list of (issue, reference date, geo type, geo value) to delete.

> Does this change require any updates to the API documentation? -> No

Acknowledged

> Are there any special instructions for the first run of the pipeline once this change is released? -> Yes: delete the cache

Acknowledged

> extend the export range for Quidel from 45 days to 175 days?

That's just shy of 4x, on top of the 7x we're adding with this PR -- I don't recommend it. Here's some rough figures imagining the impact on the database if we did that; Quidel currently adds ~110k rows to the database per day, out of a total ~2.8M rows added daily across all indicators.

| scenario | quidel rows added per day | total rows added per day | quidel pct of total |
| -- | -- | -- | -- |
| current | 110k | 2.8M | 4% |
| add age groups | 770k | 3.6M | 22% |
| extend export range | 440k | 3.1M | 14% |
| age groups + export range | 3.1M | 5.8M | 54% |

I do notice that quidel isn't currently using the archive-differ, so some of that 110k rows might be duplicates. Do you happen to know the actual rate of new/updated figures vs duplicates in each submission?

krivard:
Releasing this change has some subtle details -- here's a draft of our options. Pick:

* (A or B) plus (C or D)
* E

# A: If we think the initial run will be fast (<40 minutes)

1. Merge this PR
2. Release covidcast-indicators
3. On prod, clear the `pulled_until` cache. Swap out the production params file with one that will export from 2020-02 to present. Run the indicator. Swap the params file back to the production version.

# B: If we think the initial run will be slow

1. Merge this PR
2. On staging, clear the `pulled_until` cache. Set the params file to export from 2020-02 to present. Run the indicator. Drop the resulting files off in production receiving.
3. Release covidcast-indicators
4. On prod, clear the `pulled_until` cache.

# C: If we think the deletions will be fast (<4 hours)

Delete any time after A2 or B3

# D: If we think the deletions will be slow

Do the deletions need to be in place relatively simultaneously with release? Is it bad for the database to spend multiple days with deletions only partially applied?

## Di: if we need to be rigorous

Schedule a 3-day patch period the same way we did with the JHU patch last fall. Start A1 or B1 on day one of the patch. Complete A/B before concluding the patch period.

## Dii: if we can be more flexible

Delete in batches daily until done. Could take place before, during, or after A/B, depending on if we want deletions finished before the new code runs, or if we want the new code to run before we finalize the set of deletions needed.

# E: If we want archive-differ to handle deletions

If we don't care about deleting from past issues, archive-differ will automatically mark as deleted anything that switches from showing up in an output file to not-showing-up in an output file. "Mark as deleted" means the data is still in the database, and will appear in as-of queries for as-ofs before the deletion date, but will not appear in most-recent queries or in as-of queries for as-ofs after the deletion date.

1. Set up archive-differ bucket in S3
2. Add archiver section to quidel params
3. Initialize the archive bucket with the current snapshot: On staging, set the params file to export from 2020-02 to present. Run the indicator using indicator-runner. Drop the resulting files off in production receiving.
4. Merge this PR
5. Release covidcast-indicators
6. On prod, clear the `pulled_until` cache. Swap out the production params file with one that will export from 2020-02 to present. Run the indicator. Swap the params file back to the production version.

jingjtang:
@krivard 
> Do you happen to know the actual rate of new/updated figures vs duplicates in each submission?

I do not have that kind of data in hand.  Do you want me to check that? 


**Choice between A and B**: the runtime will be **longer than 40 mins** but ~60 mins or so, no more than 90 mins according to my experience. 

**Choice for deletion**: I think Dii is better. We can have the new signals produced as soon as possible and have those corrections for the historical release. 

We do need correction (deleting) for the past issues. So I won't choose E.

krivard:
> > Do you happen to know the actual rate of new/updated figures vs duplicates in each submission?
> 
> I do not have that kind of data in hand. Do you want me to check that?

Yes please; just for a week or two worth of data should be fine.

jingjtang:
@krivard Here is a table showing the percentage of 0 diff 
- compare the val reported today with the val reported yesterday. 
- zero_diff_pct = 90, means on issue_date D for a specific signal, we have 90% of the reported val unchanged across all counties considered and across all dates reported (-50 days to -5 days)
- If yesterday's report for location i and date d is NA, but today's is not NA, then there is a non-zero diff.
- issue_date ranging from 2021-12-21 to 2021-12-28

We can see that for the smallest geographical level (county), we have zero_diff_pct >= 80 for most of the signals. It's good to apply archive-differ to quidel_covidtest too. 
[quidel_zero_diff_pct.csv](https://github.com/cmu-delphi/covidcast-indicators/files/7988498/quidel_zero_diff_pct.csv)

krivard:
Excellent news!

So if the update to the smoothing means we expect to drop 15-50% of new rows we currently report each day for quidel,
and adding archive-differ means we expect to drop 80-90% of new rows we currently report each day for quidel, 
that gives us a 5x-20x total reduction depending on how those two sets overlap.

Let's install archive-differ, release this, see where we are, and then revisit expanding the export range once we have a better idea of where we wind up. That'll look like E without the initialization step, or:

1. ~~Set up archive-differ bucket in S3~~ Done: `quidel`
2. Add archiver section to quidel params

```
  "archive": {
    "aws_credentials": {
      "aws_access_key_id": "{{ delphi_aws_access_key_id }}",
    },
    "cache_dir": "./archivediffer_cache",
    "indicator_prefix": "quidel"
  }
```

3. Merge this PR
4. Release covidcast-indicators
5. On prod, add the `archivediffer_cache` directory. Clear the pulled_until cache. Swap out the production params file with one that will export from 2020-02 to present. Run the indicator. Swap the params file back to the production version.

quidel_covidtest/params.json.template
```diff
@@ -20,6 +20,15 @@
...
   },
+  "archive": {
+    "aws_credentials": {
+      "aws_access_key_id": "{{ delphi_aws_access_key_id }}",
+      "aws_secret_access_key": "{{ delphi_aws_secret_access_key }}"
+    },
+    "bucket_name": "delphi-covidcast-indicator-output",
+    "cache_dir": "./archivediffer_cache",
+    "indicator_prefix": "quidel"
+  }
```
krivard:
```suggestion
  },
```

jingjtang:
@krivard File created to provide deletion info: https://cmu.box.com/s/y41vjkj8z8cjvaqgtz8g5phagms1i3t0
Support code here: https://github.com/cmu-delphi/covidcast-indicators/tree/quidel_deletion

I did the sanity check by 
- randomly check the `value` and `stderr` to see whether they match what we have in our API
- the sample size should be all 50
- the number of the reports for each date at county level seems reasonable.

But it's still better to have another one to help do another sanity check for this.

PR #1470, opened 2022-01-19
Mock covidcast/meta requests.get call

### LINKED ISSUE #1456: Remove network call from validator test of `get_geo_signal_combos`

qx-teo:
### Description
Added mock for requests.get function.
Made sure that signals that are inactive have the 'active' tag as False, which in turn affects whether it is a valid geo_signal combo.

### Changelog
- datafetcher.py

### Fixes 
- Fixes #1456

### DIFF
--- a/_delphi_utils_python/tests/validator/test_datafetcher.py
+++ b/_delphi_utils_python/tests/validator/test_datafetcher.py
@@ -21,25 +21,41 @@ class TestDataFetcher:
         assert not date_filter(FILENAME_REGEX.match("20200620_a_b.csv"))
         assert not date_filter(FILENAME_REGEX.match("202006_a_b.csv"))
 
-    # pylint: disable=fixme
-    # TODO: mock out the advanced meta endpoint /covidcast/meta as well
-    # https://github.com/cmu-delphi/covidcast-indicators/issues/1456
+    # Solution from https://stackoverflow.com/questions/15753390/
+    #how-can-i-mock-requests-and-the-response
+    def mocked_requests_get(*args, **kwargs):
+        class MockResponse:
+            def __init__(self, json_data, status_code):
+                self.json_data = json_data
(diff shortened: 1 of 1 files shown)


qx-teo:
Do I need to cite code I take from stackexchange? Is there some kind of procedure for this?

_delphi_utils_python/tests/validator/test_datafetcher.py
```diff
@@ -21,25 +21,41 @@ def test_make_date_filter(self):
         assert not date_filter(FILENAME_REGEX.match("20200620_a_b.csv"))
         assert not date_filter(FILENAME_REGEX.match("202006_a_b.csv"))
 
-    # pylint: disable=fixme
-    # TODO: mock out the advanced meta endpoint /covidcast/meta as well
-    # https://github.com/cmu-delphi/covidcast-indicators/issues/1456
+    # Solution from https://stackoverflow.com/questions/15753390/
+    #how-can-i-mock-requests-and-the-response
```
nmdefries, line 25: Yeah, linking to where you found a solution is fine, especially if the approach is tricky or complex or non-standard in some way. You can also link to explanations of bugs for larger or more complex bugs.

nmdefries (approved):
Looks good

PR #1471, opened 2022-01-19
Remove hard-coding from get_geo_signal_combos

### LINKED ISSUE #1457: [Validator] Extract source subdivisions from /covidcast/meta instead of hard-coding

qx-teo:
### Description
Uses covidcast/meta to find source-signal pairings, rather than hardcoding values. covidcast/meta stores sources under different names (e.g. quidel_covid_ag instead of quidel). We use a request call to get around this. 
Issue: how to resolve the db_source quidel mapping to multiple sources quidel_covid_ag and quidel_flu?

### Changelog
- datafetcher.py

### Fixes 
- Fixes #1457

### DIFF
--- a/_delphi_utils_python/delphi_utils/validator/datafetcher.py
+++ b/_delphi_utils_python/delphi_utils/validator/datafetcher.py
@@ -112,7 +112,7 @@ def get_geo_signal_combos(data_source):
     """
     # Maps data_source name with what's in the API, lists used in case of multiple names
 
-    source_signal_mappings = {i['db_source']:i['source'] for i in 
+    source_signal_mappings = {i['db_source']:i['source'] for i in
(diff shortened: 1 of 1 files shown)


nmdefries (approved):
Looks good

PR #1472, opened 2022-01-19
Productionize!

korlaxxalrok:
### Description
Get the indicator ready for prod deployment

### Changelog
- Update prod params
- Update Jenkinsfile


nmdefries (approved):
Looks good. We'll also need to add dsew-cpr to sirCAL's params at some point.

korlaxxalrok:
Ok, yep. I think we'll make sure it deploys, then do a test run on stage, and then update SirCAL.

PR #1476, opened 2022-01-20
Flu inpatient

bwilder0:
### Description
Add a new signal for flu inpatient. This combines two new files from CHNG. The numerator is a count of new inpatient admissions with ICD code J09*-J11*. The denominator is the number of new inpatient admissions. Inpatient claims are identified by looking for both the presence of an admission date and the claim type + bill code matching an inpatient setting. The ratio is smoothed/adjusted in the same way as other CHNG signals.
  
### Changelog
- `load_data.py` was changed to load the data for the new signal (the function `load_flu_inpatient_data`)
- Small changes were made throughout `run.py` and `update_sensor.py` to accommodate the fact that this signal's base geography is the state level, not county. Several places had a hardcoded assertion that the base geography was 'fips'. This has been changed to permit 'state_code' when processing the flu inpatient signal.
- Tests were updated both to pass in the base geography explicitly (as is now required) and to test `load_flu_inpatient_data`


bwilder0:
Not sure why the linter is crashing at the very end...I'll take a look tomorrow but any suggestions are appreciated

qx-teo:
Looks like the crash is due to something on a PR I submitted, lemme look into it!

rumackaaron:
@bwilder0 If you merge in the changes mentioned by @qx-teo do all of the checks pass?

bwilder0:
Hm looks like the same thing is still happening after merging the changes. @qx-teo, any thoughts on what might be happening here?

qx-teo:
There was a failing test in _delphi_utils_python that is fixed by the merge, I'm not too sure what's causing the linter issue in changehc but that's probably not related to the issue I was talking about in my earlier comment.

Seems like the usual pylint complaints for the most part, although I've never seen the R1714 error.

krivard:
@bwilder0 the commands you want are
```make lint```
and
```make test```

more information is in the default README.md file for each indicator.

krivard (approved):
👍

krivard:
before merging, we should commit to [writing API documentation](https://github.com/cmu-delphi/delphi-epidata/blob/dev/docs/api/covidcast-signals/chng.md?plain=1) for these new signals

PR #1477, opened 2022-01-20
Fix requests.get mock

qx-teo:
### Description
#1470 mocked out requests.get call, but #1471 uses another requests.get call that did not get mocked properly. We check which call this is and provide the appropriate mocked response.

### Changelog
- test-datafetcher

### Fixes 
Should fix datafetcher test failures like in #1476.


qx-teo:
This should be what's causing the test failures

nmdefries (approved):

PR #1479, opened 2022-01-20
Fix CPR data duplication and missing nation SE and n columns

nmdefries:
### Description
Fix
- State to nation aggregation drops sample size and standard error columns. Add them back explicitly.
- A small number of daily reports were published multiple times (with different asset ids), which causes overcounting for some signals. Deduplicate rows.

### Changelog
- `pull.py`


krivard (approved):

krivard:
are there more things to add to this pr or is it ready to go?

nmdefries:
I'm going to add an assert about each timestamp-geo id having a single row, so we'll get an error message if multiple daily reports were published for a given day *and* they don't match.

nmdefries:
Okay, this is ready

krivard (approved):
🪨 ⭐

PR #1485, opened 2022-01-24
Fix failing HHS test due to type difference

nmdefries:
### Description
Switch some test types from float to int. Associated with `pandas` release of 1.4.0.

### Changelog
- `test_pull.py`


chinandrew (approved):
:shipit: :shipit: :shipit:

PR #1489, opened 2022-01-25
Google symptoms to automatically produce full history for new signals

nmdefries:
### Description
Change logic when handling API metadata so that new signals are automatically produced back to `export_start_date`.

When introducing new signals, `delphi_google_symptoms` doesn't produce their full history, just the last ~18 days (based on what data the validator needs). We want new signals to be automatically backfilled, so change handling of the API metadata to be more precise.

Specifically, in the old version, we looked at all `google-symptoms` signals in metadata to determine how far back to generate data. But this doesn't generalize well. E.g. when we deprecate signals, they will stop getting updated in metadata, but the pipeline will keep trying to update back to the last day they were available for.

This changes metadata processing so that we only look at entries in metadata for currently-generated signals (based on signal names in `constants.py`). If any signals that we want to generate *don't* appear in metadata, we assume that they are new signals and produce the full history.

### Changelog
- `run.py`

### DIFF
--- a/ansible/templates/google_symptoms-params-prod.json.j2
+++ b/ansible/templates/google_symptoms-params-prod.json.j2
@@ -5,7 +5,7 @@
   },
   "indicator": {
     "export_start_date": "2020-02-20",
-    "num_export_days": "all",
+    "num_export_days": null,
     "bigquery_credentials": {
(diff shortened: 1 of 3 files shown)


nloliveira (approved):
Looks great!

PR #1490, opened 2022-01-25
Suppress validation errors from deprecated Google Symptoms signals

nmdefries:
### Description
The validator still expects the old GS signals (`ageusia_raw_search`, etc) to be produced. When it doesn't find them, it fails, so tell it to ignore those checks.

### Changelog
- Production params
- Local params


qx-teo (approved):
lgtm!!

qx-teo:
This might (should) be fixed once the metadata updates itself btw

nmdefries:
Yeah, I expect so too, but it's not great for this to fail whenever we add new signals

PR #1494, opened 2022-01-26
Safegraph patterns: spawn instead of fork threads

nmdefries:
### Description
Safegraph patterns unit tests started hanging recently. This appears to be a problem with the multiprocessing. Based on [this article](https://pythonspeed.com/articles/python-multiprocessing/) and the [docs](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods), it seems to be related to how the threads are created. I switched from default (forking) to spawning.

### Changelog
- `run.py`


jingjtang (approved):
LGTM!

PR #1497, opened 2022-01-27
Return google symptoms to using current pandas version, with mitigations

nmdefries:
### Description
Followup to #1496.

Adapt pipeline to work correctly with new version of pandas. Add test to check for nulled-out rows. Fails when using pandas 1.4.0 and passes when using 1.3.5, which matches expected behavior.

Turn on `numeric_only=True` to suppress future deprecation warning.

### Changelog
- `setup.py`
- `pull.py`
- `geo.py`

### Fixes 
- Replace deprecated `DataFrame.append()`.
- Explicitly convert `dbdate` to `datetime64` to prevent data from being nulled out.

Bug probably related to the [datetime-like changes in 1.4.0](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.4.0.html#datetimelike) and an issue they fixed, https://github.com/pandas-dev/pandas/issues/42921


google_symptoms/delphi_google_symptoms/geo.py
```diff
@@ -37,7 +37,7 @@ def generate_transition_matrix(geo_res):
     if geo_res == "hrr":
         map_df["population"] = map_df["population"] *  map_df["weight"]
 
-    aggregated_pop = map_df.groupby(geo_res).sum().reset_index()
+    aggregated_pop = map_df.groupby(geo_res).sum(numeric_only=True).reset_index()
```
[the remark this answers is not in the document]
nmdefries: I changed this because I'm getting a warning about to-be-deprecated behavior. Thought I'd just throw it in here. 

```
FutureWarning: The default value of numeric_only in DataFrameGroupBy.sum is deprecated. 
In a future version, numeric_only will default to False. Either specify numeric_only or select 
only columns which should be valid for the function.
```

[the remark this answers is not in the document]
nmdefries: The current default does actually seem to be `numeric_only=True`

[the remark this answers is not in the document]
dshemetov: Ah thanks for that! Curious why they're moving away from `numeric_only=True`.

google_symptoms/delphi_google_symptoms/geo.py
```diff
@@ -92,5 +92,5 @@ def geo_map(df, geo_res, namescols =  None):
         newdf["geo_id"] = list(map_df.keys())[1:]
         mask = (newdf == 0)
         newdf[mask] = np.nan
-        converted_df = converted_df.append(newdf)
+        converted_df = pd.concat([converted_df, newdf])
```
krivard, line 95: How many unique timestamps are typical? repeated calls to `concat` are substantially less performant than accumulating a bunch of dfs in a list and concatenating them together all at once. if it's doing this, like, 7 times it's probably fine, but if it's hundreds or thousands we should refactor to a list.

google_symptoms/delphi_google_symptoms/pull.py
```diff
@@ -296,7 +297,7 @@ def pull_gs_data(credentials, export_start_date, export_end_date, num_export_day
         df_dc_county = dfs["state"][dfs["state"]["geo_id"] == "dc"].drop(
             "geo_id", axis=1)
         df_dc_county["geo_id"] = DC_FIPS
-        dfs["county"] = dfs["county"].append(df_dc_county)
+        dfs["county"] = pd.concat([dfs["county"], df_dc_county])
```
krivard, line 300: (not in a loop so this concat is 👍 )

google_symptoms/tests/test_data/state_data_20220916-20220924.csv
```diff
@@ -0,0 +1,19 @@
+open_covid_region_code,date,symptom_Cough,symptom_Phlegm,symptom_Sputum,symptom_Upper_respiratory_tract_infection,symptom_Nasal_congestion,symptom_Post_nasal_drip,symptom_Rhinorrhea,symptom_Sinusitis,symptom_Rhinitis,symptom_Common_cold,symptom_Fever,symptom_Hyperthermia,symptom_Chills,symptom_Shivering,symptom_Low_grade_fever,symptom_Shortness_of_breath,symptom_Wheeze,symptom_Croup,symptom_Pneumonia,symptom_Asthma,symptom_Crackles,symptom_Acute_bronchitis,symptom_Bronchitis,symptom_Anosmia,symptom_Dysgeusia,symptom_Ageusia,symptom_Laryngitis,symptom_Sore_throat,symptom_Throat_irritation,symptom_Type_2_diabetes,symptom_Urinary_tract_infection,symptom_Hair_loss,symptom_Candidiasis,symptom_Weight_gain
```
krivard, line 1: was this change intentional?

krivard (approved):
All good so long as we expect to only ever iterate over a small number of timestamps

google_symptoms/tests/test_data/state_data_20220916-20220924.csv
nmdefries: I added this file for the new test. It has the same format as existing test files but is smaller so the test runs faster. For the sake of keeping test data to a minimum, switching this to use an existing test file.

google_symptoms/delphi_google_symptoms/geo.py
nmdefries: We can have a lot of unique timestamps, e.g. on the occasional backfill run, so switching this to use a list then `concat` all at once.

nmdefries:

PR #1501, opened 2022-01-31
Suppress old `google-symptoms` signals in sirCAL settings

nmdefries:
### Description
Keep sirCAL from alerting about deprecated GS signals (ageusia, anosmia, etc).

### Changelog
- ansible params
- local sirCAL params


krivard (approved):
👍

PR #1505, opened 2022-02-01
Add hospital admissions prop signal to CPR

nmdefries:
### Description
Need prop signal to support dashboard. Add prop signal creation function and tests. Introduce proportion flag as third value in dfs key tuple.

### Changelog
- `constants.py`
- `pull.py`
- `run.py`
- `test_pull.py`
- Local and ansible params templates.

### DIFF
--- a/dsew_community_profile/delphi_dsew_community_profile/constants.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/constants.py
@@ -70,8 +70,8 @@ COUNTS_7D_SIGNALS = {key for key, value in SIGNALS.items() if not value["is_rate
 
 def make_signal_name(key, is_prop=False):
     """Convert a signal key to the corresponding signal name for the API.
-    
-    Note, this function gets called twice with the same `key` for signals that support 
+
+    Note, this function gets called twice with the same `key` for signals that support
     population-proportion ("prop") variants.
     """
     if is_prop:
@@ -79,5 +79,5 @@ def make_signal_name(key, is_prop=False):
     return SIGNALS[key]["api_name"]
 
 NEWLINE = "\n"
-NOT_PROP = "not prop"
-IS_PROP = "prop"
+IS_PROP = True
+NOT_PROP = False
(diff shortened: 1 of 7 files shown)


dsew_community_profile/delphi_dsew_community_profile/constants.py
```diff
@@ -50,22 +50,30 @@ class Transform:
...
+        "api_prop_name": "confirmed_admissions_covid_1d_prop_7dav"
     }
 }
 
 COUNTS_7D_SIGNALS = {key for key, value in SIGNALS.items() if not value["is_rate"]}
 
-def make_signal_name(key):
+def make_signal_name(key, is_prop=False):
     """Convert a signal key to the corresponding signal name for the API."""
+    if is_prop:
```
krivard, line 73:
```suggestion
def make_signal_name(key, is_prop=False):
    """Convert a signal key to the corresponding signal name for the API.
    
    Note, this function gets called twice with the same `key` for signals that support 
    population-proportion ("prop") variants.
    """
    if is_prop:
```

dsew_community_profile/delphi_dsew_community_profile/constants.py
```diff
@@ -50,22 +50,30 @@ class Transform:
...
+        return SIGNALS[key]["api_prop_name"]
     return SIGNALS[key]["api_name"]
 
-NEWLINE="\n"
+NEWLINE = "\n"
+NOT_PROP = "not prop"
+IS_PROP = "prop"
```
krivard, line 79: any reason you're not just handling these as straight booleans?

krivard (approved):
Broadly good, excellent tests, small efficiency nits (optional) & additional clarifying documentation (not)

dsew_community_profile/delphi_dsew_community_profile/constants.py
nmdefries: Ah, good point. I originally didn't have the constants defined so having strings was clearer.

krivard (approved):
🎉

PR #1509, opened 2022-02-03
[CPR] Fix validation failure by setting min expected lag to 3

krivard:
### Description
The CPR file named for 20220131 contains hospital admissions for 20220129, and is typically expected to be available from healthdata.gov on 20220201. This yields a 3-day `min_expected_lag`. 

We expect 5-day lag only on Mondays, since the CPR is not published on weekends. For example, on Monday 20220131, the most recent file available from healthdata.gov was named for 20220128, and contained hospital admissions for 20220126.

The CPR indicator only generates one reference date each day, and the validator only checks output files in a time range specified by `min_expected_lag` and `span_length`. With a 5-day `min_expected_lag`, validator will (1) ignore all files generated on Tuesdays through Fridays, and (2) crash because it can't concatenate an empty list of data frames.

This PR addresses (1), though hopefully QX can let me know if it will created unintended validation failures on Mondays.

### Changelog
- dev and prod params.json

### Fixes 
- Fixes [validator exception in #system-monitoring](https://delphi-org.slack.com/archives/C01LZ3A2UMU/p1643904192247609)

### DIFF
--- a/ansible/templates/dsew_community_profile-params-prod.json.j2
+++ b/ansible/templates/dsew_community_profile-params-prod.json.j2
@@ -12,7 +12,7 @@
     "common": {
       "data_source": "dsew-cpr",
       "span_length": 14,
-      "min_expected_lag": {"all": "5"},
+      "min_expected_lag": {"all": "3"},
       "max_expected_lag": {"all": "9"},
       "dry_run": true,
       "suppressed_errors": []
--- a/dsew_community_profile/params.json.template
+++ b/dsew_community_profile/params.json.template
@@ -18,7 +18,7 @@
     "common": {
       "data_source": "dsew_cpr",
       "span_length": 14,
-      "min_expected_lag": {"all": "5"},
+      "min_expected_lag": {"all": "3"},
       "max_expected_lag": {"all": "9"},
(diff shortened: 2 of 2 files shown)


qx-teo:
Hm, I wrote a method to calculate lags based on a particular day of the week, but that assumes that the indicator only updates once a week (a la safegraph). Currently, it seems like setting `min_expected_lag` to 3 would be the best move? 

I'm pretty sure that as long as `max_expected_lag` is greater than 5 (which it is), then there wouldn't be validation failures on Mondays.

It's still not great to have min/max at 3/9, since it ignores some weird cases (e.g. if the lag on Monday is somehow inexplicably 4 days instead of 5, and we should also want max_lag to change based off the day of the week). But fixing this would require changing how we configure [lag_converter](https://github.com/cmu-delphi/covidcast-indicators/blob/4881e68f108979d3acd934bd7bc58ad0e44cb8b4/_delphi_utils_python/delphi_utils/validator/utils.py#L84).

TLDR this shouldn't create unintended failures, but it means that some (very unlikely) anomalies may not be detected?

qx-teo (approved):
Looks good, see comment

PR #1514, opened 2022-02-04
Delete validation failures

qx-teo:
### Description
During runner runs, upon validation failure for indicators with validator gating, switch to either deleting the csv files stored within `export-dir` (which should only contain files from the run) or moving them to another cache folder, `validation-failure-dir`.

### Changelog
- runner.py
- utils.py

More info in [Slack thread](https://delphi-org.slack.com/archives/C01LZ3A2UMU/p1643906335232629?thread_ts=1643620928.802429&cid=C01LZ3A2UMU) on system-monitoring

### DIFF
--- a/_delphi_utils_python/delphi_utils/utils.py
+++ b/_delphi_utils_python/delphi_utils/utils.py
@@ -101,7 +101,8 @@ def transfer_files():
 def delete_move_files():
     """Delete or move output files depending on dir settings provided in params.
 
-    1. Delete files in export-dir if delivery-dir is specified and is different from export_dir (aka only delete files produced by the most recent run)
+    1. Delete files in export-dir if delivery-dir is specified and is different
+       from export_dir (aka only delete files produced by the most recent run)
     2. If validation-failures-dir is specified, move failed files there instead
     """
     params = read_params()
--- a/_delphi_utils_python/delphi_utils/utils.py
+++ b/_delphi_utils_python/delphi_utils/utils.py
@@ -96,4 +96,4 @@ def transfer_files():
     files_to_export = os.listdir(export_dir)
     for file_name in files_to_export:
         if file_name.endswith(".csv") or file_name.endswith(".CSV"):
-            move(os.path.join(export_dir, file_name), delivery_dir)
+            move(os.path.join(export_dir, file_name), os.path.join(delivery_dir, file_name))
--- a/_delphi_utils_python/delphi_utils/utils.py
+++ b/_delphi_utils_python/delphi_utils/utils.py
@@ -116,6 +116,7 @@ def delete_move_files():
         for file_name in files_to_delete:
             if file_name.endswith(".csv") or file_name.endswith(".CSV"):
                 if validation_failure_dir is not None:
-                    move(os.path.join(export_dir, file_name), os.path.join(validation_failure_dir, file_name))
+                    move(os.path.join(export_dir, file_name),
+                        os.path.join(validation_failure_dir, file_name))
                 else:
                     os.remove(os.path.join(export_dir, file_name))
--- a/_delphi_utils_python/delphi_utils/utils.py
+++ b/_delphi_utils_python/delphi_utils/utils.py
@@ -95,7 +95,7 @@ def transfer_files():
     delivery_dir = params["delivery"].get("delivery_dir", None)
     files_to_export = os.listdir(export_dir)
     for file_name in files_to_export:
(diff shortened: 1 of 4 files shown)


_delphi_utils_python/delphi_utils/utils.py
```diff
@@ -96,4 +96,30 @@ def transfer_files():
...
+
+    # Create validation_failure_dir if it doesn't exist
+    if (validation_failure_dir is not None) and (not os.path.exists(validation_failure_dir)):
+        os.mkdir(validation_failure_dir)
+    # Double-checking that export-dir is not delivery-dir
+    assert(export_dir is not None and export_dir != delivery_dir)
+    files_to_delete = os.listdir(export_dir)
+    if delivery_dir is not None:
+        for file_name in files_to_delete:
+            if file_name.endswith(".csv") or file_name.endswith(".CSV"):
```
nmdefries, line 120: You can combine these cases with `.lower()`

_delphi_utils_python/delphi_utils/utils.py
```diff
@@ -96,4 +96,30 @@ def transfer_files():
...
         if file_name.endswith(".csv") or file_name.endswith(".CSV"):
-            move(os.path.join(export_dir, file_name), delivery_dir)
+            move(os.path.join(export_dir, file_name), os.path.join(delivery_dir, file_name))
+
+def delete_move_files():
+    """Delete csv files in export-dir under certain conditions.
+
+    1. Delivery-dir is specified (aka we are only deleting files produced by the run
+    2. If validation-failures-dir is specified, move failures there instead
+    """
```
nmdefries, line 106: Changing the wording a bit for extra precision.

```suggestion
    """Delete or move output files depending on dir settings provided in params.

    1. Delete files in export-dir if delivery-dir is specified and is different from export_dir (aka only delete files produced by the most recent run)
    2. If validation-failures-dir is specified, move failed files there instead
    """
```

nmdefries (commented):
This looks good to me. Since it would be bad for this to go wrong, @krivard please also take a look.

_delphi_utils_python/delphi_utils/utils.py
```diff
@@ -95,5 +95,32 @@ def transfer_files():
...
+def delete_move_files():
+    """Delete or move output files depending on dir settings provided in params.
+
+    1. Delete files in export-dir if delivery-dir is specified and is different
+       from export_dir (aka only delete files produced by the most recent run)
+    2. If validation-failures-dir is specified, move failed files there instead
+    """
+    params = read_params()
+    export_dir = params["common"].get("export_dir", None)
+    delivery_dir = params["delivery"].get("delivery_dir", None)
```
krivard, line 110: Add early exit here for readability and to reduce empty-directory litter. something like
```suggestion
    delivery_dir = params["delivery"].get("delivery_dir", None)
    if delivery_dir is None:
        return
```

krivard (commented):
Checking my understanding -- here's the truth table of what this code does (numbering the cases for easier reference):

| case | dry_run | validation failure | export dir | delivery dir | v failure dir | action |
| - | - | - | - | - | - | - |
| 1 | true | never | * | * | * | no effect |
| 2 | false | no | * | * | * | no effect |
| 3 | false | yes | X | X | * | throw assertion exception |
| 4 | false | yes | X | None | * | no effect |
| 5 | false | yes | X | Y != X | None | delete CSV from X |
| 6 | false | yes | X | Y != X | Z | move CSV from X to Z |

First, do I have this right?

Second, are we sure about case 4? If no delivery dir is set, but the validation failure dir _is_ set, we still want to do nothing? I'm not sure if I have an opinion either way, it just seemed worth confirming as a corner case.

Otherwise looks solid, modulo a wee nit

_delphi_utils_python/delphi_utils/utils.py
qx-teo: Yep, seems correct. 
Should signals be allowed to be not on dry run, but not have a delivery dir?
If yes, then deleting files would be bad.
I don't think 4 should ever exist technically, so we might want to throw an assertion exception there too?

qx-teo:
Oops - this slipped between the cracks for me 

I'm not too sure if this is what you had, I feel like there's probably some way to improve readability so let me know what you all think!

nmdefries (approved):
My earlier comments have been addressed, so seems good

nmdefries:
@krivard Could you take another look at this?

krivard (approved):
👍

PR #1524, opened 2022-02-14
Small fixes to CPR vax signals to handle cases when fields are not available

nmdefries:
### Description
* Prevent error message that "doses administered" is not available before and including Apr 29, 2021. Instead, return empty df (with proper column names).
* Prevent error message that "fully vaccinated" is not available before and including Apr 11, 2021 for county and msa, and Mar 8 for all regions.
* Condense field availability checks.
* Between May 27 and June 21, 2021, the fully vaccinated field for ages 12-17 was reported with the name "People who are fully vaccinated - 12-17". This gets included in the fully vaccinated parsing such that we end up with duplicates for geo value-date combos during that period.

### Changelog
- `pull.py:: _parse_sheet`


Ananya-Joshi (approved):
Thank you + discussion on Slack!

PR #1527, opened 2022-02-17
Generate full CPR history for new signals

nmdefries:
### Description
If new exported signal does not appear in metadata, switch pipeline to generating output for all dates (instead of "new" or a date range).

Small bonus: change listing filtering by `export_start_date` to include the case when start date and publish date are the same, since cumulative vaccination signals have report date = publish date.

### Changelog
- `pull.py`

### DIFF
--- a/dsew_community_profile/delphi_dsew_community_profile/pull.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/pull.py
@@ -5,7 +5,6 @@ import datetime
 import os
 import re
 from urllib.parse import quote_plus as quote_as_url
-import covidcast
 
 import pandas as pd
 import requests
@@ -361,23 +360,6 @@ def nation_from_state(df, sig, geomapper):
 
 def fetch_new_reports(params, logger=None):
     """Retrieve, compute, and collate all data we haven't seen yet."""
-    # Fetch metadata to check how recent each signal is
-    metadata = covidcast.metadata()
-    sensor_names = {
-        SIGNALS[key][name_field]
-        for key in params["indicator"]["export_signals"]
-        for name_field in ["api_name", "api_prop_name"]
-        if name_field in SIGNALS[key]
-    }
-
-    # Filter to only those we currently want to produce, ignore any old or deprecated signals
-    cpr_metadata = metadata[(metadata.data_source == "dsew-cpr") &
-        (metadata.signal.isin(sensor_names))]
-
-    if sensor_names.difference(set(cpr_metadata.signal)):
-        # If any signal not in metadata yet, we need to backfill its full history.
-        params['indicator']['reports'] = 'all'
-
     listing = fetch_listing(params)
 
     # download and parse individual reports
--- a/dsew_community_profile/delphi_dsew_community_profile/pull.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/pull.py
@@ -5,6 +5,7 @@ import datetime
 import os
 import re
 from urllib.parse import quote_plus as quote_as_url
+import covidcast
 
 import pandas as pd
 import requests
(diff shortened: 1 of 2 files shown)


dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -473,6 +474,23 @@ def nation_from_state(df, sig, geomapper):
...
+        if name_field in SIGNALS[key]
+    }
+
+    # Filter to only those we currently want to produce, ignore any old or deprecated signals
+    cpr_metadata = metadata[(metadata.data_source == "dsew-cpr") &
+        (metadata.signal.isin(sensor_names))]
+
+    if sensor_names.difference(set(cpr_metadata.signal)):
+        # If any signal not in metadata yet, we need to backfill its full history.
+        params['indicator']['reports'] = 'all'
```
krivard, line 492: This will result in generating full history for _all_ CPR signals, not just the ones that are not in metadata. Is this intended?

nmdefries: Not sure that it matters either way; since this source doesn't have backfill, the issue dates aren't particularly important.

If we *don't* want to regenerate all the other signals, we could:
* Turn on the archive differ. The pipeline will still generate the extra files, but they won't be added to the API
* Move this metadata check into `run.py`. I was thinking that we'd do the [first `fetch_new_reports` and `for key, df in dfs.items():`](https://github.com/cmu-delphi/covidcast-indicators/blob/5c127795d96961fd84acdb4d2498079a4718c6ef/dsew_community_profile/delphi_dsew_community_profile/run.py#L59-L72) using the original params, then do a second using updated params where `reports = "all"` and `export_signals` is only the signals missing from metadata.

krivard: I'm thinking of running time (here and during acquisition) and database impact more than data versioning.

This change also makes it impossible to configure the indicator to quickly check/smoketest new signals during development, without it generating the full calendar of output. 

In the past, initial backfill has been handled with a 1-off manual run with a special params file. Is there a reason to move that functionality to the body of the indicator?

nmdefries: > Is there a reason to move that functionality to the body of the indicator?

Mainly the convenience of not having to do a manual run when we add new signals (since CPR is going to get several new ones in the near future). (Google Symptoms [does this too](https://github.com/cmu-delphi/covidcast-indicators/blob/5c127795d96961fd84acdb4d2498079a4718c6ef/google_symptoms/delphi_google_symptoms/run.py#L68-L76).)

I think running time is a toss-up; either the time is spent on a manual run or on an automated run. Automated historical backfill doesn't take *more* time, esp with approach 2 above (approach 1, generating all files and relying on the archive differ, would be somewhat slower, since we're writing extraneous CSVs + diffing, but I'm not sure by how much. We generate all signals regardless of which ones we write to CSV).

Likewise, either of the two workarounds above should have the same db usage as a manual run, since we'd be dropping already-seen dates for existing signals.

> This change also makes it impossible to configure the indicator to quickly check/smoketest new signals during development, without it generating the full calendar of output.

I believe this can be addressed as well (thanks for noting) by not allowing metadata to override provided params if `reports` is set to an explicit date range. But a local run requesting "new" `reports` for a new signal would still get overridden, yes.

Anyway, let me know if you think the pros/cons/complications don't balance out here.

krivard: Alright, you've convinced me 😄 

Let's go with option 2 (move this logic to run.py) and also prevent it from overriding explicitly-set date ranges from params.json.

dsew_community_profile/delphi_dsew_community_profile/run.py
```diff
@@ -71,6 +72,48 @@ def replace_date_param(p):
         if len(dates)>0:
             run_stats.append((max(dates), len(dates)))
 
+    ## If any requested signal is not in metadata, generate it for all dates.
+    #
+    # Only do so if params.reports is set to "new". If set to "all", the
+    # previous fetch_new_reports + CSV loop will already have generated the full
+    # history for new signals. If params.reports is set to a specific date
+    # range, that request overrides automated backfill.
```
krivard, line 80: 💯

krivard (approved):
👍  glorious

PR #1530, opened 2022-02-23
Keep sirCAL from complaining about `try_vaccinate_1m` at the MSA level

nmdefries:
### Description
Due to decreasing sample size, the weighted version of `try_vaccinate_1m` is no longer available. The unweighted version only has one MSA region available, so it will be missing soon too. Keep sirCAL from complaining about either.

### Changelog
- Local and ansible `params`


krivard (approved):
👍

PR #1535, opened 2022-02-25
add hrr to quidel ignored regions for omicron 5_17

krivard:
### Description
SirCAL complained about lag or gap for the quidel omicron signal for ages 5-17 in hrrs (it had not done so before). This is roughly as expected as similar lags/gaps in age group signals for small regions, so adding it to the ignore list.

### Changelog
- prod & dev sircal params

### Fixes 
- Fixes SirCAL false positive


jingjtang:
@krivard Does msa suffer from this as well, or just hrr?

krivard:
both:
image

jingjtang (approved):
LGTM

PR #1522, opened 2022-02-09
Fix archiver bug ignoring deletions when comparing two files with no missing columns

dshemetov:
### Description
A partial fix of the issues in #1520. 

### Changelog
Previously, the archiver did not code deletions between two old-style CSVs (old-style meaning no missing columns). This changes it to always write deletions with DELETED missing columns and NAs in the missing columns for the rest of the non-missing rows. The validation and repair for the NA missing columns is then handled by the [acquisition](https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/acquisition/covidcast/csv_importer.py#L224-L253).

To be merged after https://github.com/cmu-delphi/delphi-epidata/pull/862.

### Fixes 
- partial fix #1520

PR #1542, opened 2022-03-02
Refactor archiver tests

krivard:
Based on #1522

### Description
_Type of change (bug fix, new feature, etc), brief description, and motivation for these changes._

Archiver tests are a twisty maze of numbered csv files, each file testing a different corner case. The before, after, and expected .diff content is defined in different places in the code, sometimes multiple times. This approach is difficult to interpret without spending extensive time with the code, making it (1) difficult to update, (2) difficult to review, and (3) easy for errors or inconsistencies to creep in.

This PR refactors the archiver tests to:

* collect information about each file together
* only compute expected data once
* name files with the corner case they test, so it's easier to keep track

### Changelog
- test_archive.py: substantial refactor
    - Example class
    - Expecteds class
    - `CSV_BEFORE`, `CSV_AFTER`, various `expected_csv[0-9]_diff` definitions merged into a single `CSVS` dict
    - csv[0-9] converted to descriptive names eg "deleted" "added" "unchanged"
    - new global `EXPECTEDS` storing the filenames expected in various locations at different stages of archivediffing
    - repetitive manual lists of expected filenames converted in most cases to computational definitions based on the `CSVS` dict, with an assert check wherever changes to the test cases must be manually propagated to `EXPECTEDS`
    - ArchiveDifferTestlike class with centralized definitions of cache/export setup and final filtered exports check
    - File content spot-checks converted to exhaustive checks
    - TestArchiveDiffer, TestS3ArchiveDiffer, TestGitArchiveDiffer converted from `CSV_BEFORE` etc to `CSVS` and `EXPECTEDS`

### Fixes 
- Working with tests should not feel like this:
image


_delphi_utils_python/tests/test_archive.py
```diff
@@ -22,125 +22,199 @@
...
+    new=["add_file.csv"],
+)
+# check for incomplete modifications to tests
+assert set(EXPECTEDS.new) == set(f"{csv_name}.csv" for csv_name, dfs in CSVS.items() if dfs.before is None), \
+    "Bad programmer: added more new files to CSVS.after without updating EXPECTEDS.new"
+EXPECTEDS.raw_exports = list(EXPECTEDS.common_diffs.keys()) + \
+    EXPECTEDS.new
+EXPECTEDS.diffed_exports = EXPECTEDS.raw_exports + \
+    [diff_name for diff_name in EXPECTEDS.common_diffs.values() if diff_name is not None]
+EXPECTEDS.filtered_exports = [f[:-5] for f in EXPECTEDS.diffed_exports if f.endswith(".diff")] + \
```
dshemetov, line 194: Is the f[:-5] stripping off ".diff"? `f.replace(".diff", "")` is probably more readable.

_delphi_utils_python/tests/test_archive.py
```diff
@@ -22,125 +22,199 @@
...
+    def set_up(self, tmp_path):
+        cache_dir = join(str(tmp_path), "cache")
+        export_dir = join(str(tmp_path), "export")
+        mkdir(cache_dir)
+        mkdir(export_dir)
+        return cache_dir, export_dir
+    def check_filtered_exports(self, export_dir):
+        assert set(listdir(export_dir)) == set(EXPECTEDS.filtered_exports)
+        for f in EXPECTEDS.filtered_exports:
+            example = CSVS[f[:-4]]
```
dshemetov, line 210: Similar here `f.replace(".csv", "")`

_delphi_utils_python/tests/test_archive.py
```diff
@@ -187,47 +238,37 @@ def test_diff_and_filter_exports(self, tmp_path):
...
-            index_cols=["geo_id"]
-        )
+        assert set(listdir(export_dir)) == set(EXPECTEDS.diffed_exports)
+
+        # Check that the diff files look as expected
+        for key, diff_name in EXPECTEDS.common_diffs.items():
+            if diff_name is None: continue
+            _assert_frames_equal_ignore_row_order(
+                pd.read_csv(join(export_dir, diff_name), dtype=CSV_DTYPES),
+                CSVS[key[:-4]].diff,
```
dshemetov, line 269: Similar here

_delphi_utils_python/tests/test_archive.py
```diff
@@ -22,125 +22,199 @@
...
+    common_diffs=dict(
+        (f"{csv_name}.csv", None if dfs.diff is EMPTY else f"{csv_name}.csv.diff")
+        for csv_name, dfs in CSVS.items() if dfs.diff is not None
+    ),
+    new=["add_file.csv"],
+)
+# check for incomplete modifications to tests
+assert set(EXPECTEDS.new) == set(f"{csv_name}.csv" for csv_name, dfs in CSVS.items() if dfs.before is None), \
+    "Bad programmer: added more new files to CSVS.after without updating EXPECTEDS.new"
+EXPECTEDS.raw_exports = list(EXPECTEDS.common_diffs.keys()) + \
```
dshemetov, line 190: `raw_exports`, `diffed_exports`, and `filtered_exports` can probably be put into the `__post_init__` of `Expecteds`, since they're all derived from the inputs.

dshemetov: `raw_exports` appears to be publicly unused, maybe just a local variable in `post_init`?

_delphi_utils_python/tests/test_archive.py
```diff
@@ -187,47 +238,37 @@ def test_diff_and_filter_exports(self, tmp_path):
...
+        for csv_name, dfs in CSVS.items():
+            if dfs.before is not None:
+                dfs.before.to_csv(join(cache_dir, f"{csv_name}.csv"), index=False)
+            if dfs.after is not None:
+                dfs.after.to_csv(join(export_dir, f"{csv_name}.csv"), index=False)
         arch_diff._cache_updated = True
 
         deleted_files, common_diffs, new_files = arch_diff.diff_exports()
 
         # Check return values
```
dshemetov, line 250: Thinking this comment is worth updating. 

```suggestion
        # Check deleted, common, diffed, and new file names match expected
```

_delphi_utils_python/tests/test_archive.py
```diff
@@ -187,47 +238,37 @@ def test_diff_and_filter_exports(self, tmp_path):
...
+        assert set(deleted_files) == {join(cache_dir, f) for f in EXPECTEDS.deleted}
         assert set(common_diffs.keys()) == {
-            join(export_dir, f) for f in ["csv0.csv", "csv1.csv", "csv4.csv", "csv5.csv"]}
-        assert set(new_files) == {join(export_dir, "csv3.csv")}
-        assert common_diffs[join(export_dir, "csv0.csv")] is None
-        assert common_diffs[join(export_dir, "csv1.csv")] == join(
-            export_dir, "csv1.csv.diff")
+            join(export_dir, csv_name) for csv_name in EXPECTEDS.common_diffs}
+        assert set(new_files) == {join(export_dir, f) for f in EXPECTEDS.new}
+        assert all(
```
dshemetov, line 255:
```suggestion
        # Check that diffed file names are identical
        assert all(
```

dshemetov (approved):
Thanks for doing this, definitely overdue! Few minor suggestions, but overall looks great 🚀

_delphi_utils_python/tests/test_archive.py
krivard: hmmm i could go either way -- it's largely here as a building block for `EXPECTEDS.diffed_exports`.

pros:
* it makes the 3 `EXPECTEDS.*_exports` mirror the 3 phases of archive-diffing
* it makes the `EXPECTEDS.diffed_exports` definition (raw exports + .diff files), which matches the (my?) mental model of execution better than (common_diffs.keys() + new + .diff files)

cons:
* not actually used in tests
* adds clutter / cognitive load

i don't feel strongly about it though so if cons outweigh pros for you i can make the change

_delphi_utils_python/tests/test_archive.py
krivard:
```suggestion
                CSVS[key.replace(".csv", "")].diff,
```

krivard (commented):
this was excellent, thank you

_delphi_utils_python/tests/test_archive.py
krivard: also TIL `__post_init__` !!

dshemetov: Yea, doesn't seem like a big deal. I'll make the `__post_init__` suggestion above.

_delphi_utils_python/tests/test_archive.py
```diff
@@ -22,125 +22,199 @@
...
+
+@dataclass
+class Expecteds:
+    deleted: Any
+    common_diffs: Any
+    new: Any
+    raw_exports: Any = None
+    diffed_exports: Any = None
+    filtered_exports: Any = None
+
```
dshemetov, line 50: Just this change should be good enough

```suggestion
    def __post_init__(self):
        self.raw_exports = list(self.common_diffs.keys()) + self.new
        self.diffed_exports = self.raw_exports + [diff_name for diff_name in self.common_diffs.values() if diff_name is not None]
        self.filtered_exports = [f.replace(".diff", "") for f in self.diffed_exports if f.endswith(".diff")] + self.new
```

_delphi_utils_python/tests/test_archive.py
```diff
@@ -22,125 +22,199 @@
...
+        return df
+
+@dataclass
+class Expecteds:
+    deleted: Any
+    common_diffs: Any
+    new: Any
+    raw_exports: Any = None
+    diffed_exports: Any = None
+    filtered_exports: Any = None
```
dshemetov, line 49:
```suggestion
    deleted: List[str]
    common_diffs: Dict[str, str]
    new: List[str]
    raw_exports: List[str] = None
    diffed_exports: List[str] = None
    filtered_exports: List[str] = None
```

_delphi_utils_python/tests/test_archive.py
```diff
@@ -22,125 +22,199 @@
...
+)
+# check for incomplete modifications to tests
+assert set(EXPECTEDS.new) == set(f"{csv_name}.csv" for csv_name, dfs in CSVS.items() if dfs.before is None), \
+    "Bad programmer: added more new files to CSVS.after without updating EXPECTEDS.new"
+EXPECTEDS.raw_exports = list(EXPECTEDS.common_diffs.keys()) + \
+    EXPECTEDS.new
+EXPECTEDS.diffed_exports = EXPECTEDS.raw_exports + \
+    [diff_name for diff_name in EXPECTEDS.common_diffs.values() if diff_name is not None]
+EXPECTEDS.filtered_exports = [f.replace(".diff", "") for f in EXPECTEDS.diffed_exports if f.endswith(".diff")] + \
+    EXPECTEDS.new
```
dshemetov, line 195:
```suggestion
```

_delphi_utils_python/tests/test_archive.py
dshemetov: The thing I haven't figured out about dataclasses yet is how to specify a public variable that's not meant to be initialized. Like here, a user can set these, but they'll just get overwritten. Haven't found a way around that yet and overwriting is probably still better than hiding these public variables in the body of the class.

dshemetov: Oh I think this is how you do it.

_delphi_utils_python/tests/test_archive.py
```diff
@@ -1,4 +1,4 @@
-
+from dataclasses import dataclass
```
dshemetov, line 1:
```suggestion
from dataclasses import dataclass, field
```

PR #1548, opened 2022-03-03
Filter out small test-volume signals in CPR

### LINKED ISSUE #1513: Visual checks: CPR num tests and test positivity at the county level
[Originating Asana task](https://app.asana.com/0/0/1201679922381906/f) with line plots and choropleths for hospital admissions, test volume, and test positivity.

**Outstanding question:** Should we suppress test positivity for small counts? Small counts cause the signal to be highly variable.

Number of tests and test positivity at the county level need further visualization, stratified in some way to make plots easier to interpret.

nmdefries:
### Description
Turn test volume signal into the sample size for test positivity and drop obs where test volume <= 5 for a given day/geo level.

Bonus:
- Switch testing names to [new approved ones](https://delphi-org.slack.com/archives/C013AH5N01E/p1643313185085300?thread_ts=1643045140.069100&cid=C013AH5N01E).
- More descriptive names and comments for `SIGNALS`.

### Changelog
- `pull.py`
- `constants.py` - clarify comments and names
- `test_pull.py` - update tests

### Fixes 
Closes #1513.


Ananya-Joshi:
Should we make a specific test to address this PR specifically? Specifically, tests for unify_testing_sigs, add_max_ts_col, and std_err.

nmdefries:
Yep, that makes sense

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -513,7 +521,29 @@ def fetch_new_reports(params, logger=None):
...
+                "publish_date", axis=1
+            )
+
+            # No longer need "total" signal.
+            del ret[total_key]
+        elif sig == "total":
+            # If the signal is test volume, keep publish date to be used when
+            # combining with test positivity.
+            continue
+        else:
```
krivard, line 541: I was momentarily confused here on whether we're assuming a particular traversal order -- check my understanding?
* we're iterating over a copy, so we'll always encounter both naat `sig` keys no matter what
* we don't actually care which key we see first; the `total` case is only here to prevent it from falling into the `.drop` handling in the else case

If that's correct, I have a slight preference for the following, which makes the latter explicit:

```suggestion
        elif sig != "total":
```

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -546,3 +576,109 @@ def generate_prop_signal(df, geo, geo_mapper):
...
 
     return df
+
+def unify_testing_sigs(positivity_df, volume_df):
+    """
+    Drop any observations with a sample size of 5 or less. Generate standard errors.
+
+    This combines test positivity and testing volume into a single signal,
+    where testing volume *from the same spreadsheet/publish date* (NOT the
+    same reported date) is used as the sample size for test positivity.
```
krivard, line 586: Delphi's been using "reported date" as a soft synonym for publish date, so we should probably use "reference date" here
```suggestion
    same reference date) is used as the sample size for test positivity.
```

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -546,3 +576,109 @@ def generate_prop_signal(df, geo, geo_mapper):
...
+    before the test positivity period. Since the CPR is only published on
+    weekdays, test positivity and test volume are only available for the same
+    reported dates 3 times a week. We have chosen to censor 7dav test
+    positivity based on the 7dav test volume provided in the same originating
+    spreadsheet, corresponding to a period ~4 days earlier.
+
+    This approach makes the signals maximally available (5 days per week) with
+    low latency. It avoids complications of having to process multiple
+    spreadsheets each day, and the fact that test positivity and test volume
+    are not available for all the same reported dates.
```
krivard, line 598:
```suggestion
    are not available for all the same reference dates.
```

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -546,3 +576,109 @@ def generate_prop_signal(df, geo, geo_mapper):
...
+
+    Discussion of decision and alternatives:
```
krivard, line 600:
```suggestion
    Discussion of decision and alternatives (Delphi-internal share drive):
```

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -546,3 +576,109 @@ def generate_prop_signal(df, geo, geo_mapper):
...
+    https://docs.google.com/document/d/1MoIimdM_8OwG4SygoeQ9QEVZzIuDl339_a0xoYa6vuA/edit#
+
+    """
+    # Combine test positivity and test volume.
```
krivard, line 604:
```suggestion
    # Combine test positivity and test volume, maintaining "this week" and "previous week" status.
```

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -546,3 +576,109 @@ def generate_prop_signal(df, geo, geo_mapper):
...
+
+    # Drop everything with 5 or fewer total tests.
+    df = df.loc[df.sample_size > 5]
+
+    # Generate stderr.
+    df = df.assign(se=std_err(df))
+
+    return df
+
+def add_max_ts_col(df):
```
krivard, line 630: Add a test for this

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -546,3 +576,109 @@ def generate_prop_signal(df, geo, geo_mapper):
     df.drop(["population", geo], axis=1, inplace=True)
 
     return df
+
+def unify_testing_sigs(positivity_df, volume_df):
```
krivard, line 580: Add a test for this

krivard (changes requested):
Great work! minor readability nit, vocabulary nit, and two missing tests just to verify the logic

dsew_community_profile/delphi_dsew_community_profile/pull.py
nmdefries: Yep, that's the desired logic.

PR #1550, opened 2022-03-07
Fix breaking moto update

dshemetov:
### Description
Fixes a few tests in `_delphi_utils/../test_archive.py`.

### Changelog
Remove the custom mock_s3 fixture, use the built-in provided by moto, and duplicate the s3_client definition. Might have something to do with the breaking changes in v3.0 [here](https://github.com/spulec/moto/blob/master/CHANGELOG.md#300).
> The behaviour of the class-decorator has been reworked - the state is now reset before every test-method.

### Fixes 
- Breaking tests in other branches.

### DIFF
--- a/_delphi_utils_python/tests/test_archive.py
+++ b/_delphi_utils_python/tests/test_archive.py
@@ -267,18 +267,13 @@ AWS_CREDENTIALS = {
 }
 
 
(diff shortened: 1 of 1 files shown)


krivard (approved):
Looks legit!

Issue #1523, opened 2022-02-10
2019 FIPS 02261, removed in 2020 census data, appears in DSEW CPR 


krivard:
**Actual Behavior:**

Validator failure:
```
check_bad_geo_id_value failed for confirmed_admissions_covid_1d_7dav at resolution county on 2022-02-07: 
Unrecognized geo_ids (not in historical data) ['02261']
```

Looks like 02261 is not listed in our 2020 census data, but is present in the 2019 data.

**Expected behavior**

No validation errors. Not sure whether we should switch to 2019 for this indicator, or if there are also FIPS in it which were added in 2020 and aren't present in the 2019 data. Ideally DSEW would use 2020 FIPS for its data collection but 🤷‍♀️ 

**Context**

nmdefries:
cc @neul3 @M5Skid

nmdefries:
`dsew-cpr` has been removed https://github.com/cmu-delphi/covidcast-indicators/pull/1871. Do we want to patch this?

PR #1555, opened 2022-03-10
[DSEW CPR] [DRAFT] Add basic interpolation function and a test stub

### LINKED ISSUE #1539: Add retrospective gapfilling to CPR hospital admissions signals
Because the CPR only includes one (sometimes two) reference dates for each signal, and because the CPR is not published on weekends, the resulting COVIDcast signals are only available 5 days a week. See the green line in this timeseries chart:

image

We should use simple interpolation to fill these gaps **retrospectively** (once the next file becomes available). Extrapolation on days where no new files are posted is left as a future research project.

We'll need to do something like this:

* When computing the list of CPR files to process, extend the series backwards by one file ("additional CPR file")
* Process the CPR files into a df as usual
* Expand the df to cover a contiguous date sequence, NA-filling
* Impute the missing values
* Drop the per-signal reference dates in the additional CPR file
* Aggregate to nation, export, etc as usual

## Complication
The existing imputation utility in `delphi_utils.smooth` assumes we only ever impute a value for date X using data from dates Y<X. We will probably need to add support for some kind of symmetrical mode. Expected signal data for this change looks like `[x1, NA, NA, x4]` -- literally, since these gaps mostly occur at weekends. `x1` is from the additional CPR file, `x4` is from today's CPR file, and we want to publish the two imputed values and `x4`. This probably means a linear or other low-degree fit that doesn't need a lot of context to work.

This revision must be completed before we can begin showing county-level hospital admissions in the visualizations on the website.

dshemetov:
### Description
Adds interpolation to most DSEW CPR signals. Partially fixes #1539. Final work will be in #1576.

### Changelog
- Add an interpolation function that fills in missing rows with reasonable defaults.
- Add tests for the interpolation.

### Fixes 
- part of #1539.


dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -543,8 +546,29 @@ def fetch_new_reports(params, logger=None):
...
 
+    ret = interpolate_missing_values(ret)
+
     return ret
 
+def interpolate_missing_values(dfs: DataDict) -> DataDict:
+    interpolate_df = dict()
+    for key, df in dfs.items():
+        geo_dfs = []
+        for geo, group_df in df.groupby("geo_id"):
```
krivard, line 557: how painful will this be for counties?

dsew_community_profile/tests/test_pull.py
```diff
@@ -421,3 +434,37 @@ def test_std_err(self):
                     "sample_size": [2, 2, 5, 10, 20, 0]
                 })
             )
+
+    def test_interpolation(self):
+        ex1 = Example(
+            geo_id = ["1"] * 8 + ["2"] * 6,
+            timestamp = list(pd.date_range("2022-01-01", "2022-01-05")) + list(pd.date_range("2022-01-08", "2022-01-10")) + list(pd.date_range("2022-01-01", "2022-01-03")) + list(pd.date_range("2022-01-08", "2022-01-10")),
+            val = [i ** 2 - i + 5 for i in range(2, 12) if i % 7 >= 2] + [3 * i ** 2 + 4 * i - 10 for i in range(4, 14) if i % 7 >= 4]
```
krivard, line 442: what does this do?

dsew_community_profile/delphi_dsew_community_profile/pull.py
dshemetov: on first impression, I don't think it will change at all. the only edge case that comes up is if certain counties are _really_ sparse, like to the point of only being reported selectively, when data is available (like fb survey). then things would get quite bad. not sure if that's a possibility in this indicator though?

dsew_community_profile/tests/test_pull.py
dshemetov: so it's a test of a data signal with two different geos and with two different missing lengths (the first one is missing dates 2022-01-06 and 2022-01-07 and the second one misses 2022-01-04 -- 2022-01-7). the values of each signal comes from one of two second-degree polynomials, something that a cubic interpolator can infer with 100% accuracy.

krivard: cool -- let's make that more legible; i got lost among the exponents and moduli. mix-n-match strategies for this could include:
* write a comment explaining what is being done and why ("something that a cubic interpolator can infer with 100% accuracy" is 💯  but non-obvious)
* define the second-degree polynomials first in a dedicated variable or function, then handle the gap logic when building the data frame
* define the data frame for the dense date range, then selectively drop the desired missing days

dsew_community_profile/delphi_dsew_community_profile/pull.py
krivard: i meant more, will it take a long time to loop over 3600 counties x 7 signals?

dshemetov: Ran a profile of the code on the first week of March data. Most of the indicator time is taken by reading the excel files, with the interpolation showing up at the bottom here (.5s total time on the scale of excel reading taking 10+s).

```py
         162936279 function calls (161659417 primitive calls) in 92.702 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     8622    8.433    0.001    8.433    0.001 {method 'feed' of 'xml.etree.ElementTree.XMLParser' objects}
  4513240    5.941    0.000   13.615    0.000 _reader.py:177(parse_cell)
  4513240    4.569    0.000    5.190    0.000 _openpyxl.py:552(_convert_cell)
  4513240    3.941    0.000    4.305    0.000 cell.py:193(coordinate_to_tuple)
21827158/21826918    2.893    0.000    3.721    0.000 {built-in method builtins.isinstance}
  8916288    2.510    0.000    3.605    0.000 ElementTree.py:1292(read_events)
    34280    2.281    0.000    4.530    0.000 _read_only.py:101(_get_row)
    34312    2.037    0.000   31.770    0.001 _reader.py:113(parse)
  4513240    1.692    0.000    2.249    0.000 read_only.py:14(__init__)
  4040858    1.443    0.000    1.443    0.000 _reader.py:78(_cast_number)
8907712/8907673    1.259    0.000   13.695    0.000 ElementTree.py:1224(iterator)
    46905    1.046    0.000    3.345    0.000 managers.py:1061(iset)
  4513240    0.997    0.000    0.997    0.000 {method 'findtext' of 'xml.etree.ElementTree.Element' objects}
    34280    0.970    0.000   14.585    0.000 _reader.py:286(<listcomp>)
   279075    0.934    0.000    0.934    0.000 {method 'reduce' of 'numpy.ufunc' objects}
 13556161    0.931    0.000    0.931    0.000 {method 'get' of 'xml.etree.ElementTree.Element' objects}
526926/205994    0.803    0.000    5.524    0.000 {built-in method numpy.core._multiarray_umath.implement_array_function}
    34280    0.740    0.000    5.930    0.000 _openpyxl.py:576(<listcomp>)
3397777/2731437    0.725    0.000    0.942    0.000 {built-in method builtins.len}
4325556/4296437    0.661    0.000    0.857    0.000 {built-in method builtins.getattr}
   168610    0.622    0.000    1.118    0.000 generic.py:5525(__finalize__)
  8793866    0.621    0.000    0.621    0.000 read_only.py:108(value)
128896/25837    0.607    0.000    3.062    0.000 serialisable.py:46(from_tree)
  4513240    0.557    0.000    0.557    0.000 read_only.py:112(value)
        1    0.553    0.553   41.121   41.121 pull.py:553(interpolate_missing_values)
```

dshemetov:
Pushed more and clearer tests. Also made the interpolation default to linear - requires only two points to do, has simpler behavior, and should produce consistent interpolation when datasets don't fully overlap.

Want to do a few more tests today and convince myself that the archiver will be given consistent input, week after week.

dshemetov:
So I ran the current indicator on two overlapping date ranges to see how the outputs would differ.

dates1: 2022-03-05 -- 2022-03-11
dates2: 2022-03-08 -- 2022-03-14

Comparing the two receiving directories, we get a few cases:
1. Files that are only in receiving_dates1 have the date ranges [20220225 -- 20220307]
2. Files that are only in receiving_dates2 have the date ranges [20220309 -- 20220314]
3. Common files on the left time boundary contain a few rows present in receiving_dates1 and missing in receiving_dates2
    - this is either a) an actual deletion, b) an interpolation that loses a value on the _left_ of the NA, so can't fill anymore
4. Common files in middle of the time boundary, with 90% row changes. This is likely updating data.
5. Common files on the right time boundary contain a few rows present in receiving_dates2 and missing in receiving_dates1
    - either a) an actual new value, b) an interpolation that loses its value on the _right_ of the NA, so can't be filled

Seems like it will be easy to introduce bugs here if we aren't careful. Possible issue: the archivediffer treating missing rows in (3) and (5) as deletions and deleting valid interpolations. Possible solution is to restrict the receiving output of the indicator to a given range, such as that middle common files range (4). Now can we find that time range consistently, given the strange behavior of weekends in this indicator? Unclear. Currently, the indicator is still limited to only processing a ~6-day window for the data or else it runs into an error with having multiple values for the same (geo, time, signal) pair.

dsew_community_profile/delphi_dsew_community_profile/pull.py
krivard: Oh cool. It looks like we typically run in 1m15, so doubling that won't be a big deal day-to-day. Initial runs when adding new signals are 45m, so we'll need to be a bit careful when we add the test positivity signals, but that won't happen very often. 👍

dshemetov:
@krivard I updated your comment with the filled table.

krivard:
Awesome. I had some time to kill while waiting for builds to finish so I fiddled with this a bit, and it looks like (at least for hospital admissions) we're getting exactly what we want:

* neither the main branch nor the interpolation branch makes changes to previously-computed days where `reports` ranges overlap
* the interpolation branch makes no changes to days where we have source data
* the interpolation branch generates output for days where we don't have source data

What I did:
```
$ for branch in main ds/dsew-interpolation; do git checkout $branch; for r in 2022-03-05--2022-03-11 2022-03-08--2022-03-14; do rcv="receiving_${branch/\//_}_$r"; mkdir $rcv ||break;  env/bin/python -m delphi_utils set indicator.reports $r common.export_dir $rcv; env/bin/python -m delphi_dsew_community_profile ||break; done; done;
$ diff -rq receiving_main_2022-03-0* |less
# ^^ looks like the old pipeline doesn't sort its output, but we can get around that
$ diff -rq receiving_main_2022-03-0* |grep differ |awk '{print "diff -u <(sort " $2 ")","<(sort " $4 ")"}' |bash |less
# ^^ no files truly differ after sorting
$ diff -rq receiving_ds_dsew-interpolation_2022-03-0* |less
# ^^ new pipeline does sort its output, so there are no files that differ
$ diff -rq receiving_*_2022-03-05--2022-03-11 |less
# ^^ no new files, as expected; differing files is probably the sorting issue
$ diff -rq receiving_*_2022-03-05--2022-03-11 |grep differ |awk '{print "diff -u <(sort " $2 ")","<(sort " $4 ")"}' |bash |less
# ^^ no files truly differ
$ diff -rq receiving_*_2022-03-08--2022-03-14 |less
# ^^ new files covering the gap in the interpolation branch; differing files is probably the sorting issue
$ diff -rq receiving_*_2022-03-08--2022-03-14 |grep differ |awk '{print "diff -u <(sort " $2 ")","<(sort " $4 ")"}' |bash |less
# ^^ no files truly differ
```

What needs to happen next?

dshemetov:
Oh you're right, this is good news. So what I would like to double check is which signals _were_ having differences when I diffed the folders yesterday. There might be some numerical issues going on.

dshemetov:
I just reran the folder comparison analysis above and the only signal with differing files is `naat_pct_positive_7dav`.
image

Is there anything special about that signal? Here is a sample deletion from `20220228_county_covid_naat_pct_positive_7dav.csv`
image

Here is another example `20220302_county_covid_naat_pct_positive_7dav.csv`, where the majority of the file is changed, but as you can see the changes to the values are very minor (~1%). Could it be a numerical issue?
image

dshemetov:
So I suppose, we wait until the naats logic is fixed and then rerun these comparisons?

krivard:
Okay, you're right that something is definitely horked in interpolation, and for whatever reason it's only showing up with naats.

I aspirationally merged Nat's naats fix into a spare branch, like this:

```
$ git checkout ds/dsew-interpolation
$ git checkout -b krivard/dsew-interpolation
$ git merge ndefries/cpr-lenient-check-ts-per-publishdate
```

(don't merge those changes into `ds/dsew-interpolation` yet, in case Ananya finds something and Nat needs to rebase. this is purely a throwaway merge)

Then I ran each branch (`krivard/dsew-interpolation` and `ndefries/cpr-lenient-check-ts-per-publishdate`) twice: once with `reports` set to `2022-03-09--2022-03-11` and once with `reports` set to `2022-03-11--2022-03-14`. The smaller ranges are necessary so that we can avoid overlaps unique to the naats data.

Here are the available files in those two ranges, along with the positivity and volume reference dates they contain:

| publish date | week | positivity refdate | volume refdate |
| - | - | - | - |
|2022-03-09|previous|2022-02-27|2022-02-23|
|2022-03-09|last|2022-03-06|2022-03-02|
|2022-03-10|previous|2022-02-28|2022-02-24|
|2022-03-10|last|2022-03-07|2022-03-03|
|2022-03-11|previous|2022-03-01|2022-02-25|
|2022-03-11|last|2022-03-08|2022-03-04|
|2022-03-14|previous|2022-03-04|2022-02-28|
|2022-03-14|last|2022-03-11|2022-03-07|

Sorted by positivity refdate (which is the refdate used for output) we get:

| publish date | positivity refdate | volume refdate |
| - | - | - | 
|2022-03-09|2022-02-27|2022-02-23|
|2022-03-10|2022-02-28|2022-02-24|
|2022-03-11|2022-03-01|2022-02-25|
|2022-03-14|2022-03-04|2022-02-28|
|2022-03-09|2022-03-06|2022-03-02|
|2022-03-10|2022-03-07|2022-03-03|
|2022-03-11|2022-03-08|2022-03-04|
|2022-03-14|2022-03-11|2022-03-07|

There are no backfill entries across these two ranges: no reference date is reissued with updated data.

Between the two ranges (9-11 and 11-14) there is one overlapping file, 2022-03-11. If the code were working correctly, the output from that file (refdates 2022-03-01 and 2022-03-08) should match between the two runs. Right?

The output from `ndefries/cpr-lenient-check-ts-per-publishdate` matches, so I'm pretty sure we're not dealing with a logic bug in the naats handling per se. 

The output from `krivard/dsew-interpolation` does not match. Sample diff (sorted; header manually edited for readability):

```
--- receiving_krivard_dsew-interpolation_2022-03-09--2022-03-11/20220301_county_covid_naat_pct_positive_7dav.csv 2022-03-31 17:46:26.517258105 -0400
+++ receiving_krivard_dsew-interpolation_2022-03-11--2022-03-14/20220301_county_covid_naat_pct_positive_7dav.csv 2022-03-31 17:46:26.517258105 -0400
@@ -347,7 +347,6 @@
 13055,0.046,NA,26.285714285714285
 13057,0.045,NA,383.2857142857143
 13059,0.042,NA,171.14285714285714
-13061,0.15,NA,15.535714285714285
 13063,0.013,NA,2531.8571428571427
 13067,0.038,NA,1430.4285714285713
 13069,0.051,NA,34.0
@@ -450,7 +449,6 @@
 13299,0.078,NA,27.857142857142858
 13303,0.079,NA,10.857142857142858
 13305,0.057,NA,23.0
-13309,0.0578571,NA,6.6938775510204085
 13311,0.068,NA,42.0
 13313,0.03,NA,264.0
 13317,0.0,NA,6.142857142857143
@@ -752,7 +750,6 @@
 20173,0.044,NA,686.4285714285714
 20175,0.047,NA,10.0
 20177,0.052,NA,382.2857142857143
-20181,0.0335,NA,6.0476190476190474
 20189,0.032,NA,6.571428571428571
 20191,0.071,NA,19.714285714285715
 20193,0.234,NA,6.0
[...]
```

Hopefully that gives you a solid place to start looking for the cause?

dshemetov:
Hm, ok. So the mismatch is still: a) unique to interpolation, b) unique to naats signal.

So just to make sure I have the terminology down: publish_date is the date on the file, positivity_refdate is the date in the positivity data, and volume_refdate is the date in the sample size data? Can you say more about how the positivity and volume data are used? Do we compute ratios for any signals by matching positivity_refdate and volume_refdate across files?

krivard:
Correct on publish date, positivity refdate, and volume refdate. 

Consult https://github.com/cmu-delphi/covidcast-indicators/pull/1562 and/or Nat for how the two refdates are used -- the TL;DR is that we match volume to positivity only within each file, and only within each week ("last week" or "previous week").

nmdefries:
I haven't been following this super closely, but is the missing geo perhaps related to [this](https://github.com/cmu-delphi/covidcast-indicators/issues/1523)?

dshemetov:
@nmdefries shouldn't be, since we are comparing two `receiving` outputs from the same branch using the same geomapper source. The only thing changing in our comparisons is the time window over which the indicator aggregates input files.

dshemetov:
TL;DR: I found two concrete examples of the ways the deletions and updates occurred. It's basically what was expected:
1. the data contains is in [a, b] and [b, c] and so the interpolation for both doesn't cover the same values
2. the data contains interleaved data and so the interpolation imputes different values

### Case 1
Let's look at the county "13061" for the `naat positivity` signal. Here is what the data looks like:
```py
# 2022-03-09--2022-03-11
            val        se  sample_size
timestamp                             
2022-02-27  0.2  0.100905    15.714286
2022-03-07  0.0  0.000000    15.000000
2022-03-08  0.0  0.000000    15.285714

# 2022-03-11--2022-03-14
            val   se  sample_size
timestamp                        
2022-03-08  0.0  0.0    15.285714
2022-03-11  0.0  0.0    15.285714
```

When smoothed, we get the following
```py
# 2022-03-09--2022-03-11
              val        se  sample_size
timestamp                               
2022-02-27  0.200  0.100905    15.714286
2022-02-28  0.175  0.088292    15.625000
2022-03-01  0.150  0.075679    15.535714
2022-03-02  0.125  0.063066    15.446429
2022-03-03  0.100  0.050452    15.357143
2022-03-04  0.075  0.037839    15.267857
2022-03-05  0.050  0.025226    15.178571
2022-03-06  0.025  0.012613    15.089286
2022-03-07  0.000  0.000000    15.000000
2022-03-08  0.000  0.000000    15.285714

# 2022-03-11--2022-03-14
            val   se  sample_size
timestamp                        
2022-03-08  0.0  0.0    15.285714
2022-03-09  0.0  0.0    15.285714
2022-03-10  0.0  0.0    15.285714
2022-03-11  0.0  0.0    15.285714
```

This is a simple case: the only overlap for the data is the date 2022-03-08, where the values are the same. Clearly the rest of the values won't show up in receiving and will be viewed as deletions for this geo.

### Case 2
Here is another geo "13059", with a less trivial case:
```py
# 2022-03-09--2022-03-11
              val        se  sample_size
timestamp                               
2022-02-27  0.050  0.015800   190.285714
2022-02-28  0.045  0.015458   179.857143
2022-03-01  0.042  0.015333   171.142857
2022-03-06  0.015  0.008446   207.142857
2022-03-07  0.011  0.006831   233.142857
2022-03-08  0.011  0.006637   247.000000

# 2022-03-11--2022-03-14
              val        se  sample_size
timestamp                               
2022-03-01  0.042  0.015333   171.142857
2022-03-04  0.021  0.010839   175.000000
2022-03-08  0.011  0.006637   247.000000
2022-03-11  0.022  0.009534   236.714286
```

When smoothed, we get
```py
# 2022-03-09--2022-03-11
               val        se  sample_size
timestamp                                
2022-02-27  0.0500  0.015800   190.285714
2022-02-28  0.0450  0.015458   179.857143
2022-03-01  0.0420  0.015333   171.142857
2022-03-02  0.0366  0.013956   178.342857
2022-03-03  0.0312  0.012578   185.542857
2022-03-04  0.0258  0.011201   192.742857
2022-03-05  0.0204  0.009823   199.942857
2022-03-06  0.0150  0.008446   207.142857
2022-03-07  0.0110  0.006831   233.142857
2022-03-08  0.0110  0.006637   247.000000

# 2022-03-11--2022-03-14
                 val        se  sample_size
timestamp                                  
2022-03-01  0.042000  0.015333   171.142857
2022-03-02  0.035000  0.013835   172.428571
2022-03-03  0.028000  0.012337   173.714286
2022-03-04  0.021000  0.010839   175.000000
2022-03-05  0.018500  0.009788   193.000000
2022-03-06  0.016000  0.008738   211.000000
2022-03-07  0.013500  0.007687   229.000000
2022-03-08  0.011000  0.006637   247.000000
2022-03-09  0.014667  0.007602   243.571429
2022-03-10  0.018333  0.008568   240.142857
2022-03-11  0.022000  0.009534   236.714286
```

Here the data from the two date ranges is interleaved: both have 2022-03-01 and 2022-03-08, but only the first has 2022-03-06 and only the second has 2022-03-04. This naturally causes the interpolations on 2022-03-02, 2022-03-03, 2022-03-05, 2022-03-07 to be different.

### Conclusions
- The first case seems like an unfortunate case of inconsistent updates across geos.
- The second case seems like a legitimate data update and it seems reasonable to add values in that case.

Ideally we could get the archiver to ignore the first case and not delete values. I'm not sure how we can make that happen without adding even more complex logic here.

krivard:
### Case 1

> This is a simple case: the only overlap for the data is the date 2022-03-08, where the values are the same. Clearly the rest of the values won't show up in receiving and will be viewed as deletions for this geo.

A deletion is only detected when a file is present in both places, but a row is missing -- if the rest of the values are in files that aren't output, we shouldn't get any deletions. In this particular case, the 11-14 run has output coverage over reference dates from 3/01 through 3/11 via other regions, so we will get deletions for 3/01-3/07, but not for 2/27-2/28.

### Case 2

I agree with your reasoning that the interpolator is working as intended here, but I don't like the idea of replacing the report-confirmed values for 3/06 and 3/07 with interpolated ones. This probably means extending the date range we load in each run, but that's going to get tricky with the naats matching algorithm Nat developed. If we can get it to work though, that will probably *also* take care of the mistaken deletions problem from Case 1.

### What next

* write up the problem for Ryan, since this is more complicated than we anticipated
* get you, me, Nat, and maybe Ananya in a meeting to figure out how to get interpolation to activate *only* for reference dates where we have no data, while maintaining our ability to match total naats and naats positivity by week

dshemetov:
Case 1: good clarification. The deletions come from the inconsistency in reporting across geos for overlapping dates.
Case 2: while extending the range will solve the issue of replacing a known value with an interpolation, I am skeptical that it will solve the problem of case 1. I expect that there will data files with publish dates on the boundaries of the range that will present inconsistent reference dates across geos and lead to the same artificial deletion phenomena. 

The archiver limits us because it was designed to handle indicators that output either full date ranges or partial date ranges on each run, but it was not designed to handle partial geo reports.

krivard:
> I expect that there will data files with publish dates on the boundaries of the range that will present inconsistent reference dates across geos and lead to the same artificial deletion phenomena.

If we ran with `reports: "all"` each day, we wouldn't have boundaries 😄  

...but the running time would slowly devour the world, so that's probably not the solution we want long-term

dshemetov:
Yeaaaa, I don't think that's feasible even now - opening and interpolating just 7 files takes ~5 minutes on my machine, so if the full history is 300+ files, it would take hours.

dshemetov:
@krivard I think this is ready for a final review, once I figure out what's failing with the build.

krivard:
pretty sure it's just the linter

dshemetov:
@krivard the linter is happy! i ran this with a modification of your
```
for branch in ds/dsew-interpolation; do git checkout $branch; for r in 2022-03-05--2022-03-11 2022-03-08--2022-03-14; do rcv="receiving_${branch/\//_}_$r"; mkdir $rcv ||break;  env/bin/python -m delphi_utils set indicator.reports $r common.export_dir $rcv; env/bin/python -m delphi_dsew_community_profile ||break; done; done;
```
and found that there were no changed files, only new and removed in the diff. 

maybe to be safe i should try this on even more date windows...

dshemetov:
Alright, I started running this 
```
date_ranges=()
for i in {1..60}; do
    date_ranges[i]=$(date -I -d "2022-01-01 +$i days")--$(date -I -d "2022-01-01 +$(expr $i + 5) days")
done

for branch in ds/dsew-interpolation; do git checkout $branch; for r in "$date_ranges[@]"; do rcv="receiving_${branch/\//_}_$r"; mkdir $rcv ||break;  env/bin/python -m delphi_utils set indicator.reports $r common.export_dir $rcv; env/bin/python -m delphi_dsew_community_profile ||break; done; done;
```
and was going to diff all consecutive folders, just to make sure and I already found something strange.

Setting reports to "2022-01-02--2022-01-06" breaks because the 2021-12-30 publish date only has 1 reference value, but `add_max_ts_col` expects 2. The strange thing is that the 2021-12-30 publish date file has 2 reference values and setting reports to "2022-01-02--2022-01-05" confirms this, since it runs just fine. My guess is that extending to 2022-01-06 somehow filters out reference values from the 2021-12-30 file.

dshemetov:
I think I see what happened. The 2021-12-30 file has two reference dates: 2021-12-27 and 2021-12-20. The 2022-01-06 file also has the 2021-12-27 reference date, so when we hit this line of code, the 2022-01-06 version overwrites the 2021-12-30 one and leaves it with a single reference date.
```
latest_sig_df = pd.concat(
            lst
        ).groupby(
            "timestamp"
        ).apply(
            lambda x: x[x["publish_date"] == x["publish_date"].max()]
        ).drop_duplicates(
        )
```

Aaand it turns out that Nat already fixed this issue and I just haven't merged from main into this branch in a while.

krivard (approved):
👍

PR #1557, opened 2022-03-15
[CITS QSF utils] Add new tool to print matrix item names

nmdefries:
### Description
To help documentation efforts, this creates a new QSF tool that prints a list of survey questions that use matrix format (or that we treat as such, like matrices of text question, which Qualtrics labels as "Text" format). To avoid duplicating code between the new tool and the codebook script, do some refactoring and put the new functions into the utils file.

Add UMD/CMU flag to codebook.

### Changelog
- `qsf-tools/generate-codebook.R`
- `qsf-tools/list-matrix-items.R`
- `qsf-tools/qsf-utils.R`


capnrefsmmat (approved):
Looks fine to me, but I'm not sure I'm able to thoroughly review the QSF tools, since I don't use them. But since they're not in production systems, it's probably okay if you test yourself and I just check for obvious things

nmdefries:

PR #1562, opened 2022-03-23
Fix CPR bugs associated with non-unique timestamps by publish date

### LINKED ISSUE #1556: DSEW-CPR testing signals fail for Community_Profile_Report_20220225
Got the following output while trying out the new testing signals on staging:

```
[...]
{"filename": "./input_cache/Community Profile Report 20220225--539a654c-7b48-4730-b4c2-34f1c64b5d49.xlsx", "event": "Downloading file", "logger": "delphi_dsew_community_profile.run", "level": "info", "timestamp": "2022-03-11T15:42:11.893222Z"}
{"sheet": "Regions", "filename": "./input_cache/Community Profile Report 20220225--539a654c-7b48-4730-b4c2-34f1c64b5d49.xlsx", "event": "Building dfs", "logger": "delphi_dsew_community_profile.run", "level": "info", "timestamp": "2022-03-11T15:42:12.857682Z"}
{"sheet": "States", "filename": "./input_cache/Community Profile Report 20220225--539a654c-7b48-4730-b4c2-34f1c64b5d49.xlsx", "event": "Building dfs", "logger": "delphi_dsew_community_profile.run", "level": "info", "timestamp": "2022-03-11T15:42:12.972835Z"}
{"sheet": "CBSAs", "filename": "./input_cache/Community Profile Report 20220225--539a654c-7b48-4730-b4c2-34f1c64b5d49.xlsx", "event": "Building dfs", "logger": "delphi_dsew_community_profile.run", "level": "info", "timestamp": "2022-03-11T15:42:13.219382Z"}
{"sheet": "Counties", "filename": "./input_cache/Community Profile Report 20220225--539a654c-7b48-4730-b4c2-34f1c64b5d49.xlsx", "event": "Building dfs", "logger": "delphi_dsew_community_profile.run", "level": "info", "timestamp": "2022-03-11T15:42:15.193136Z"}
{"event": "Top-level exception occurred", "logger": "delphi_dsew_community_profile.run", "level": "error", "timestamp": "2022-03-11T15:42:23.090096Z", "exception":[reformatted below]}
    exec(code, run_globals)
  File "/mnt/data/indicators/runtime/dsew_community_profile/delphi_dsew_community_profile/__main__.py", line 12, in <module>
    run_module(read_params())  # pragma: no cover
  File "/mnt/data/indicators/runtime/dsew_community_profile/delphi_dsew_community_profile/run.py", line 60, in run_module
    dfs = fetch_new_reports(params, logger)
  File "/mnt/data/indicators/runtime/dsew_community_profile/delphi_dsew_community_profile/pull.py", line 529, in fetch_new_reports
    ret[key] = unify_testing_sigs(
  File "/mnt/data/indicators/runtime/dsew_community_profile/delphi_dsew_community_profile/pull.py", line 603, in unify_testing_sigs
    volume_df = add_max_ts_col(volume_df)[
  File "/mnt/data/indicators/runtime/dsew_community_profile/delphi_dsew_community_profile/pull.py", line 641, in add_max_ts_col
    assert all(
AssertionError: Testing signals should have two unique timestamps per publish date-region combination.
```

A quick look at the excel file (attached) suggests a logic error somewhere, since the report does include both LAST WEEK and PREVIOUS WEEK for testing signals.

[Community_Profile_Report_20220225.xlsx](https://github.com/cmu-delphi/covidcast-indicators/files/8234901/Community_Profile_Report_20220225.xlsx)

nmdefries:
### Description
Fix two CPR bugs:

- Some asserts are too strict given earlier logic.
- On 2021-03-17, the reference dates for test volume and test positivity changed from being the same within a report to being different. This caused some timestamps to appear for multiple report dates, which the existing logic can't handle appropriately. Fix by splitting data into before/after that date and processing separately.

### Changelog
- `pull.py`
- tests

### Fixes 
Closes #1556.

### DIFF
--- a/dsew_community_profile/delphi_dsew_community_profile/pull.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/pull.py
@@ -515,7 +515,7 @@ def fetch_new_reports(params, logger=None):
     for key, lst in datasets.items():
         (_, sig, _) = key
         latest_key_df = pd.concat(lst)
-        if sig == "total":
+        if sig in ("total", "positivity"):
             latest_key_df = pd.concat(apply_thres_change_date(
                 keep_latest_report,
                 latest_key_df,
@@ -534,7 +534,7 @@ def fetch_new_reports(params, logger=None):
         if state_key not in ret:
             continue
 
-        if sig == "total":
+        if sig in ("total", "positivity"):
             nation_df = pd.concat(apply_thres_change_date(
                 nation_from_state,
                 ret[state_key].rename(columns={"geo_id": "state_id"}),
--- a/dsew_community_profile/delphi_dsew_community_profile/pull.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/pull.py
@@ -626,6 +626,7 @@ def unify_testing_sigs(positivity_df, volume_df):
     https://docs.google.com/document/d/1MoIimdM_8OwG4SygoeQ9QEVZzIuDl339_a0xoYa6vuA/edit#
 
     """
+    # check that we have positivity *and* volume for each publishdate+geo
     pos_groups_df = positivity_df.drop(
         ["val", "se", "sample_size", "timestamp"],
         axis=1
--- a/dsew_community_profile/delphi_dsew_community_profile/pull.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/pull.py
@@ -383,6 +383,7 @@ class Dataset:
                 })
                 for si in sig_select
             ])
+
         for sig in COUNTS_7D_SIGNALS:
             assert (sheet.level, sig, NOT_PROP) in self.dfs.keys()
             self.dfs[(sheet.level, sig, NOT_PROP)]["val"] /= 7 # 7-day total -> 7-day average
@@ -474,8 +475,18 @@ def nation_from_state(df, sig, geomapper):
     )
     df["se"] = None
     df["sample_size"] = None
+    # Recreate publish_date column
     df =  pd.merge(df, publish_date_by_ts, on="timestamp", how="left")
 
+    return df
+
+def keep_latest_report(df):
+    df = df.groupby(
+            "timestamp"
+        ).apply(
+            lambda x: x[x["publish_date"] == x["publish_date"].max()]
+        ).drop_duplicates(
+        )
     return df
 
 def fetch_new_reports(params, logger=None):
@@ -486,27 +497,33 @@ def fetch_new_reports(params, logger=None):
     datasets = download_and_parse(listing, logger)
     # collect like signals together, keeping most recent publish date
     ret = {}
-    for sig, lst in datasets.items():
-        latest_sig_df = pd.concat(
-            lst
-        ).groupby(
-            "timestamp"
-        ).apply(
-            lambda x: x[x["publish_date"] == x["publish_date"].max()]
-        ).drop_duplicates(
-        )
 
-        if len(latest_sig_df.index) > 0:
-            latest_sig_df = latest_sig_df.reset_index(drop=True)
-            assert all(latest_sig_df.groupby(
+    def generate_assert_check_value(df):
+        return all(df.groupby(
                     ["timestamp", "geo_id"]
                 ).size(
                 ).reset_index(
                     drop=True
-                ) == 1), f"Duplicate rows in {sig} indicate that one or" \
+                ) == 1)
+
+    for key, lst in datasets.items():
+        (_, sig, _) = key
+        latest_key_df = pd.concat(lst)
+        if sig == "total":
+            latest_key_df = pd.concat(apply_thres_change_date(keep_latest_report, latest_key_df))
+        else:
+            latest_key_df = keep_latest_report(latest_key_df)
+
+        if len(latest_key_df.index) > 0:
+            latest_key_df = latest_key_df.reset_index(drop=True)
+            if sig == "total":
+                assert_check_value = all(apply_thres_change_date(generate_assert_check_value, latest_key_df))
+            else:
+                assert_check_value = generate_assert_check_value(latest_key_df)
+            assert assert_check_value, f"Duplicate rows in {key} indicate that one or" \
                 + " more reports were published multiple times and the copies differ"
 
-            ret[sig] = latest_sig_df
+            ret[key] = latest_key_df
 
     # add nation from state
     geomapper = GeoMapper()
@@ -514,11 +531,20 @@ def fetch_new_reports(params, logger=None):
         state_key = ("state", sig, NOT_PROP)
         if state_key not in ret:
             continue
-        ret[("nation", sig, NOT_PROP)] = nation_from_state(
-            ret[state_key].rename(columns={"geo_id": "state_id"}),
-            sig,
-            geomapper
-        )
+
+        def nation_from_state_wrapper(df):
+            """Call nation_from_state with sig and geomapper info already set."""
+            return nation_from_state(
+                    df.rename(columns={"geo_id": "state_id"}),
+                    sig,
+                    geomapper
+                )
+
+        if sig == "total":
+            nation_df = pd.concat(apply_thres_change_date(nation_from_state_wrapper, ret[state_key]))
+        else:
+            nation_df = nation_from_state_wrapper(ret[state_key])
+        ret[("nation", sig, NOT_PROP)] = nation_df
 
     for key, df in ret.copy().items():
         (geo, sig, prop) = key
@@ -535,6 +561,8 @@ def fetch_new_reports(params, logger=None):
             # No longer need "total" signal.
             del ret[total_key]
         elif sig != "total":
+            # TODO: pretty sure this is duplicate code. Should be able to take
+            # the `ret[key]` assignment out of the conditional.
             # If signal is not test volume or test positivity, we don't need
             # publish date.
             df = df.drop("publish_date", axis=1)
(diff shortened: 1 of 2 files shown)


dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -486,39 +511,43 @@ def fetch_new_reports(params, logger=None):
...
+            latest_key_df = pd.concat(apply_thres_change_date(
+                keep_latest_report,
+                latest_key_df,
+                [sig] * 2
+            ))
+        else:
+            latest_key_df = keep_latest_report(latest_key_df, sig)
 
-            ret[sig] = latest_sig_df
+        if len(latest_key_df.index) > 0:
```
krivard, line 527: Is this the same or different from `not latest_key_df.empty`?

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -597,9 +626,29 @@ def unify_testing_sigs(positivity_df, volume_df):
     https://docs.google.com/document/d/1MoIimdM_8OwG4SygoeQ9QEVZzIuDl339_a0xoYa6vuA/edit#
 
     """
-    # Combine test positivity and test volume, maintaining "this week" and "previous week" status.
+    pos_groups_df = positivity_df.drop(
```
krivard, line 629:
```suggestion
    # check that we have positivity *and* volume for each publishdate+geo
    pos_groups_df = positivity_df.drop(
```

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -597,9 +626,29 @@ def unify_testing_sigs(positivity_df, volume_df):
...
+    )
+    merged = pos_groups_df.merge(
+        vol_groups_df,
+        on=["geo_id", "publish_date"],
+        how="outer",
+        indicator=True
+    )
+    assert all(merged["_merge"] == "both"), \
+        "Each publish date-geo value combination should be available for both " + \
+        "test positivity and test volume."
```
krivard, line 647: Community_Profile_Report_20220225.xlsx has 158 counties where positivity is not available for "last week" but volume is, and 7 counties where volume is not available but positivity is. Will this assert inappropriately reject that file? Or do you mean that previous steps should have already dropped the entries with mismatched availability?

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -628,8 +684,8 @@ def add_max_ts_col(df):
     """
     Add column to differentiate timestamps for a given publish date-geo combo.
 
-    Each publish date is associated with two timestamps for test volume and
-    test positivity. The older timestamp corresponds to data from the
+    Each publish date is associated with up to two timestamps for test volume
+    and test positivity. The older timestamp corresponds to data from the
```
krivard, line 688: does this change mean that you found files that only have one week of test data in them, instead of both "last week" and "previous week"?

krivard (commented):
Some missing documentation, and I may also be getting confused about one context of two reference dates ("last week" "previous week") and another ("last week" positivity vs volume after 2021-03-17)

dsew_community_profile/delphi_dsew_community_profile/pull.py
nmdefries: In short, no, this is the bit that actually fixes issue #1556.

Previously, we expected that "[t]esting signals should have two unique timestamps per publish date-region combination." This is true of _raw_ data, but causes false positives for our processed data.

This particular check [is called in `unify_testing_sigs`](https://github.com/cmu-delphi/covidcast-indicators/blob/eaeda1d87b644726cc2f997cbb80db3869653732/dsew_community_profile/delphi_dsew_community_profile/pull.py#L558-L559) after we [deduplicate timestamps](https://github.com/cmu-delphi/covidcast-indicators/blob/eaeda1d87b644726cc2f997cbb80db3869653732/dsew_community_profile/delphi_dsew_community_profile/pull.py#L518-L525). We only keep those with the most recent publish date, which means that some publish date-timestamp combinations are dropped.

Specifically, if we're generating output over a wide enough time period so that we see a given date reported for "last week" and also for the "previous week", we will always keep the version reported for the "previous week" since it's publish date will be 7 days newer. As a result, the older publish date will have all of it's "last week" data discarded; it will only be associated with the "previous week"'s timestamp.

krivard: Are there ever cases where only one timestamp is available for volume but two are available for positivity, or vice-versa?

dsew_community_profile/delphi_dsew_community_profile/pull.py
nmdefries: The wording may be ambiguous on my part. "available" here means that there is a row for a given region and date. It doesn't check if the signal is `NA`or not. `pandas.read_excel` reads empty fields as `NA/NaN`, so obs are included as `NA` where the signal was not available in the original report.

The way the sample size filter is set up at the moment, any obs with missing (`NA`) n or n <= 5 are dropped.

nmdefries: Rows are included if `val` (test positivity) is missing but sample size is available and > 5.

dsew_community_profile/delphi_dsew_community_profile/pull.py
nmdefries: * checks that each publish date has the same geos for test volume and test positivity [here](https://github.com/cmu-delphi/covidcast-indicators/blob/ndefries/cpr-lenient-check-ts-per-publishdate/dsew_community_profile/delphi_dsew_community_profile/pull.py#L646-L648)
* checks that test volume and positivity have the same number of observations. Since we've already checked that they have the same geos by publish date, this ~checks that they have the same number of ts per publish date-geo.

nmdefries: Assuming there are always the same number of timestamps available for test volume and positivity in the raw data, there should always be the same number in the processed data. The only logic that influences this is the timestamp deduplication (where we keep the most recent publish date for each timestamp). It is [applied to both test volume and positivity](https://github.com/cmu-delphi/covidcast-indicators/blob/ndefries/cpr-lenient-check-ts-per-publishdate/dsew_community_profile/delphi_dsew_community_profile/pull.py#L518-L523) using the same ["change date" threshold approach](https://github.com/cmu-delphi/covidcast-indicators/blob/4eee156c887eed6e25f1ddd0f9add27cd3213531/dsew_community_profile/delphi_dsew_community_profile/pull.py#L750).

This isn't strictly necessary for test positivity, but it makes it easier to match up volume and positivity signals for each publish date.

Empirically, running the pipeline over all reports with a new check for # of timestamps, I don't see any cases where the number of timestamps differs.

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -682,3 +761,37 @@ def std_err(df):
...
+    apply_fn: function
+        function to apply to data before and after the test volume change date
+    df: pd.DataFrame
+        Columns: val, sample_size, ...
+    apply_fn_args: tuple of lists
+        variable number of additional arguments to pass to the `apply_fn`.
+        Each additional argument should be a list of length 2. The first
+        element of each list will be passed to the `apply_fn` when processing
+        pre-change date data; the second element will be used for the
+        post-change date data.
```
krivard, line 783: it looks like we never actually call this with different arguments to be used before vs after the change date. was that intended?

nmdefries: Yeah, the before/after change date data is always processed using the same args.

The alternative implementation is to pass single values as `*args` instead of lists, and duplicate them to the right length within `apply_thres_change_date`.

krivard (approved):
lgtm! 👍  want me to wait for ananya before merging or no?

Ananya-Joshi:
Tests and some sample dates I looked at from march 2022 seem to be working fine/generally in the right range. Happy to approve pending some information on the comments above. Good job Nat!

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -638,11 +710,18 @@ def add_max_ts_col(df):
...
+        num_unique_obs=("timestamp", "nunique")
+    )
     assert all(
-        df.groupby(["publish_date", "geo_id"])["timestamp"].count() == 2
+        assert_df.num_obs <= 2
     ) and all(
-        df.groupby(["publish_date", "geo_id"])["timestamp"].nunique() == 2
-    ), "Testing signals should have two unique timestamps per publish date-region combination."
+        assert_df[assert_df.num_obs == 2].num_unique_obs == 2
+    ), "Testing signals should have up to two timestamps per publish date-geo level " + \
```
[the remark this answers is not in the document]
nmdefries: The first clause of the assert is checking that each publish date-geo combo should have up to 2 observations (AKA timestamps). The second clause is checking that when a publish date-geo combo has 2 observations/timestamps, there are also 2 _unique_ timestamps (of course, this is trivially true in the case that a publish date-geo combo only has 1 observation/timestamp).

Another way to check the second bit would be seeing if `all(assert_df.num_obs == assert_df.num_unique_obs)`. I can add that in if you find it clearer.

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -682,3 +761,37 @@ def std_err(df):
     p = df.val
     n = df.sample_size
     return np.sqrt(p * (1 - p) / n)
+
+def apply_thres_change_date(apply_fn, df, *apply_fn_args):
+    """
+    Apply a function separately to data before and after the test volume change date.
+
+    The test volume change date is when test volume and test positivity started being
+    repported for different reference dates within the same report.
```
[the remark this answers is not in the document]
nmdefries: That's right, it started on [2021-03-17](https://github.com/cmu-delphi/covidcast-indicators/blob/3d1e60aadecc10415238959ad0cf4abedb71a3a2/dsew_community_profile/delphi_dsew_community_profile/pull.py#L790).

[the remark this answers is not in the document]
Ananya-Joshi: Sorry, I meant there was a typo in reported and that it might be good to put that date in this comment!

dsew_community_profile/delphi_dsew_community_profile/pull.py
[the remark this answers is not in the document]
Ananya-Joshi: Ah now I understand so it's all okay. I think your suggestion would make it clearer even in the case of 1 observation/timestamp.

Ananya-Joshi (approved):

PR #1563, opened 2022-03-29
Suppress sirCAL warnings for CTIS `vaccine_barrier_*_tried` and weighted `dontneed_reason_*` signals

nmdefries:
### Description
Due to low sample size, `vaccine_barrier_*_tried` signals are no longer available at the county or state levels (in addition to the previously unavailable MSA data), and weighted `dontneed_reason_*` signals are no longer available at the MSA level (in addition to the previously unavailable HRR data).

### Changelog
- sirCAL Ansible template params
- local template params


krivard (approved):
👍

PR #1564, opened 2022-03-30
[CTIS qsf tools] For changed matrix subquestions, the differ lists the particular items that changed

nmdefries:
### Description
The diff output now includes an `impacted_subquestions` column to indicate which particular matrix subquestion changed, if any. Subquestions are indicated with the full output name (carried forward from other questions and recoded with `ChoiceDataExportTags`, if provided). This will make it easier to join the annotated diff onto the codebook for the purpose of generating a changelog document.

### Changelog
- `qsf-differ.R`

### DIFF
--- a/facebook/qsf-tools/qsf-differ.R
+++ b/facebook/qsf-tools/qsf-differ.R
@@ -57,7 +57,6 @@ get_qsf_file <- function(path,
 ) {
   # Read file as json.
   qsf <- read_json(path)
-  
   ## Block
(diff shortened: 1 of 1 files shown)


nmdefries:
@leonardzh This adds an extra column to the diff export.

capnrefsmmat (approved):
As with other QSF tools, approving on the assumption it'll get tested thoroughly when we try it on both surveys

nmdefries:

PR #1572, opened 2022-04-08
[CTIS] Add weekly weights to microdata

nmdefries:
### Description
Add two new fields to the microdata.

`weight_wf` are the weekly "full" (new definition) weights; `weight_wp` are the weekly "partial" (old "full" definition) weights. Each of these weights files covers a week period, Sunday through Saturday. Weekly weight files are usually released 1-3 days after the end of the week; the "normal max" is 5-6 days and files have been uploaded up to 9 days after the end of the week.

This is set up so that every time the pipeline is run, it attempts to add weekly weights. If no appropriate weekly weights are available (i.e. the weekly weights aren't recent enough), then those columns are `NA` but are still reported in the microdata. Them being missing doesn't impact an observation's inclusion in the microdata.

The new `make` targets are structured assuming that we'll be running the `WEEKLY_WEIGHTS` target every day. If new weekly weights are available, the weekly weights pipeline is run.

We will need to delay the monthly roll-up process to account for the several days past the end of the week it takes to receive weekly weights. I'm assuming this is set in a cron job.

### Changelog
- Makefile -- add new targets and variables for weekly weights pipeline, and generalize existing targets (e.g. `tidy`) to work in the weekly case.
- Test params files, and local and production template files -- add weekly weights dir field.
- `contingency_run.R`, `test-contingency-run.R`, `test-integration.R`, `test-weights.R` -- changed name of `join_weights` to `add_weights` since, with refactoring, it acts more like a wrapper.
- `run.R` -- change name of `join_weights`; Add flag indicating that we want to include weekly weights in the microdata.
- `weights.R` -- 
    - Rename `join_weights` to `add_weights`
    - Add boolean arg to indicate if want to add weekly weights or not. `FALSE` (do not add) by default. If `TRUE`, we load weekly weights from the params-provided weekly weights dir and join `full` (new definition) and `partial` (old `full` definition) them on. The normal `weight` column is temporarily renamed to avoid overwriting it.
    - Since we now have three join steps, factor that out into a new function `load_and_join_weights`
- `test-gold-individual.R` -- create new weekly weights fields to the reference data.


facebook/Makefile
```diff
@@ -133,6 +148,23 @@ $(WEIGHTS): $(TODAY)
...
 
+$(WEEKLY_WEIGHTS): $(TODAY)
+	[ -f $(WEEKLY_WEIGHTS) ] || mkdir -p $(WEEKLY_WEIGHTS)
+	cd "$(WEEKLY_WEIGHTS)"; \
+	BATCH="cd fb-interchange/cmu_respondent_ww_weights\nls -1"; \
+	NEW=`LC_ALL=C comm -23 <(sshpass -p $(DELPHI_SURVEY_SFTP_PASSWORD) sftp $(SFTP_OPTIONS) -b <(echo -e "$${BATCH}") -P 2222 $(DELPHI_SURVEY_SFTP_USER) | grep "^202" | LC_ALL=C sort) <(ls -1 | LC_ALL=C sort)`; \
+	echo "New weekly weights files:"; \
+	echo $${NEW}; \
+	if [[ $$ncids == "0" ]]; then \
+		echo "ERROR: 0 CIDs reported for $$f"; \
```
nmdefries, line 159: Change error text since we're not uploading CIDs for weekly.

facebook/Makefile
```diff
@@ -114,6 +118,17 @@ params.json: $(TODAY)
 		output cids,individual,covidalert,archive,community \
 		start_date $(YESTERDAY)
 
+params.weekly-weights.json: $(TODAY)
+	PAT=`grep fb-survey params.json | awk 'BEGIN{FS="\""}{print $$2}' | sed 's/ /_/g;s/^/-e /'`; \
+	$(PYTHON) -m delphi_utils set \
+		debug false \
+		produce_individual_raceeth true \
+		end_date $(LAST_SATURDAY) \
+		input <(find $(QUALTRICS) -maxdepth 1 -newer $(LAST_SUNDAY) -not -newer $(shell date -d "$(LAST_SATURDAY) +2 day" +"%Y-%m-%d") -type f -name "*.csv" | sort | grep $${PAT}  | tr '\n' ',' | sed 's_$(QUALTRICS)/__g;s/,$$//' ) \
```
nmdefries, line 127: Don't need to load the full range of dates for the previous week since the archive contains the last 14 days of data. If the weekly weights are too delayed (more than 7 days), however, the archive won't contain enough data. Not sure how to deal with that. Should we assume that we'll manually regenerate in that case?

krivard: Is this solvable by increasing the depth of the archive to 17 days, or is there a concern that that will have other repercussions?

krivard (approved):
Looks good modulo the switch to $(MAKE).

To roll this out what setup will we need in prod? any new volumes for the docker run command? current contents are

```
[indicators@bigchunk-dev-02 facebook]$ less docker_run.sh 
#!/bin/bash

workdir=$1
shift
DRY=$1
shift
LOG=$1
shift

DOCKER_IMAGE="ghcr.io/cmu-delphi/covidcast-indicators-facebook"
DOCKER_TAG="latest"
docker run \
    -v "${workdir}"/logs:/facebook/logs \
    -v "${workdir}"/receiving:/facebook/receiving \
    -v "${workdir}"/params.json:/facebook/params.json \
    -v "${workdir}"/individual:/facebook/individual \
    -v "${workdir}"/individual_raceeth:/facebook/individual_raceeth \
    -v "${workdir}"/tidy:/facebook/tidy \
    -v "${workdir}"/input:/facebook/input \
    -v "${workdir}"/fb-incoming:/facebook/fb-incoming \
    -v "${workdir}"/archive:/facebook/archive \
    -v "${workdir}"/fb-outgoing:/facebook/fb-outgoing \
    -v "${workdir}"/exp-fb-outgoing:/facebook/exp-fb-outgoing \
    -v "${workdir}"/micro:/facebook/micro \
    -v "${workdir}"/history:/facebook/history \
    -v "${workdir}"/static:/facebook/static \
    -v "${workdir}"/messages:/facebook/messages \
    -v "${workdir}"/scratch:/facebook/scratch \
    -e DRY="${DRY}" \
    -e LOG="${LOG}" \
    -i --rm "${DOCKER_IMAGE}:${DOCKER_TAG}" /bin/bash -c "$1"
```

capnrefsmmat (approved):
I think this looks good on the R side. Will be interesting to see the results of this once we release it

facebook/Makefile
nmdefries: Good idea. Yep, increasing the archive size would work _but_ would also require fancier logic elsewhere in the Makefile.

Right now, the pipeline is pretty simplistic. A run [always covers the previous week](https://github.com/cmu-delphi/covidcast-indicators/blob/ndefries/add-weekly-weights-microdata/facebook/Makefile#L128-L132) only, regardless of the dates included in the newly fetched weekly weights files. So if weekly weights are available 10 days after the end of the week, the weights will be be for 2 weeks ago but the pipeline will run for 1 week ago. The week before last will be skipped.

For fetched files to change the pipeline date range, the filenames will have to be parsed in the `WEEKLY_WEIGHTS` target and passed to `params.weekly-weights.json` through `weekly-weights-pipeline`. I'll look around to see how to do that.

nmdefries:
> what setup will we need in prod?

We need to add this line to `docker_run.sh`
```
    -v "${workdir}"/fb-incoming-weekly:/facebook/fb-incoming-weekly \
```
The microdata with weekly weights are saved to the normal microdata folders so there's nothing new to add there.

We need to delay the monthly roll up run (I'm assuming this is done as a cron job?) so that it waits for microdata with weekly weights to become available. I guess the worst case is that the last day of the month is on a Sunday, the first day of the weekly weighting range. The weights wouldn't be available until the next Thursday (5 days after the end of the week) at the "normal max" delay, which would be 11 days after the end of the month.

facebook/Makefile
```diff
@@ -133,6 +150,34 @@ $(WEIGHTS): $(TODAY)
 	  echo "WARNING: $${MSG}" >> $(MESSAGES); \
 	fi
 
+$(WEEKLY_WEIGHTS): $(TODAY)
+	# Want to schedule this to run every day. A pipeline run is triggered when new weekly weights files are available.
```
nmdefries, line 154: Since this is going to run daily, instead of scheduling a separate cron job, can the `WEEKLY_WEIGHTS` target be added as a final [dependency of the normal pipeline instead](https://github.com/cmu-delphi/covidcast-indicators/blob/ndefries/add-weekly-weights-microdata/facebook/Makefile#L194)? @krivard

nmdefries: And in that case would we still want separate messages and log files?

krivard: Yeah, it makes sense to add it to the normal daily run, and in that case I agree we don't need new messages or log files.

facebook/Makefile
krivard: Would it make sense to hold off on a generalized solution for now, and instead have it complain loudly if a new file is for a previous week?

facebook/Makefile
```diff
@@ -124,7 +125,7 @@ params.weekly-weights.json: $(TODAY)
 		debug false \
 		produce_individual_raceeth true \
 		end_date $(LAST_SATURDAY) \
-		input <(find $(QUALTRICS) -maxdepth 1 -newer $(LAST_SUNDAY) -not -newer $(shell date -d "$(LAST_SATURDAY) +2 day" +"%Y-%m-%d") -type f -name "*.csv" | sort | grep $${PAT}  | tr '\n' ',' | sed 's_$(QUALTRICS)/__g;s/,$$//' ) \
+		input <(find $(QUALTRICS) -maxdepth 1 -newer $< -type f -name "*.csv" | sort | grep $${PAT}  | tr '\n' ',' | sed 's_$(QUALTRICS)/__g;s/,$$//' ) \
```
krivard, line 128: hmm this will only capture the files that are downloaded the day the pipeline runs, which only contain survey responses from the last 7 days. depending on the day of the week, you might have as little as 1 day of output available for the time span from start_date $(LAST_SUNDAY) to end_date $(LAST_SATURDAY). is that the intended behavior?

facebook/Makefile
```diff
@@ -178,18 +197,18 @@ run-R: $(CIDS) $(CIDS_EXP)
...
 	chmod o+w $(MESSAGES)
 
-weekly-weights-pipeline: scratch init-qualtrics params.weekly-weights.json $(WEEKLY_WEIGHTS) run-R post-individual post-individual-raceeth tidy-weekly-weights
-	# Want to schedule on Wednesday or Thursday to be very likely to have received weekly weights.
+weekly-weights-pipeline: scratch init-qualtrics params.weekly-weights.json run-R post-individual post-individual-raceeth tidy-weekly-weights
 	touch $@
-	echo "SUCCESS: $(DRY_MESSAGE)weekly weights pipeline complete" >> $(WEEKLY_WEIGHTS_MESSAGES)
-	chmod o+w $(WEEKLY_WEIGHTS_MESSAGES)
+	echo "SUCCESS: $(DRY_MESSAGE)weekly weights pipeline complete" >> $(MESSAGES)
+	chmod o+w $(MESSAGES)
```
krivard, line 210: avoid collision with a check in `cron.sh` that controls whether the covidcast CSVs are posted to the server:
```suggestion
	echo "SUCCESS: $(DRY_MESSAGE)completed weekly weights pipeline" >> $(MESSAGES)
```

facebook/Makefile
```diff
@@ -178,18 +197,18 @@ run-R: $(CIDS) $(CIDS_EXP)
 	grep "scheduled core" tmp ; \
 	[ "$$?" -eq 1 ]
 
-pipeline: scratch init-qualtrics params.json $(WEIGHTS) run-R post-cids post-experimental-cids post-individual post-individual-raceeth post-done tidy
+pipeline: scratch init-qualtrics params.json $(WEIGHTS) run-R post-cids post-experimental-cids post-individual post-individual-raceeth post-done tidy $(WEEKLY_WEIGHTS)
 	grep $(TODAY) params.json
```
krivard, line 201: will this still succeed if params.json has been configured for weekly weights?

krivard (commented):
questions about params, small edit to success message

facebook/Makefile
nmdefries: The last 14 days of data should be loaded from the archive, so I don't think this will be a problem.

facebook/Makefile
nmdefries: hmmmm

nmdefries: Addressed in https://github.com/cmu-delphi/covidcast-indicators/pull/1572/commits/432eeab9597e5109039150d0cd2cb64ad9d5eed9 by saving an existing params file before running the weekly weights pipeline and restoring after the pipeline is finished.

PR #1581, opened 2022-04-15
[CTIS qsf tools] Grab bag of fixes and improvements

nmdefries:
### Description
- In CMU Wave 10, manually rename C6 -> C6a and C8 -> C8a to match [bodge function](https://github.com/cmu-delphi/covidcast-indicators/blob/6dba43d13ebe5a092249c9af23930d09e60badc3/facebook/delphiFacebook/R/responses.R#L474) and intended naming.
- In CMU Wave 11, manually rename E1 subquestions to match [bodge function](https://github.com/cmu-delphi/covidcast-indicators/blob/6dba43d13ebe5a092249c9af23930d09e60badc3/facebook/delphiFacebook/R/responses.R#L547).
- Standardize meaning of "none" and "NA" in the codebook. `NA` means that that particular field isn't valid for the item listed. For example, the `module` field isn't a survey item so `display_logic` can't be defined for it. For survey items with no display logic, `display_logic` is `none` because it could be defined for a survey item, it's just not in this case. `none` is the opposite of `all`.
- Fix how matrix items are created when joining the codebook onto the diff. The diff contains one observation per matrix item. The codebook contains one observation per matrix subquestion (so 3+ per matrix question).

### Changelog
- `qsf-tools/generate-changelog.R`
- `qsf-tools/generate-codebook.R`
- `qsf-tools/qsf-differ.R`

### DIFF
--- a/facebook/qsf-tools/generate-changelog.R
+++ b/facebook/qsf-tools/generate-changelog.R
@@ -189,19 +189,15 @@ generate_changelog <- function(path_to_codebook,
 }
 
 
-# args <- commandArgs(TRUE)
-# 
-# if (length(args) != 4) {
-#   stop("Usage: Rscript generate-changelog.R [UMD/CMU] path/to/codebook path/to/annotated/diff path/to/changelog")
-# }
-# 
-# survey_version <- args[1]
-# path_to_codebook <- args[2]
-# path_to_diff <- args[3]
-# path_to_changelog <- args[4]
-path_to_codebook <- "codebook.csv"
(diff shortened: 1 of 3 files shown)


nmdefries:
@leonardzh Please take a look at this. Let me know if this version works for you and if you have anything you want to change.

nmdefries:
@capnrefsmmat reminder. This is working for Zheng and me.

capnrefsmmat (approved):

nmdefries:
@krivard this is ready to merge.

PR #1582, opened 2022-04-15
[CTIS qsf tools] Let changelog use rationales from old changelog, and read in a directory of diffs

nmdefries:
### Description
- If path to previous version of changelog is provided, then the rationales from that version are used as mu;ch as possible in the new version. This lets us reuse old rationales when we make changes to the diffs or codebook.
- Warn if any items are missing rationales and print out a list of implicated item names.
- Allow changelog to use a single diff file or all diffs in a directory, to avoid a separate append step.
- Refactor `generate-changelog` for readability.

### Changelog
- `generate-changelog.R`

### DIFF
--- a/facebook/qsf-tools/generate-changelog.R
+++ b/facebook/qsf-tools/generate-changelog.R
@@ -23,6 +23,42 @@ generate_changelog <- function(path_to_codebook,
                                rename_map_file="item_rename_map.csv") {
   # Get the codebook. Contains details about each question (text, answer
   # choices, display logic) by wave.
+  codebook <- get_codebook(path_to_codebook)
+  
+  # Get the diffs + rationale. Contains info about which items changed between
+  # waves, plus a description of what changed and why.
+  annotated_diff <- get_diff(path_to_diff)
+  
+  if (!("notes" %in% names(annotated_diff))) {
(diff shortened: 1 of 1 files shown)


nmdefries:
@leonardzh This changes how the changelog is generated. The major changes are:

1) You can now pass the script a path to a directory of diffs instead of a path to a single diff. The script will load all CSVs in the directory, and it will fail (on purpose) if they aren't in the same format (e.g. different columns). You can still pass a path to a single diff file if you want.

2) You can pass the script the path to an old version of the changelog (optional). If a path is provided, the script will replace any existing `notes` field with `notes` from the old version of the changelog. Rationales are matched by new and old wave, variable name, and change type. If the underlying diffs have changed, it is possible that some items from the old changelog are not used, and that some items in the new changelog don't have rationales. The script will warn about any items that are missing rationales.

I haven't fully tested this yet, but please try it out when you have a chance and let me know if you encounter any issues.

capnrefsmmat:
Is this meant to be merged into `ndefries/qsf-tools-small-fixes` instead of main?

nmdefries:
Yes, it is intentional.

capnrefsmmat (commented):
Looks good, with one tiny typo fix

PR #1578, opened 2022-04-14
DSEW-CPR: Extend files to be processed for interpolation

krivard:
### Description
when filtering the file listing for what to process, extend the final list by (INTERP_LENGTH-1) days backward from each desired file, so that we have something to interpolate with.

Linting errors are extant in #1555 

### Changelog
Itemize code/test/documentation changes and files added/removed.
- constants: add INTERP_LENGTH
- pull: add extend_listing_for_interp
- test: check single range, ranges that don't overlap, and ranges that do overlap.

### Fixes 
- Fixes TODO in #1555

PR #1583, opened 2022-04-18
[dsew-cpr] Fix UnboundLocalError

krivard:
### Description
The change from #1578 included a call to a local variable that (turns out) does not get initialized when `reports` is set to `all` 😭 

initialize it, but also don't do the fileset extension if we already know we've got all the files.

### Changelog
- pull.py:fetch_listing() - initialize `keep` and add guard around `extend_listing_for_interp`

### Fixes 
- Fixes failing staging tests


dshemetov (approved):
lgtm

PR #1584, opened 2022-04-19
Add archive params to prod dsew-cpr configuration

krivard:
### Description
Archive-differ is required for interpolation. Destination prefix is already set up with output through 2022-04-18. Once this is merged, we'll do a release; once the release is merged; interpolation will be activated in prod.

### Changelog
- Production params for dsew-cpr

### Fixes 
- Release mechanism for interpolation

### DIFF
--- a/ansible/templates/dsew_community_profile-params-prod.json.j2
+++ b/ansible/templates/dsew_community_profile-params-prod.json.j2
@@ -39,5 +39,14 @@
         "booster_doses_admin_7dav"
       ]
     }
+  },
+  "archive": {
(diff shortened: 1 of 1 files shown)


krivard:
whoops, timezones. looping in nat as a sub

dshemetov:
- Going to assume that the credentials {{ key }} piece somehow pulls from a local env file 
- Assuming the bucket name exists (not sure how I can verify that, I only used our S3 a while ago)
- Cache dir seems correct
- Indicator prefix looks correct

Issue #1588, opened 2022-04-20
create_export_csv should have an option to sort by geo before writing


dshemetov:
Reading csv diffs requires the geos to be sorted. This is something every indicator should need, so ideally we put this in here. `create_export_csv(..., sort_index = False)` for now and think about making the default True in a future refactor.

PR #1590, opened 2022-04-20
Don't drop positivity signal when interpolating

### LINKED ISSUE #1585: Interpolation should skip test positivity, not drop it
A side effect of line 620 that I didn't notice at review time is that it removes all the test positivity dfs from the possible output. It should probably add the df to interpolate_df unchanged before `continue`ing.

https://github.com/cmu-delphi/covidcast-indicators/blob/1e5e5cf8dc6e23d9a15bc0390b4b38a1a96e8354/dsew_community_profile/delphi_dsew_community_profile/pull.py#L614-L620

This is a fail-safe bug for the moment, since the test positivity signals haven't been released yet.

dshemetov:
### Description
Fixes positivity signal drop.

### Changelog
- Add the positivity dataframe before moving on
- Sort by geo for every signal, before outputting (for easier diffs). Not that much slower. Eventually, I'd like to refactor this with #1588 

### Fixes 
- Fixes #1585


krivard (approved):
👍

PR #1591, opened 2022-04-20
Add sort option to create_export_csv

### LINKED ISSUE #1588: create_export_csv should have an option to sort by geo before writing

dshemetov:
### Description
Adds the option to sort by geo before writing a csv.

### Changelog
- Test included.

### Fixes 
- Fixes #1588

### DIFF
--- a/_delphi_utils_python/tests/test_export.py
+++ b/_delphi_utils_python/tests/test_export.py
@@ -30,7 +30,9 @@ def _non_ignored_files_set(directory):
     return out
 
 def _set_df_dtypes(df: pd.DataFrame, dtypes: Dict[str, Any]) -> pd.DataFrame:
-    assert all(isinstance(e, type) for e in dtypes.values()), "Values must be types."
+    assert all(isinstance(e, type) or isinstance(e, str) for e in dtypes.values()), (
+        "Values must be types or Pandas string aliases for types."
+    )
 
     df = df.copy()
     for k, v in dtypes.items():
--- a/_delphi_utils_python/tests/test_export.py
+++ b/_delphi_utils_python/tests/test_export.py
@@ -30,6 +30,8 @@ def _non_ignored_files_set(directory):
     return out
 
 def _set_df_dtypes(df: pd.DataFrame, dtypes: Dict[str, Any]) -> pd.DataFrame:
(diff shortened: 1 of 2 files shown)


_delphi_utils_python/tests/test_export.py
```diff
@@ -372,3 +381,40 @@ def test_export_df_with_contradictory_missingness(self, mock_logger):
...
+        })
+        create_export_csv(
+            unsorted_df,
+            export_dir=self.TEST_DIR,
+            geo_res="county",
+            sensor="test"
+        )
+        assert_frame_equal(
+            _set_df_dtypes(pd.read_csv(join(self.TEST_DIR, "20200215_county_test.csv")), dtypes={"geo_id": str}),
+            unsorted_df.query("timestamp == '2020-02-15'").drop(columns="timestamp").reset_index(drop=True)
```
krivard, line 406: The expected df only has two rows, so it may actually be clearer here to write it out instead of computing it

_delphi_utils_python/tests/test_export.py
```diff
@@ -372,3 +381,40 @@ def test_export_df_with_contradictory_missingness(self, mock_logger):
...
+        create_export_csv(
+            unsorted_df,
+            export_dir=self.TEST_DIR,
+            geo_res="county",
+            sensor="test",
+            sort_geos=True
+        )
+        assert_frame_equal(
+            _set_df_dtypes(pd.read_csv(join(self.TEST_DIR, "20200215_county_test.csv")), dtypes={"geo_id": str}),
+            unsorted_df.query("timestamp == '2020-02-15'").sort_values(by="geo_id").drop(columns="timestamp").reset_index(drop=True)
```
krivard, line 419: generally bad practice to use the same code to produce expecteds as is used in the code under test, just in case something is squirrely. since the expected df is only two rows, may as well write it out.

krivard (changes requested):
Not urgent, but it's better not to compute the expecteds here

krivard (approved):
👍

Issue #1593, opened 2022-04-21
DSEW-CPR hospital admissions submits megacounties with NULL values


krivard:
DSEW-CPR shouldn't be submitting null values, however, it has apparently been doing so since its initial release. There are however some patterns. 

The "53" below accounts for 52 megacounties and 1 MSA in Puerto Rico. My understanding is that DSEW-CPR should typically cover all counties, so that megacounties are not used.

We should prevent this indicator from publishing megacounties, if possible, or drop NULL-valued entries before exporting to CSV, if not.

```
MariaDB [epidata]> select issue,count(1) from covidcast where source="dsew-cpr" and `signal`="confirmed_admissions_covid_1d_7dav" and time_type="day" and value is NULL group by issue;
+----------+----------+
| issue    | count(1) |
+----------+----------+
| 20220202 |   215027 |
| 20220203 |      795 |
| 20220204 |      795 |
| 20220207 |      795 |
| 20220208 |      795 |
| 20220209 |      795 |
| 20220210 |      795 |
| 20220211 |      795 |
| 20220214 |      795 |
| 20220215 |      795 |
| 20220216 |      795 |
| 20220217 |      795 |
| 20220218 |      795 |
| 20220222 |      795 |
| 20220223 |      795 |
| 20220303 |      795 |
| 20220304 |       53 |
| 20220307 |       53 |
| 20220308 |       53 |
| 20220309 |       53 |
| 20220310 |       53 |
| 20220311 |       53 |
| 20220314 |       53 |
| 20220315 |       53 |
| 20220316 |       53 |
| 20220317 |       53 |
| 20220318 |       53 |
| 20220321 |       53 |
| 20220322 |       53 |
| 20220323 |       53 |
| 20220324 |       53 |
| 20220325 |       53 |
| 20220328 |       53 |
| 20220329 |       53 |
| 20220330 |       53 |
| 20220331 |       53 |
| 20220401 |       53 |
| 20220404 |       53 |
| 20220405 |       53 |
| 20220406 |       53 |
| 20220407 |       53 |
| 20220408 |       53 |
| 20220411 |       53 |
| 20220412 |       53 |
| 20220413 |       53 |
| 20220414 |       53 |
| 20220415 |       53 |
| 20220418 |       53 |
| 20220419 |   528145 |
| 20220420 |       53 |
+----------+----------+
50 rows in set (5.84 sec)

[...]

PR #1594, opened 2022-04-21
Cast CPR value columns to float before interpolating

nmdefries:
### Description
Explicitly cast `val`, `se`, and `sample_size` fields to `float` right before performing interpolation. Add a test to check if interpolation works correctly when the input df has value columns of type `object`.

Drop obs where `val` is missing.

Bonus: update `assert_frame_equal` import to silence deprecation warning.

### Changelog
- `pull.py`
- `test_pull.py`

### Fixes 
In some cases, interpolation doesn't work correctly, reporting all `NA`s for a geo type-signal-date combination that should have been interpolatable.

[Combining dfs across reports](https://github.com/cmu-delphi/covidcast-indicators/blob/939501ba577fc8990a3a5e7cab2a52e90eeb7088/dsew_community_profile/delphi_dsew_community_profile/pull.py#L550) sometimes implicitly converts an int or float `val` column to an `object` type.

This happens when the `val` field is different types in different subdfs (the per-signal per-report dfs).

When a date range includes at least one report for which a signal is not available, we produce an [empty dataframe with the correct column names but all of `object` type](https://github.com/cmu-delphi/covidcast-indicators/blob/939501ba577fc8990a3a5e7cab2a52e90eeb7088/dsew_community_profile/delphi_dsew_community_profile/pull.py#L362-L365), the default type. On the other hand, when a signal is available in a given report, we read in that data with `val` as a numeric.

Since a standard df column can't hold multiple types, when we combine the empty subdfs with the correctly-typed subdfs, the resulting columns are auto-converted to `object` types, which can hold mixed types. (`object` fields are also used to hold strings.)

The `interpolate` function [doesn't perform linear interpolation for `object`-typed columns](https://note.nkmk.me/en/python-pandas-interpolate/), because it thinks those columns are either strings or mixed types that can't be manipulated mathematically in a uniform fashion. (Although in our case, we have an `object` field that actually only contains a single type of element.) As a result, `interpolate` doesn't overwrite the `NaN`s that `reindex` inserts.

So limited-scope, recent time periods (e.g. 2022-04-01 through 2022-04-07) interpolate correctly, since all geo types have all signals available in every report, but older/longer time periods won't interpolate correctly. Specifically, they won't work for any signal not included in one or more reports and _included_ in one or more reports.

Incidentally also closes https://github.com/cmu-delphi/covidcast-indicators/issues/1593.

### DIFF
--- a/dsew_community_profile/delphi_dsew_community_profile/pull.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/pull.py
@@ -650,6 +650,7 @@ def interpolate_missing_values(dfs: DataDict) -> DataDict:
                 reindexed_group_df["publish_date"] = reindexed_group_df["publish_date"].fillna(
                     method="bfill"
                 )
+            reindexed_group_df = reindexed_group_df[~reindexed_group_df.val.isna()]
             geo_dfs.append(reindexed_group_df)
         interpolate_df[key] = (
             pd.concat(geo_dfs).reset_index().rename(columns={"index": "timestamp"})
--- a/dsew_community_profile/delphi_dsew_community_profile/pull.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/pull.py
@@ -630,14 +630,12 @@ def interpolate_missing_values(dfs: DataDict) -> DataDict:
                     reindexed_group_df["val"]
                     .astype(float)
                     .interpolate(method="linear", limit_area="inside")
-                    .astype(float)
                 )
             if "se" in reindexed_group_df.columns:
                 reindexed_group_df["se"] = (
                     reindexed_group_df["se"]
                     .astype(float)
                     .interpolate(method="linear", limit_area="inside")
-                    .astype(float)
                 )
             if (
                 "sample_size" in reindexed_group_df.columns
@@ -647,7 +645,6 @@ def interpolate_missing_values(dfs: DataDict) -> DataDict:
                     reindexed_group_df["sample_size"]
                     .astype(float)
                     .interpolate(method="linear", limit_area="inside")
-                    .astype(float)
                 )
             if "publish_date" in reindexed_group_df.columns:
                 reindexed_group_df["publish_date"] = reindexed_group_df["publish_date"].fillna(
--- a/dsew_community_profile/delphi_dsew_community_profile/pull.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/pull.py
@@ -628,12 +628,14 @@ def interpolate_missing_values(dfs: DataDict) -> DataDict:
             if "val" in reindexed_group_df.columns and not reindexed_group_df["val"].isna().all():
                 reindexed_group_df["val"] = (
                     reindexed_group_df["val"]
+                    .astype(float)
                     .interpolate(method="linear", limit_area="inside")
                     .astype(float)
                 )
             if "se" in reindexed_group_df.columns:
                 reindexed_group_df["se"] = (
                     reindexed_group_df["se"]
+                    .astype(float)
                     .interpolate(method="linear", limit_area="inside")
                     .astype(float)
                 )
@@ -643,6 +645,7 @@ def interpolate_missing_values(dfs: DataDict) -> DataDict:
             ):
                 reindexed_group_df["sample_size"] = (
                     reindexed_group_df["sample_size"]
+                    .astype(float)
                     .interpolate(method="linear", limit_area="inside")
                     .astype(float)
                 )
--- a/dsew_community_profile/tests/test_pull.py
+++ b/dsew_community_profile/tests/test_pull.py
@@ -525,6 +525,26 @@ class TestPull:
         expected_dfs = pd.concat([sig1, sig2, sig3, sig4])
         _assert_frame_equal(interpolated_dfs1[("src", "sig", False)], expected_dfs, index_cols=["geo_id", "timestamp"])
 
+    def test_interpolation_object_type(self):
+        DTYPES = {"geo_id": str, "timestamp": "datetime64[ns]", "val": float, "se": float, "sample_size": float, "publish_date": "datetime64[ns]"}
+        line = lambda x: 3 * x + 5
+
+        sig1 = _set_df_dtypes(pd.DataFrame({
+            "geo_id": "1",
+            "timestamp": pd.date_range("2022-01-01", "2022-01-10"),
+            "val": [line(i) for i in range(2, 12)],
+            "se": [line(i) for i in range(1, 11)],
+            "sample_size": [line(i) for i in range(0, 10)],
+            "publish_date": pd.to_datetime("2022-01-10")
+        }), dtypes=DTYPES)
+        # A linear signal missing two days which should be filled exactly by the linear interpolation.
+        missing_sig1 = sig1[(sig1.timestamp <= "2022-01-05") | (sig1.timestamp >= "2022-01-08")]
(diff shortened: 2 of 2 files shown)


dsew_community_profile/tests/test_pull.py
```diff
@@ -525,6 +525,26 @@ def test_interpolation(self):
...
+            "geo_id": "1",
+            "timestamp": pd.date_range("2022-01-01", "2022-01-10"),
+            "val": [line(i) for i in range(2, 12)],
+            "se": [line(i) for i in range(1, 11)],
+            "sample_size": [line(i) for i in range(0, 10)],
+            "publish_date": pd.to_datetime("2022-01-10")
+        }), dtypes=DTYPES)
+        # A linear signal missing two days which should be filled exactly by the linear interpolation.
+        missing_sig1 = sig1[(sig1.timestamp <= "2022-01-05") | (sig1.timestamp >= "2022-01-08")]
+        missing_sig1 = _set_df_dtypes(missing_sig1, {key: object for key in DTYPES.keys()})
```
krivard, line 542:
```suggestion
        # set all columns to object type to simulate the miscast we sometimes see when combining dfs
        missing_sig1 = _set_df_dtypes(missing_sig1, {key: object for key in DTYPES.keys()})
```

krivard (approved):
Nicely done! Added a clarifying comment since my brain kept skipping that line and wondering what made this test different 🤷‍♀️

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -628,12 +628,14 @@ def interpolate_missing_values(dfs: DataDict) -> DataDict:
             if "val" in reindexed_group_df.columns and not reindexed_group_df["val"].isna().all():
                 reindexed_group_df["val"] = (
                     reindexed_group_df["val"]
+                    .astype(float)
                     .interpolate(method="linear", limit_area="inside")
                     .astype(float)
```
dshemetov, line 633: Wondering if this cast to float at the end is unneeded now.

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -628,12 +628,14 @@ def interpolate_missing_values(dfs: DataDict) -> DataDict:
...
                 )
             if "se" in reindexed_group_df.columns:
                 reindexed_group_df["se"] = (
                     reindexed_group_df["se"]
+                    .astype(float)
                     .interpolate(method="linear", limit_area="inside")
                     .astype(float)
```
dshemetov, line 640: Same here.

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -643,6 +645,7 @@ def interpolate_missing_values(dfs: DataDict) -> DataDict:
             ):
                 reindexed_group_df["sample_size"] = (
                     reindexed_group_df["sample_size"]
+                    .astype(float)
                     .interpolate(method="linear", limit_area="inside")
                     .astype(float)
```
dshemetov, line 650: And here.

dshemetov (approved):
Good find, with a clean solution! I have a small note on the float casts, but otherwise lgtm.

dshemetov:
Oh, one more question: what will the response of the indicator be to a signal with all nans? All nans? Should we do some nan-pruning after the interpolation? Something like
```
ret = ret[~ret.val.isna()]
```

This should reduce nans in cases like these (note that this changes the behavior of some previous tests)
```
import pandas as pd
import numpy as np

sig1 = _set_df_dtypes(
    pd.DataFrame({
        "geo_id": "1",
        "timestamp": pd.date_range("2022-01-01", "2022-01-10"),
        "val": [line(i) for i in range(2, 12)],
        "se": [line(i) for i in range(1, 11)],
        "sample_size": [line(i) for i in range(0, 10)],
        "publish_date": pd.to_datetime("2022-01-10"),
    }),
    dtypes=DTYPES,
)
# A linear signal missing two days which should be filled exactly by the linear interpolation.
missing_sig1 = sig1[(sig1.timestamp <= "2022-01-05") | (sig1.timestamp >= "2022-01-08")]

sig2 = _set_df_dtypes(
    pd.DataFrame({
        "geo_id": "1",
        "timestamp": pd.date_range("2022-01-01", "2022-01-10"),
        "val": None,
        "se": [line(i) for i in range(1, 11)],
        "sample_size": [line(i) for i in range(0, 10)],
        "publish_date": pd.to_datetime("2022-01-10"),
    }),
    dtypes=DTYPES,
)
# A linear signal missing everything, which should be dropped since it's all NAs.
missing_sig2 = sig2[(sig2.timestamp <= "2022-01-05") | (sig2.timestamp >= "2022-01-08")]

sig3 = _set_df_dtypes(
    pd.DataFrame({
        "geo_id": "3",
        "timestamp": pd.date_range("2022-01-02", "2022-01-09"),
        "val": [line(i) for i in range(3, 11)],
        "se": [line(i) for i in range(1, 9)],
        "sample_size": [line(i) for i in range(0, 8)],
        "publish_date": pd.to_datetime("2022-01-10"),
    }),
    dtypes=DTYPES,
)
# A signal shorter than sig1, so it should output a reduced range without NAs.
missing_sig3 = sig3[(sig3.timestamp <= "2022-01-05") | (sig3.timestamp >= "2022-01-08")]

missing_dfs = [missing_sig1, missing_sig2, missing_sig3, missing_sig4]
interpolated_dfs1 = interpolate_missing_values(
    {("src", "sig", False): pd.concat(missing_dfs)}
)
expected_df = pd.concat([sig1, sig3])
_assert_frame_equal(
    interpolated_dfs1[("src", "sig", False)],
    expected_dfs,
    index_cols=["geo_id", "timestamp"],
)
```

dsew_community_profile/delphi_dsew_community_profile/pull.py
nmdefries: Good catch. Removing the second float conversion doesn't impact test output or tests.

nmdefries:
> what will the response of the indicator be to a signal with all nans? All nans?

nmdefries:
Summary of Slack discussion with Dmitry:

> what will the response of the indicator be to a signal with all nans? All nans?

Yep. Right now, rows with `NaN` values are included in the exported CSVs. CSVs can include any number of rows with `NaN` values, so CSVs with fully-`NaN` values are technically possible, but should only happen when interpolation fails.

The [interpolation reindex procedure](https://github.com/cmu-delphi/covidcast-indicators/blob/5404e7d2b6f1fb917be2b38da508a69c91c6e081/dsew_community_profile/delphi_dsew_community_profile/pull.py#L626) does _not_ add new `NaN`s on dates before or after the current signal was originally available for. Each geo-signal combo uses its own index (so one geo with a longer date range won’t impose its index on another geo). 

So the only `NA`s we expect to see are directly from the reports. The reports have occasional "real" missing values for specific geos (especially smaller geos like counties and territories). Most of those get filled in by interpolation. Rarely, a `NaN` doesn't get interpolated because it is on the leading or lagging edge.

We have two options to handle these:

- leave them in there and use the `NA` coding functionality to annotate them
- drop all `NA`s

`NA`-coding gets a consistency vote. Dropping gets a simplicity vote and a “minimizing uselessness” vote. It's unclear what additional info someone would actually get from an `NA` being present with an annotation versus the value just being missing.

krivard:
Dropping NAs would also resolve #1593 (the ham-handed way, but better than not)

dshemetov (approved):
lgtm!

PR #1602, opened 2022-04-27
Produce CPR positivity in production

nmdefries:
### Description
Update CPR production params so that test positivity is output.

Also remove total tests signal from params; it is reported as the sample size for test positivity, not as a separate signal.

### Changelog
- Ansible template params
- Indicator local params

Note: before releasing this, we need to verify in staging that test positivity files are produced, but values are not interpolated. This looks correct in local testing.


dshemetov (approved):
lgtm

PR #1609, opened 2022-05-03
Only raise errors if reference data is empty after padding

qx-teo:
**DRAFT PULL REQUEST**: Still requires testing

### Description
Raise empty_reference_data error only if padding fails. If padding succeeds, raise warning instead.

### Changelog
- dynamic.py

### Fixes 
- GS outages: If a signal has an outage for longer than the reference period, then reference_data would be empty and padding would not be done. (Probably an oversight when writing pad_reference_df)

### DIFF
--- a/_delphi_utils_python/delphi_utils/validator/dynamic.py
+++ b/_delphi_utils_python/delphi_utils/validator/dynamic.py
@@ -273,8 +273,8 @@ class DynamicValidator:
             - report: ValidationReport; report where results are added
 
         Returns:
-            - False if recent_df is empty, else (recent_df, reference_api_df)
-            (after reference_api_df has been padded if necessary)
+            - False if recent_df is empty after padding, else (recent_df, reference_api_df)
+            (reference_api_df will be padded if necessary)
         """
         # recent_lookbehind: start from the check date and working backward in time,
         # how many days at a time do we want to check for anomalies?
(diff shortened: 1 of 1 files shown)


qx-teo:
@nmdefries The way this is written:
1: No need to pad - no errors
2: Need to pad, padding succeeds: 1 warning
3: need to pad, padding fails: 1 warning, 1 error

Should there be no warning for # 3?

nmdefries:
> Should there be no warning for # 3?

Yeah, it makes sense to only raise the error. I think the best way to do so is before padding, save the status of the reference data, like

```
pre_pad_empty_flag = reference_api_df.empty
```

and then add another clause to the later `if` statement with the error, like

```
        if reference_api_df.empty:
            report.add_raised_error(
                ValidationFailure("empty_reference_data",
                                  checking_date,
                                  geo_type,
                                  signal_type,
                                  "reference data is empty; comparative checks could not "
                                  "be performed"))
            return False
       elif pre_pad_empty_flag:
           report.add_raised_warning(
                ValidationFailure(...)
           )
```

_delphi_utils_python/delphi_utils/validator/dynamic.py
```diff
@@ -320,47 +320,71 @@ def create_dfs(self, geo_sig_df, api_df_or_error, checking_date, geo_type, signa
...
+    # `reference_start_date` is used in the call to `geo_sig_df.query()`
+    # below but pylint doesn't recognize that.
+    # pylint: disable=unused-argument
+    def pad_reference_api_df(self, reference_api_df, geo_sig_df,
+                             reference_start_date, reference_end_date):
         """Check if API data is missing, and supplement from test data.
 
         Arguments:
-            - reference_api_df: API data within lookbehind range
+            - reference_api_df: API data within lookbehind range with at least one observation
```
nmdefries, line 354:
```suggestion
            - reference_api_df: API data within lookbehind range
```

nmdefries (approved):
Small change to docstrings, otherwise this looks good.

nmdefries:
@qx-teo You should review this as well to check the changes I made.

qx-teo:
I looked through them, everything seems fine! The log output also seems to indicate that things are working as intended, so that's good

PR #1612, opened 2022-05-06
Drop CPR test positivity rows with missing `val` even if `sample_size` is not missing

nmdefries:
### Description
[Context in Slack](https://delphi-org.slack.com/archives/C013AH5N01E/p1651699160417189?thread_ts=1651600391.284619&cid=C013AH5N01E).

### Changelog
- `pull.py`


krivard (approved):
legit 👍

PR #1613, opened 2022-05-09
Use date portion of datetime when param not available

krivard:
### Description
wee bug we missed in #1610 -- everything to do with listings is expressed in dates, not datetimes, and you can't compare datetimes with dates.

### Changelog
- pull.py:fetch_listing(): if export_start_date param not available, use date portion only of default datetime 

### Fixes 
- Fixes failing run in staging


nmdefries (approved):
Looks good, works in a local run.

PR #1625, opened 2022-05-24
Fix up weekly weights Makefile logic

nmdefries:
### Description
Revert testing changes.

Fix syntax and logic issues discovered during testing.

### Changelog
- Makefile
    - `tidy_*` target: check if any CSV files exist in `receiving` before trying to copy. Exclude "done" files when making tarballs.
    - `weekly-weights-pipeline` target: make weekly weights "done" file in `tidy_weekly` so it gets copied out of the docker environment.
    - `$(WEEKLY_WEIGHTS)` target: fix lots of syntax issues mostly involving command continuation. Took out most comments because they disrupt command continuation. Set modify date on weekly weights input dir to "yesterday" so weekly weights target and thus weekly pipeline always run, including in same-day reruns. Change pipeline conditional logic to run if the correct weekly weights are available in the local dir and last week has no "done" file yet, instead of running only when new weights files are fetched -- this wasn't great if we needed to do a rerun.
- `binary.R`
- `count.R`
- `build-container-images.yml`


facebook/Makefile
```diff
@@ -70,10 +70,11 @@ tidy_%: receiving
 	rm -f $@/params.json
 	mkdir -p $@ $@/$(RECEIVING) $@/$(INDIVIDUAL) $@/$(INDIVIDUAL_RACEETH)
 	cp params.json $@/
-	mv $(RECEIVING)/*.csv $@/$(RECEIVING)
+	# Check for _any_ matching files using https://stackoverflow.com/a/6364244/14401472
+	if compgen -G "$(RECEIVING)/*.csv" > /dev/null; then mv $(RECEIVING)/*.csv $@/$(RECEIVING); fi
```
krivard, line 74: TIL `compgen`

krivard (approved):
👍

PR #1626, opened 2022-06-02
Basic Change Flags Code

Ananya-Joshi:
TODOs Left: 
- create and run unit tests. 
- update with more feedback.


Ananya-Joshi:
Discussion Items Left: 
Logger finalize
Input finalize

We likely need to deploy it first so that I can test the output logging system.

Ananya-Joshi:
Converted to draft so that we can change the structure of the code from being an 'indicator; to part of the utils package.

PR #1630, opened 2022-06-06
[CTIS QSF tools] Generate changelog from diffing codebook

nmdefries:
### Description
Currently, the survey changelog is generated by joining descriptive fields from the codebook onto QSF diffs. Since the codebook is meant to be a parsed version of the QSFS that matches the microdata and the diffs are meant to report raw fields from the QSFs, a lot of processing is required to make the two compatible (e.g. some survey items are renamed in specific waves). This means that the code to create the survey changelog is complex, making it both hard to maintain and likely to contain bugs.

This PR generates the changelog instead by joining the codebook with itself, and diffing pairs of columns (old and new version of a characteristic) for each observation.

Fix a few issues with the codebook, mostly related to display logic edge cases.

### Changelog
- `generate-codebook.R`
- `generate-changelog.R`


capnrefsmmat:
I trust you've run this and compared the output, e.g. diffing the old changelog against the new one?

nmdefries:
Yep.

Differences are minimal -- a few buggy lines in the changelog have been removed (e.g. the old version showed display logic as having changed if just the QID of the conditional question changed, even if all text, logic, and answer choices were the same). There are a few additional obs (O(10)) included since this version diffs a few extra fields. And a few changelog lines have changed type. "Matrix subquestion changed" specifically has been broken down into more specific categories.

nmdefries:

PR #1633, opened 2022-06-09
Set weekly weights params `end_date` to `YESTERDAY`

nmdefries:
### Description
Microdata output filenames include a `recordedby` date based on the last requested output date (`params$end_date`, set to `YESTERDAY` for the daily pipeline). The most recently generated microdata files should have the most recent `recordedby` date so users and our monthly rollup script know which files are current.

The weekly weights pipeline requests output for `LAST_SUNDAY` through `LAST_SATURDAY`, so the `recordedby` date is `LAST_SATURDAY`. This means that if it's too far into the next week, weekly microdata output files won't appear as the most current version of a given date.

For example,  it it's Wednesday, the daily pipeline will generate microdata for Fri, Sat, Sun, Mon recorded by Tues. The weekly pipeline would generate microdata for last Sun-Sat recorded by Sat, so the daily files for Fri and Sat would appear to be more current.

Fix this by asking the weekly weights pipeline to generate output through `YESTERDAY` as well, so it's output will always appear as up-to-date as that from the daily pipeline.

### Changelog
- `Makefile` `params.weekly-weights.json` target

### DIFF
--- a/facebook/Makefile
+++ b/facebook/Makefile
@@ -135,7 +135,7 @@ params.weekly-weights.json: $(TODAY)
 	$(PYTHON) -m delphi_utils set \
 		debug false \
 		produce_individual_raceeth true \
-		end_date $(LAST_SATURDAY) \
+		end_date $(YESTERDAY) \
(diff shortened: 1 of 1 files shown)


krivard:
Legit. Want me to merge&release today or hold off until Monday?

nmdefries:
Today would be good

PR #1634, opened 2022-06-09
Update test_run.py

### LINKED ISSUE #1451: Switch Safegraph indicator to use the Places API

neul3:
### Description
Changed two file path names 

### Fixes 
- Fixes #1451

### DIFF
--- a/safegraph_patterns/tests/test_run.py
+++ b/safegraph_patterns/tests/test_run.py
@@ -10,7 +10,7 @@ from delphi_safegraph_patterns.run import (run_module, METRICS,
 class TestRun:
     PARAMS = {
         "common": {
-            "export_dir": "./test_data/receiving"
+            "export_dir": "./receiving"
         },
         "indicator": {
             "static_file_dir": "../static",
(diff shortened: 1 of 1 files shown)


krivard:
This was a reasonable fallback procedure since you didn't have push access to this repo, but it does mean that CI didn't run for this PR since it doesn't work on forks. I'll fix the access problem and merge the PR, and we'll use #1454 to see the CI results.

PR #1635, opened 2022-06-10
[CTIS] New indicators to round out contingency tables before archival

nmdefries:
### Description
This only includes non-demographic indicators from the [new indicators doc](https://docs.google.com/spreadsheets/d/149mTWcacPL76VW6xBbn3vyHvDaiAmoDvercse6Q3GSI/edit#gid=0).

Support percentile fields for numeric metrics.

### Changelog
- `contingency_indicators.R`
- `contingency_variables.R`
- `contingency_calculate.R`
- `variables.R`


facebook/delphiFacebook/R/contingency_calculate.R
```diff
@@ -104,3 +104,53 @@ post_convert_count_to_pct <- function(df) {
...
+      represented = NA_real_
+    ))
+  }
+  
+  design <- svydesign(id = ~1, weight = ~weight, data = data.frame(response, weight))
+  return(list(
+    val = as.data.frame(svymean(~response, na.rm = TRUE, design = design))[,"mean"],
+    se = NA_real_,
+    sd = as.data.frame(sqrt(svyvar(~response, na.rm = TRUE, design = design)))[,"variance"],
+    p25 = as.data.frame(oldsvyquantile(~response, na.rm = TRUE, design = design, quantiles = 0.25))[,"0.25"],
```
capnrefsmmat, line 149: Hmm, why are we using `oldsvyquantile` instead of `svyquantile`? I don't know the details of what changed

capnrefsmmat: Incidentally you might like [srvyr](http://gdfe.co/srvyr/index.html), which provides a tidy interface to the survey package

facebook/delphiFacebook/R/contingency_variables.R
```diff
@@ -271,12 +289,143 @@ code_addl_vaccines <- function(input_data, wave) {
...
+    input_data$t_unusual_symptom_tested_positive <- NA
+  }
+
+  if ("B6" %in% names(input_data)) {
+    input_data$t_unusual_symptom_hospital <- input_data$B6 == 1
+  } else {
+    input_data$t_unusual_symptom_hospital <- NA
+  }
+
+  if ("B7" %in% names(input_data)) {
```
capnrefsmmat, line 327: Given the issue with B7 I raised on Slack, maybe we need to throw out Wave 10 or add some logic to correct the aggregates of it; worth raising with Adrianne maybe?

facebook/delphiFacebook/R/contingency_calculate.R
nmdefries: `svyquantile` doesn't work for older versions of R, so Esther switched her code to use the `old` version of the function. [Context](https://app.asana.com/0/0/1202106541327630/1202179687386809/f). Looks like the old one is slower and less flexible.

The approach I implemented here is just a simplified version of [Esther's code](https://app.asana.com/0/0/1202106541327630/1202143955278131/f), since we already have sample size filters, e.g., elsewhere.

facebook/delphiFacebook/R/contingency_variables.R
nmdefries: Told Adrianne about the issue.

What corrective logic are you thinking of? If B7 is being shown to only those who didn't respond at all to B2, then even if we filter it using other questions, we're not going to be able to extract any responses we're interested in (where someone answered both B2 and B2c with symptoms).

It seems to me that the only option is to drop all Wave 10 responses to B7.

nmdefries: Adrianne says dropping Wave 10 sounds fine.

capnrefsmmat (approved):

nmdefries:

PR #1647, opened 2022-06-23
Add county info to contingency tables

nmdefries:
### Description
Join county and state metadata on using existing files in `static` dir to include regional (state) and county identifiers in county contingency tables.

### Changelog
- `contingency_write.R`


facebook/delphiFacebook/R/contingency_write.R
```diff
@@ -109,15 +107,62 @@ add_geo_vars <- function(data, params, geo_type) {
         .data$county_fips
       )
   } else if (geo_type == "county") {
-    warning("county metadata not supported")
+    counties <- read_csv(
+      file.path(params$static_dir, "02_20_uszips.csv"),
+      col_types = cols(.default = "c")
+    ) %>% 
+      mutate(
+        fips = stri_pad(.data$fips, 5, pad="0")
```
capnrefsmmat, line 115: I usually use `sprintf("%05d", .data$fips)`, but I guess you've already imported `stri` so this is fine (and probably easier to understand)

facebook/delphiFacebook/R/contingency_write.R
```diff
@@ -109,15 +107,62 @@ add_geo_vars <- function(data, params, geo_type) {
...
+        state = .data$state_id,
+        county = .data$county_name,
+        county_fips = .data$county_fips
+      ) %>% 
+      select(
+        .data$region,
+        .data$state,
+        .data$county,
+        .data$county_fips
+      )
```
capnrefsmmat, line 136: `select` can rename columns while selecting (e.g. `select(foo = bar, baz = spam)`), so is it necessary to have two steps here?

capnrefsmmat (commented):
I think this is good, just one possible simplification

facebook/delphiFacebook/R/contingency_write.R
nmdefries: Ah, didn't know that, I'll combine the steps.

capnrefsmmat (approved):

PR #1650, opened 2022-06-28
Script to amend monthly rollup microdata

nmdefries:
### Description
Add state field based on county FIPS. Change name of `wave` field to
`version`.

### Changelog
- Add script `microdata_add_state_col__rename_wave.R`

### DIFF
--- a/facebook/amend_monthly_microdata.R
+++ b/facebook/amend_monthly_microdata.R
@@ -13,6 +13,7 @@ suppressPackageStartupMessages({
   library(readr)
   library(rlang)
   library(stringi)
+  library(delphiFacebook)
 })
 
 amend_microdata <- function(input_dir, output_dir, static_dir, pattern = ".*[.]csv[.]gz$") {
@@ -40,6 +41,8 @@ amend_microdata <- function(input_dir, output_dir, static_dir, pattern = ".*[.]c
 
     # Add state column based on county FIPS code.
     data <- left_join(data, state_county_map, by="fips")
+
+    assert(is.na(data$fips) == is.na(data$state))
 
     # Save file under original name but in output directory.
     message("writing data for ", fname)
--- a/facebook/amend_monthly_microdata.R
+++ b/facebook/amend_monthly_microdata.R
@@ -44,6 +44,9 @@ amend_microdata <- function(input_dir, output_dir, static_dir, pattern = ".*[.]c
 
     assert(is.na(data$fips) == is.na(data$state))
 
+    # Drop any territories.
+    data <- filter(data, !(state %in% c("AS", "GU", "PR", "VI", "MP")))
+
     # Save file under original name but in output directory.
     message("writing data for ", fname)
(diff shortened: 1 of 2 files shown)


capnrefsmmat (commented):
I think this will be fine, just slightly more complicated than it has to be, since the FIPS conversion can be easier.

Not sure if it's worth changing the conversion if it already works. Could you maybe add an assertion after line 42 that checks that all rows with FIPS codes get a non-NA state? That should definitely happen, but if there's something weird about our mapping files (like with territories), we could get issues, and we don't want to have missingness because of that.

facebook/microdata_add_state_col__rename_wave.R
```diff
@@ -0,0 +1,69 @@
...
+    # stop readr from thinking commas = thousand separators,
+    # and from inferring column types incorrectly
+    message("reading data in")
+    data <- read_csv(file.path(input_dir, fname), locale = locale(grouping_mark = ""),
+             col_types = cols(
+               .default = col_character())) %>%
+      rename(version = .data$wave)
+
+    # Add state column based on county FIPS code.
+    data <- left_join(data, state_county_map, by="fips")
```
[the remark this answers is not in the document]
nmdefries: :facepalm: and I thought this *was* fairly straightforward

facebook/amend_monthly_microdata.R
```diff
@@ -47,12 +53,43 @@ amend_microdata <- function(input_dir, output_dir, static_dir, pattern = ".*[.]c
...
+
+  # clean the ZIP data
+  data$zip5 <- stri_replace_all(data$zip5, "", regex = " *")
+  data$zip5 <- stri_replace(data$zip5, "", regex ="-.*")
+
+  # some people enter 9-digit ZIPs, which could make them easily identifiable in
+  # the individual output files. rather than truncating to 5 digits -- which may
+  # turn nonsense entered by some respondents into a valid ZIP5 -- we simply
+  # replace these ZIPs with NA.
+  data$zip5 <- ifelse(nchar(data$zip5) > 5, NA_character_,
```
nmdefries, line 77: The logic for these two functions is borrowed from `delphiFacebook/R/responses.R`, but I'm wondering if this line is supposed to be handling A3. We drop `zip5` anyway, so nulling it out here seems unnecessary. @capnrefsmmat

capnrefsmmat: `filter_complete_responses` does the other half of the work: 

https://github.com/cmu-delphi/covidcast-indicators/blob/3590e64ea10aade257030b6f8f0815f9ccc3f30b/facebook/delphiFacebook/R/responses.R#L772-L777

We join with the ZIP metadata based on zip5, then blank out A3 based on the results (not zip5). Don't ask me why we need to use two different columns; undoubtedly the code did that in early 2020 and the logic just got ported over to this version

capnrefsmmat (approved):

PR #1651, opened 2022-06-30
Use underscore instead of dash for variable names

korlaxxalrok:
### Description
Fix:

Update a variable name to use an underscore instead of a dash.

### Changelog
- Updates multiple occurences of `template-secrets` to `template_secrets`


krivard (approved):
👍

PR #1657, opened 2022-07-15
Amend microdata improve zips and FIPS handling

nmdefries:
### Description
A couple fixes to the "amend microdata" script to deal with geo edge cases.

- Local zips info file was missing two FIPS codes. Switch to `covidcast` to generate state column instead.
- Blank invalid (low population) zip codes instead of blanking everything _but_ valid zip codes. Makes later assertion easier and preserves nonsense/invalid A3 values that have `NA` zip codes.
- Filter out territories by zip code in addition to filter on `state` in case state field is not available.

### Changelog
- `amend_monthly_microdata.R`


capnrefsmmat (approved):

nmdefries:

PR #1666, opened 2022-07-28
Remove explicit check for unique days = n_days

neul3:
### Description
Removing explicit check to let validator handle missing dates 

### Fixes 
- USA-facts will start working again


krivard:
Looks like you have failing tests -- largely the one testing the behavior this PR removes though, so it's okay to remove that test as well.

krivard (approved):
👍 
Noting for our records: before making this change, we verified that 7-day averages will compute correctly for weeks with missing days. The Validation and SirCAL systems will take care of alerting for gaps.

PR #1671, opened 2022-08-01
Add missing claims_hosp config

korlaxxalrok:
### Description
Fix:

Add missing template variables

### Changelog
- Add var ref
- Add var


jingjtang (approved):
LGTM

PR #1675, opened 2022-08-04
Backfill/claims_hosp

jingjtang:
### Description
Generate and store county level backfill file locally (On MIDAS) for `claims_hosp`. The newly added function will check the input file daily and store backfill file on a daily basis. Every 4 weeks, we will check the daily files and merge them into a combined file.

### Changelog
- `backfill.py`, `run.py` modify the current code, to save the county level intermediate file to `backfill_dir` before generating signals if there are new input raw files added by DV.
- `setup.py` add pyarrow as a necessary package so as to save the backfill file in the parquet format. 
- `tests/test_backfill.py` add the unit tests for backfill related helper functions.
- `params.json.template` add `backfill_dir`, `backfill_merge_day`


claims_hosp/delphi_claims_hosp/backfill.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    backfilldata = pd.read_csv(
+        claims_filepath,
+        usecols=Config.CLAIMS_DTYPES.keys(),
+        dtype=Config.CLAIMS_DTYPES,
+        parse_dates=[Config.CLAIMS_DATE_COL],
+    )
+
+    backfilldata.rename({"ServiceDate": "time_value",
+                         "PatCountyFIPS": "fips",
```
nmdefries, line 38: Same thing about exporting state data.

nmdefries (approved):

claims_hosp/delphi_claims_hosp/backfill.py
```diff
@@ -0,0 +1,116 @@
...
+        number to be 28, but we use a looser criteria from practical
+        considerations
+    """
+    new_files = glob.glob(backfill_dir + "/claims_hosp_as_of_*")
+
+    def get_date(file_link):
+        # Keep the function here consistent with the backfill path in
+        # function `store_backfill_file`
+        fn = file_link.split("/")[-1].split(".parquet")[0].split("_")[-1]
+        return datetime.strptime(fn, "%Y%m%d")
```
krivard, line 87: Will this same transformation need to be done for the backfill components of all pipelines? If so, should we put it in delphi_utils instead, so updates only ever need to be made in one place?

claims_hosp/tests/test_backfill.py
```diff
@@ -0,0 +1,83 @@
...
+        assert set(selected_columns) == set(backfill_df.columns)  
+        
+        os.remove(backfill_dir + "/" + fn)
+        assert fn not in os.listdir(backfill_dir)
+        
+    def test_merge_backfill_file(self):
+        
+        today = datetime.today()
+        
+        new_files = glob.glob(backfill_dir + "/claims_hosp*.parquet")
```
krivard, line 50: These files are pretty large -- 3MB each. Do they need to be that large in order to test this behavior, or would smaller files work?

claims_hosp/delphi_claims_hosp/backfill.py
jingjtang: Definitely. But it's not a very big deal and currently only needed for these pipelines.

claims_hosp/tests/test_backfill.py
jingjtang: We can use smaller test files instead.

claims_hosp/delphi_claims_hosp/backfill.py
krivard: That's okay for the initial release, but [I've written it up to refactor at some future date](https://github.com/cmu-delphi/covidcast-indicators/issues/1692). It should make a good onboarding task for new members.

nmdefries:
@jingjtang Can you resolve the conflict with `main`? Looks like Brian changed the `receiving_dir` in tandem; I'm not sure which is more up to date.

Added the daily file format changes.

jingjtang:
> @jingjtang Can you resolve the conflict with `main`? Looks like Brian changed the `receiving_dir` in tandem; I'm not sure which is more up to date.
> 
> Added the daily file format changes.

Use the one that Brian uses, that one should be more up-to-date

krivard (approved):
👍

PR #1681, opened 2022-08-16
Switch failing validation to if not statement 

neul3:
### Description
Bug Fix for Validator


neul3:
Local tests are successful after removing `APIDataFetchError` from the imported errors.

krivard:
is there a reason to leave it in?

neul3:
I don't believe so.

krivard:
Then we should remove it 🙂

krivard (approved):
👍

Issue #1685, opened 2022-08-23
Incorrect values in `jhu-csse:confirmed_incidence_num`


dshemetov:
Working on JIT A/B tests https://github.com/cmu-delphi/delphi-epidata/pull/947, I found this.

**Actual Behavior:**

The incidence values in the county '02100' don't agree with their cumulative counterparts.

```py
from epidatpy.request import EpiRange, Epidata

manual_incidence = Epidata.covidcast("jhu-csse", "confirmed_cumulative_num", "day", "county", EpiRange(20200620, 20200630), "02100").df().value.diff()[1:]
api_incidence    = Epidata.covidcast("jhu-csse", "confirmed_incidence_num",  "day", "county", EpiRange(20200621, 20200630), "02100").df().value
manual_incidence.to_numpy()
# array([ 0.,  0.,  1.,  2.,  0.,  0.,  0.,  0.,  0., -3.])
api_incidence.to_numpy()
# array([ 0.,  0.,  1.,  0.,  0.,  0.,  0.,  0.,  0.,  0.])
```

**Expected behavior**

These should match. The dates line up.

**Context**

Looking at `confirmed_cumulative_num`, there's definitely a jump of 2 cases on 2020-06-24

```py
Epidata.covidcast("jhu-csse", "confirmed_cumulative_num", "day", "county", EpiRange(20200620, 20200630), "02100").df()

#       source                    signal geo_type geo_value time_type time_value      issue  lag  value stderr sample_size direction  missing_value  missing_stderr  missing_sample_size
# 0   jhu-csse  confirmed_cumulative_num   county     02100       day 2020-06-20 2020-06-21    1    2.0   None        None      None              0               0                    0
# 1   jhu-csse  confirmed_cumulative_num   county     02100       day 2020-06-21 2020-10-29  130    2.0   None        None      None              0               0                    0
# 2   jhu-csse  confirmed_cumulative_num   county     02100       day 2020-06-22 2020-06-23    1    2.0   None        None      None              0               0                    0
# 3   jhu-csse  confirmed_cumulative_num   county     02100       day 2020-06-23 2020-06-24    1    3.0   None        None      None              0               0                    0
# 4   jhu-csse  confirmed_cumulative_num   county     02100       day 2020-06-24 2021-04-01  281    5.0   None        None      None              0               0                    0
[...]

PR #1687, opened 2022-08-26
Minor update to `delphi_utils` tests

dshemetov:
### Description
While working on #1685, I thought there was an issue with county zero padding. Turns out I was wrong, but I figured we can add the tests I wrote, so others don't have the same doubts later.

### Changelog
* add a county with zero padding to test_archive
* adjust spacings for readability in test_archive
* remove manual temp_dir in test_export and use pytest tmp_path


dshemetov:
I'm thinking the county test is probably worth it.

nolangormley (approved):
LGTM!

PR #1704, opened 2022-10-10
Get backfill corrections running via `make` targets

nmdefries:
### Description
Get pipeline running when called with `make run-local`.

### Changelog
- ...


nmdefries:
@jingjtang This is still in progress but feel free to take an initial look/do a light review.

backfill_corrections/delphiBackfillCorrection/R/beta_prior_estimation.R
```diff
@@ -146,13 +146,18 @@ frac_adj <- function(train_data, test_data, prior_test_data,
                      train_models = TRUE,
                      make_predictions = TRUE) {
   train_data$value_target <- frac_adj_with_pseudo(train_data, NULL, 1, 100, "value_target_num", "value_target_denom")
-  train_data$value_7dav <- frac_adj_with_pseudo(train_data, NULL, 1, 100, "value_7dav_num", "value_7dav_denom")
-  prior_test_data$value_7dav <- frac_adj_with_pseudo(prior_test_data, NULL, 1, 100, "value_7dav_num", "value_7dav_denom")
-  
   train_data$log_value_target <- log(train_data$value_target)
+
+  test_data$value_target <- frac_adj_with_pseudo(test_data, NULL, 1, 100, "value_target_num", "value_target_denom")
+  test_data$log_value_target <- log(test_data$value_target)
```
nmdefries, line 152: @jingjtang I added these definitions to get the pipeline to run. Fraction test data didn't have any `value_target_*` fields defined so it would fail [during evaluation](https://github.com/cmu-delphi/covidcast-indicators/blob/efc5f5886d7134c46b743033df5fcf5d16a4f380/backfill_corrections/delphiBackfillCorrection/R/model.R#L165). Wanted to run this change by you.

Because of this change, I had to add some fields to the fake test data in [test-beta_prior_estimation.R](https://github.com/cmu-delphi/covidcast-indicators/pull/1704/files#diff-d6f041733ee06a2788aa41f47dca9e6dd431d5cd5fd825cfa02f00a815f85c6d).

backfill_corrections/delphiBackfillCorrection/NAMESPACE
```diff
@@ -29,9 +31,11 @@ importFrom(dplyr,filter)
 importFrom(dplyr,group_by)
 importFrom(dplyr,group_split)
 importFrom(dplyr,if_else)
+importFrom(dplyr,mutate)
 importFrom(dplyr,pull)
 importFrom(dplyr,select)
 importFrom(dplyr,summarize)
+importFrom(dplyr,ungroup)
 importFrom(evalcast,weighted_interval_score)
```
nmdefries, line 39: @jingjtang `evalcast` takes a long time to install because it has a lot of dependencies. Would it be possible to define our own `weighted_interval_score` instead of importing it? How important is making sure that our definition matches that in evalcast?

jingjtang: I would say it's important. @brookslogan  Does the forecast team also use `weighted_interval_score` from `evalcast`?

backfill_corrections/delphiBackfillCorrection/R/beta_prior_estimation.R
jingjtang: It's definitely a bug. Thanks for pointing it out.

backfill_corrections/delphiBackfillCorrection/NAMESPACE
nmdefries: Okay, thanks for verifying 👍

brookslogan: Yes, we use `weighted_interval_score` from `evalcast`.  Matching its mathematical definition is very important, as this is a metric the community has developed for some evaluations.  It would be nice to have a faster implementation.  It would also be nice to have an implementation outside of `evalcast` since we're going to replace some of its functionality with `epiprocess`, but I don't know if that place should be `epiprocess` or something else.

jingjtang:
Except for the problem that I mentioned in #1712 and #1711, other codes look fine to me. The only but maybe big problem left is that I am not able to run `make run-local`. Is it finished or should I have something else installed for doing that?

jingjtang (approved):
LGTM

PR #1706, opened 2022-10-10
[Refactor] Add type hints to `geomap.py` and a small test refactor

dshemetov:
### Description
- Add type hints and improves formatting in `geomap.py`
- Refactors a few other tests in `_delphi_utils`

### Changelog
You can see the changes in the commit log.

Also adds `.git-blame-ignore-revs`, which you can use to ignore the changes in the `83cb164` commit (or whatever commits you paste into it) when using `git blame` (idea taken from [`black` documentation](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html)). If you're a GitLens user, [here's the solution](https://github.com/gitkraken/vscode-gitlens/issues/413#issuecomment-523640184).

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Format geomap.py with black
+d4b056e7a4c11982324e9224c9f9f6fd5d5ec65c
--- a/_delphi_utils_python/data_proc/geomap/geo_data_proc.py
+++ b/_delphi_utils_python/data_proc/geomap/geo_data_proc.py
@@ -12,7 +12,6 @@ from io import BytesIO
 from os import remove, listdir
 from os.path import join, isfile
 from zipfile import ZipFile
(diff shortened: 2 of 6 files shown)


.git-blame-ignore-revs
krivard, line 1: note to future travelers this is so that git-annotate skips Black (purely formatting) commits

krivard (commented):
should we refresh this or trash it?

dshemetov:
I'd say refresh. I did this because I was working geomapper at some point and the lack of argument types was confusing my IDE.

krivard:
alright, when you get a minute, rebase onto main to bring it back up to date, then i'll tag nat for a review

nmdefries (approved):
I don't see the `.git-blame-ignore-revs` file -- did you mean to remove that from the PR?

Aside from that, looks good!

nmdefries:
praise: good job finishing this off!

PR #1707, opened 2022-10-12
Docker setup for backfill corrections

nmdefries:
### Description
Create Dockerfile. Add S3 push `make` target. Add docker run `make` target.

The pipeline has been tested to work in docker on my machine. I can share params and input files to do a test run.

### Changelog
- Dockerfile
- Makefile


backfill_corrections/Makefile
```diff
@@ -15,12 +26,27 @@ dev: delphiBackfillCorrection_1.0.tar.gz
...
+	    -v "${PWD}"/receiving:/backfill_corrections/receiving \
+	    -v "${PWD}"/input:/backfill_corrections/input \
+	    -v "${PWD}"/cache:/backfill_corrections/cache \
+	    -v "${PWD}"/params.json:/backfill_corrections/params.json \
+		-i "${DOCKER_IMAGE}:${DOCKER_TAG}" \
+		/bin/bash -c "make run-local OPTIONS=\"${OPTIONS}\""
+
+deploy: run
+	# Assumes receiving is output dir
+	aws s3 cp receiving/ $(S3_BUCKET)/ --recursive --exclude "*" --include "*csv.gz" --acl public-read
```
krivard, line 48: Does this need credentials installed somehow?

backfill_corrections/Makefile
```diff
@@ -1,8 +1,19 @@
 SHELL:=/bin/bash
 
-TODAY:=$(shell date -u +"%Y-%m-%d")
-CURR_TIME:=$(shell date -u +"%Hh%Mm%Ss")
-LOG_FILE:=$(TODAY)_$(CURR_TIME).log
+DOCKER_IMAGE=ghcr.io/cmu-delphi/covidcast-indicators-backfill_corrections
+DOCKER_TAG=latest
+# Change training options during `make` call via `make <command> OPTIONS="<options>"`
+OPTIONS=--train_models --make_predictions
```
krivard, line 6: pros/cons of making the default the lengthiest and rarest configuration? it does serve as sortof documentation of the available options but are there other reasons to do it?

backfill_corrections/Makefile
```diff
@@ -1,8 +1,19 @@
...
+
+S3_URL=https://backfill_corrections.s3.us-east-2.amazonaws.com
+S3_BUCKET=s3://backfill_corrections
```
krivard, line 9: All our other pipelines read S3 locations out of params.json; is there a compelling reason to hard-code these ones?

nmdefries: Good point. I was modeling this after the `forecast-eval` Makefile. I'll take a look at other ways to set this up.

backfill_corrections/Makefile
nmdefries: There's no particular reason for doing this beyond it being convenient for manual testing.

Since we're going to be separating the training and prediction steps in production, we should always modify the `OPTIONS` to be explicit about what procedure we're intending to run (+ robustness in case default `OPTIONS` changes). So the default shouldn't have an impact there.

RE documenting options, the `run.R` file shows them pretty clearly so I wouldn't worry about that too much either way.

backfill_corrections/Makefile
nmdefries: Added an `aws configure` step right before this.

ansible/templates/backfillcorr-params-prod.json.j2
```diff
@@ -9,5 +9,12 @@
   "geo_levels": ["state", "county"],
   "value_types": ["count", "fraction"],
   "num_col": "num",
-  "denom_col": "den"
+  "denom_col": "den",
+  "archive": {
```
krivard, line 13: This isn't archive-differ data; it deserves its own section

```suggestion
  "post": {
```

backfill_corrections/Makefile
```diff
@@ -33,19 +53,41 @@ run-local:
...
 
 run:
-	# Assumes receiving is output dir
 	docker run --rm --pull=always \
-	    -v "${PWD}"/logs:/backfill_corrections/logs \
-	    -v "${PWD}"/receiving:/backfill_corrections/receiving \
-	    -v "${PWD}"/input:/backfill_corrections/input \
-	    -v "${PWD}"/cache:/backfill_corrections/cache \
+	    -v "${PWD}/${LOG_DIR}:/backfill_corrections/${LOG_DIR}" \
+	    -v "`realpath $(USR_EXPORT_DIR)`:/backfill_corrections/${EXPORT_DIR}" \
```
krivard, line 58: what's `realpath`?

nmdefries: It turns a relative path into an absolute path, which Docker volume mounting requires. I was previously combining the full path of the current working dir with the various params-provided directories, but realized there's no requirement that those directories actually exist in the current working dir or that they're not already abs paths. Thought this would be more flexible.

We may need to install `realpath`.

backfill_corrections/Makefile
krivard: mmm if we want to always specify OPTIONS explicitly, would it be better to set it empty here and fail if it hasn't been overridden?

nmdefries: Sure, that sounds good. The pipeline is already set up to [short-circuit](https://github.com/cmu-delphi/covidcast-indicators/blob/ace7a987c8bdbea144db1ee84dc4dc263cd98c24/backfill_corrections/delphiBackfillCorrection/R/main.R#L232-L235) if no options have been set, so I won't add any extra handling here.

korlaxxalrok (approved):
Looks good!

backfill_corrections/Makefile
```diff
@@ -1,28 +1,49 @@
...
+S3_BUCKET=$(shell $(PYTHON) -m delphi_utils get bucket_name)
+
 DOCKER_IMAGE=ghcr.io/cmu-delphi/covidcast-indicators-backfill_corrections
 DOCKER_TAG=latest
-# Change training options during `make` call via `make <command> OPTIONS="<options>"`
-OPTIONS=--train_models --make_predictions
 
-S3_URL=https://backfill_corrections.s3.us-east-2.amazonaws.com
-S3_BUCKET=s3://backfill_corrections
+# Static dir names for use inside Docker container
```
krivard, line 19: What happens if these names aren't the same as the ones in `params.json`?

backfill_corrections/Makefile
```diff
@@ -33,19 +54,41 @@ run-local:
...
 	    -v "${PWD}"/params.json:/backfill_corrections/params.json \
 		-i "${DOCKER_IMAGE}:${DOCKER_TAG}" \
 		/bin/bash -c "make run-local OPTIONS=\"${OPTIONS}\""
 
-deploy: run
-	# Assumes receiving is output dir
-	aws s3 cp receiving/ $(S3_BUCKET)/ --recursive --exclude "*" --include "*csv.gz" --acl public-read
+publish:
+ 	aws configure set aws_access_key_id AWS_KEY_ID
+ 	aws configure set aws_secret_access_key AWS_SECRET_KEY
```
krivard, line 68: did you mean these to be variables?
```suggestion
 	aws configure set aws_access_key_id $(AWS_KEY_ID)
 	aws configure set aws_secret_access_key $(AWS_SECRET_KEY)
```

backfill_corrections/Makefile
```diff
@@ -33,19 +54,41 @@ run-local:
...
+ 	aws s3 cp $(USR_INPUT_DIR) $(S3_BUCKET)/ --recursive --exclude "*" --include "*.csv.gz" --acl public-read
+ 	echo "SUCCESS: published `ls -1 $(USR_EXPORT_DIR)/*.csv.gz | wc -l` files to the S3 bucket" >> $(LOG_FILE)
+
+pipeline: setup run publish teardown
+
+# Make sure all required dirs exist. Set dir names to standard values for use in Docker.
+setup:
+	[ -f $(USR_INPUT_DIR) ] || mkdir -p $(USR_INPUT_DIR)
+	[ -f $(USR_CACHE_DIR) ] || mkdir -p $(USR_CACHE_DIR)
+	[ -f $(USR_EXPORT_DIR) ] || mkdir -p $(USR_EXPORT_DIR)
```
krivard, line 78: ...are we ensuring these directories exist, then immediately configuring params.json to use other directories?

nmdefries: Yep. Definitely open to suggestions here, but let me explain my thinking.

* User provides path to dirs in `params`. Sometimes dirs can be outside of the working dir. This is especially true of the input dir, which at the moment we've fixed as `/common/backfill/<indicator name>`.
* Check that dirs exist locally and create them if not. This is because we need to mount them in the docker container. If we attempt to mount a dir that doesn't actually exist locally, docker will happily use it but fail to copy the contents (output, in this case) back to the host.
* Since the dirs inside docker are obscured, I figured it would be easier if they just had fixed names/locations hence reconfiguring `params`

nmdefries: But... I guess I can just use the user-provided names inside the container directly? They won't necessarily end up in the working directory, but that should be fine.

nmdefries: An additional complication with using user-provided dir names inside the container is that we need different processing if the dir name in `params` is an absolute or a relative path. This is because when [mounting the dirs in `docker run`](https://github.com/cmu-delphi/covidcast-indicators/blob/6c140016f48045599a137c240daecd976997a1d4/backfill_corrections/Makefile#L59) we would need to specify the parent dir for relative paths.

Instead of checking if dir paths from `params` are absolute or not, we will keep the current approach and document the choice. The two approaches are equally finicky, but the current approach has been tested already.

nmdefries: Addressed in https://github.com/cmu-delphi/covidcast-indicators/pull/1707/commits/bbbb59f370dec0bdeb0c7ce836813b620544fa22

ansible/templates/backfillcorr-params-prod.json.j2
```diff
@@ -0,0 +1,25 @@
...
+  "denom_col": "den",
+  "post": {
+    "aws_credentials": {
+      "aws_access_key_id": "{{ backfillcorr_aws_access_key_id }}",
+      "aws_secret_access_key": "{{ backfillcorr_aws_secret_access_key }}"
+    },
+    "bucket_name": "{{ backfillcorr_aws_bucket_name }}"
+  },
+  "gurobi": {
+    "GRB_LICENSEID": "{{ GRB_LICENSEID }}",
```
korlaxxalrok, line 21:
```suggestion
    "GRB_LICENSEID": "{{ grb_licenseid }}",
```

ansible/templates/backfillcorr-params-prod.json.j2
```diff
@@ -0,0 +1,25 @@
...
+    "GRB_WLSACCESSID": "{{ GRB_WLSACCESSID }}",
```
korlaxxalrok, line 22:
```suggestion
    "GRB_WLSACCESSID": "{{ grb_wlsaccessid }}",
```

ansible/templates/backfillcorr-params-prod.json.j2
```diff
@@ -0,0 +1,25 @@
...
+    "GRB_WLSSECRET": "{{ GRB_WLSSECRET }}"
```
korlaxxalrok, line 23:
```suggestion
    "GRB_WLSSECRET": "{{ grb_wlssecret }}"
```

korlaxxalrok (approved):
Looks good 👍

krivard (approved):
👍

PR #1708, opened 2022-10-17
Change illegal empty array to empty string

krivard:
### Description
Apparently empty arrays aren't allowed in GitHub Action workflows, but empty strings are?

### Changelog
- build-container-images.yml: change from empty array `[ ]` to empty string `""`

### Fixes 
- Fixes [failing build](https://github.com/cmu-delphi/covidcast-indicators/actions/runs/3265501961)

### DIFF
--- a/.github/workflows/build-container-images.yml
+++ b/.github/workflows/build-container-images.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        packages: [ ]
+        packages: ""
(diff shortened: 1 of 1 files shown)


nmdefries:
If `packages` is an empty string, does the workflow try to run? Seems like we might get odd behavior.

I think it would be more straightforward to just disable this workflow for now. We'll need it shortly for the backfill corrections build so we can re-enable it in a week or so.

krivard:
Done!

PR #1711, opened 2022-10-19
Fetch modeling date from model filenames

nmdefries:
### Description
Instead of saving model training date to a separate cached file, `./cache/training_end_date.txt`, get the training end date from cached model files, which use the training date as a prefix.

### Changelog
- `main.R` - stop passing `training_end_date` to `run_backfill` as an arg, call new fn instead. Stop reading/writing training date cache file.
- `model.R` - define function to get the training end date in various cases.
- `utils.R` - define helper function to check if a key exists in `params`. Check if user-provided training end date is in the future.
- New documentation `.Rd` files


nmdefries:
What do we do if we're missing some but not all models?

backfill_corrections/delphiBackfillCorrection/R/model.R
```diff
@@ -262,3 +262,55 @@ generate_filename <- function(indicator, signal,
...
+      # We know we'll be retraining models today.
+      training_end_date <- TODAY
+    } else {
+      # If only some models are in the cache, they will be used and those
+      # missing will be regenerated as-of the training end date.
+      training_end_date <- max(as.Date(substr(model_files, 1, 10)))
+    }
+  }
+
+  training_start_date <- training_end_date - params$training_days
```
krivard, line 305: what happens if `params$training_days` is different from the `training_days` used to train the stored model? /under what circumstances would we switch to a different value for `training_days`?

backfill_corrections/delphiBackfillCorrection/R/model.R
```diff
@@ -262,3 +262,55 @@ generate_filename <- function(indicator, signal,
...
+    # Assumes filename format like `2022-06-28_changehc_covid_state_lambda0.1_count_ca_lag5_tau0.9.model`
+    # where the leading date is the training end date for that model.
+    model_files <- list.files(params$cache_dir, "202[0-9]-[0-9]{2}-[0-9]{2}*.model")
+    if (length(model_files) == 0) {
+      # We know we'll be retraining models today.
+      training_end_date <- TODAY
+    } else {
+      # If only some models are in the cache, they will be used and those
+      # missing will be regenerated as-of the training end date.
+      training_end_date <- max(as.Date(substr(model_files, 1, 10)))
```
krivard, line 301: nice!

krivard (approved):
Great work! minor Q about training_days but otherwise good to go.

backfill_corrections/delphiBackfillCorrection/R/model.R
nmdefries: > what happens if `params$training_days` is different from the `training_days` used to train the stored model?

Yeah, in that case we could end up using an incorrect model from the cache. I think it's unlikely, even with local use testing different training ranges, but it's easy enough to add `training_start_date` into the cached model names.

@jingjtang optimized the current value for `training_days` as part of the model development process, so we'd be unlikely to change it on a short time scale for production runs. I imagine we'd only want to change `training_days` if the signal behavior changed significantly and wasn't being captured well within the current training range.

nmdefries: Added this in https://github.com/cmu-delphi/covidcast-indicators/pull/1711/commits/1ecfa81aa1f5f6db02c6e75ad49fa8ecd2c5c932.

backfill_corrections/delphiBackfillCorrection/R/model.R
```diff
@@ -262,3 +267,63 @@ generate_filename <- function(indicator, signal,
...
+    } else {
+      # Default end date is today.
+      training_end_date <- TODAY
+    }
+  } else {
+    # Get end date from cached model files. Assumes filename format like
+    # `20220628_20220529_changehc_covid_state_lambda0.1_count_ca_lag5_tau0.9.model`
+    # where the leading date is the training end date for that model, and the
+    # second date is the training start date.
+    model_files <- list.files(params$cache_dir, "202[0-9]{5}_202[0-9]{5}*.model")
```
jingjtang, line 302: I always run into problem with this line. R told me that 

```
Error in list.files(params$cache_dir, "202[0-9]-[0-9]{2}-[0-9]{2}*.model") : 
  invalid 'pattern' regular expression
```

backfill_corrections/delphiBackfillCorrection/R/model.R
```diff
@@ -262,3 +267,63 @@ generate_filename <- function(indicator, signal,
...
+#'   1. We are training new models.
+#'   2. We are not training new models and cached models exist.
+#'   3. We are not training new models and cached models don't exist.
+#'
+#' Sometimes we want to allow the user to specify an end date in
+#' params that overrides the automatically-generated end date. This is
+#' only relevant when the user requests to train new models.
+#'
+#' @template params-template
+get_training_date_range <- function(params) {
```
jingjtang, line 288: Together with the problem that I mentioned in PR #1712, I would suggest that we put this function under `utils.R` or `io.R`.

jingjtang (changes requested):
Q closely related to PR #1712

backfill_corrections/delphiBackfillCorrection/R/model.R
nmdefries: Sorry, I have a fix for that but haven't pushed yet.

backfill_corrections/delphiBackfillCorrection/R/model.R
jingjtang: NVM. But remember that we don't have unit tests for `get_training_date_range` currently.

nmdefries: I'll make an issue for that and aim to address soon.

backfill_corrections/delphiBackfillCorrection/R/main.R
```diff
@@ -56,13 +56,16 @@ run_backfill <- function(df, params,
...
-
+    
     msg_ts("Splitting data into geo groups")
     group_dfs <- group_split(df, .data$geo_value)
 
     # Build model for each location
     for (subdf in group_dfs) {
       geo <- subdf$geo_value[1]
+      
+      if (!(geo %in% c("pa", "ma", "ny"))) {next}
```
nmdefries, line 67: Test code?

nmdefries (commented):
@jingjtang Other than the one line that looks like it was for testing, everything looks good.

backfill_corrections/delphiBackfillCorrection/R/main.R
jingjtang: Oh, yes! Sorry, I forgot to delete it. Ignore it currently.

jingjtang (approved):
LGTM

PR #1712, opened 2022-10-20
Make input file filter rely on user inputs

nmdefries:
### Description
The previous filter definition only included recent days (the last `ref_lag + training_days` days before "today"), regardless of user settings. However, a user can request models trained on older data or predictions made for older dates. Those cases require data entirely not in that recent range.

### Changelog
- `io.R`


jingjtang (changes requested):
It seems the current change can not help us to achieve this goal. Isn't it? 
```
However, a user can request models trained on older data or predictions made for older dates
```

backfill_corrections/delphiBackfillCorrection/R/io.R
```diff
@@ -96,41 +96,59 @@ get_files_list <- function(indicator, signal, params, sub_dir) {
...
+  #
+  # Train data doesn't have an explicit lower bound on issue_date, but we can
+  # derive one.
+  #
+  # Since target_date = reference_date + params$ref_lag and issue_date >=
+  # reference_date, the requirement that training_start_date < target_date
+  # also implies that issue date must be > training_start_date - params$ref_lag
+  result <- get_training_date_range(params)
+  training_start_date <- result$training_start_date
+  training_end_date <- result$training_end_date
```
jingjtang, line 143: The `training_start_date` and `training_end_date` here are defined entirely based on the `training_end_date` in `params` which has nothing to do with `test_date`

backfill_corrections/delphiBackfillCorrection/R/io.R
```diff
@@ -96,41 +96,59 @@ get_files_list <- function(indicator, signal, params, sub_dir) {
...
+  start_issue <- min(training_start_date - params$ref_lag, params$test_dates)
+  end_issue <- max(training_end_date, params$test_dates)
```
jingjtang, line 145: If we have an older `test_date`, but have the model trained up-to-date. Isn't the `end_issue` only based on `training_end_date`? Thus, when we actually training the model for the older `test_date`, we don't have enough issue dates.

jingjtang (changes requested):
Sorry I didn't correctly add the details in the last request

jingjtang:
So now the order of the steps are:
- Check `training_end_date` and `training_start_date` according to `params`
- Together with `test_date`, modify the `issue_start_date` and `issue_end_date`

However, if we really want to train the model for older `test_dates`, isn't the order to be:
- Check the `training_end_date` according to `test_date` and `training_end_date` from the model we trained already
- Modify the `issue_end_date` and `issue_start_date` accordingly

backfill_corrections/delphiBackfillCorrection/R/io.R
nmdefries: Thanks, that's a good observation. We can just extend the `start_issue` date, changing this line

```
start_issue <- min(training_start_date - params$ref_lag, params$test_dates)
```
to
```
start_issue <- min(training_start_date - params$ref_lag, params$test_dates - params$ref_lag)
```

backfill_corrections/delphiBackfillCorrection/R/io.R
jingjtang: I agree with you. I think the current `get_training_date_range` function is fine.

nmdefries:
Hm, I don't understand your proposal

nmdefries:
> * Check the training_end_date according to test_date and training_end_date from the model we trained already
> * Modify the issue_end_date and issue_start_date accordingly
Hm, I don't understand this bit.

The point of this function, `subset_valid_files`, is to reduce the full list of input files to only those needed for the current task. Either we are training models (and don't need testing input data), making predictions (we need both testing and training input data. The training data is in case some models are missing from the cache and we need to recreate them), or both training and predicting (we need both testing and training input data).

The two threshold dates, `issue_end_date` and `issue_start_date`, are the first and last issue dates we want to keep to get the correct input data.

`training_start_date - ref_lag` and `training_end_date` are the first and last issue dates we would want to keep if we were _only training models_ and not making any predictions.

`test_dates` are used in an issue date [filter here](https://github.com/cmu-delphi/covidcast-indicators/blob/8a7167f5e03da6dbd80977c0c7606810439ccc3e/backfill_corrections/delphiBackfillCorrection/R/main.R#L130-L132) to create testing data. So `min(test_dates - ref_lag)` and `min(test_dates)` are the first and last issue dates for input data we would want to keep for making predictions.

In cases where we need input data for both training and predicting, I'm just taking the most extreme values for the start and end issue dates (min of `training_start_date - ref_lag` and `min(test_dates - ref_lag)`, etc).

It sounds like maybe you're concerned that the [defined `training_end_date`](https://github.com/cmu-delphi/covidcast-indicators/blob/fbef1a7c7a0460efa8a823cddcf45cb2dfa8f7c9/backfill_corrections/delphiBackfillCorrection/R/model.R#L281) isn't appropriate in some cases.

> Check the `training_end_date` according to `test_date` and `training_end_date` from the model we trained already

When `train_models` is not turned on, `get_training_date_range` does check dates from cached models. If there are no cached models, `training_end_date` is set to today; if there are some cached models, `training_end_date` is set to match those and any models missing from the cache for that particular `training_end_date` are regenerated.

Unless I'm misunderstanding some bit of code logic, `training_end_date` should be entirely independent of the `test_dates`. For example, we can train a model on the last month of data and use it to make predictions for test dates from a year ago.

jingjtang:
> > * Check the training_end_date according to test_date and training_end_date from the model we trained already
> > * Modify the issue_end_date and issue_start_date accordingly
> 
> Hm, I don't understand this bit.

Backfill correction is highly time sensitive. It's inappropriate to make prediction of historical data based on a model trained using future data.

My understanding is that, if we only keep the most recent model saved, once we want to make prediction of an old test date t0 (t0<T, T is today), we always need to train the model based on data (t<t0).

nmdefries:
>  It's inappropriate to make prediction of historical data based on a model trained using future data.

> For example, we can train a model on the last month of data and use it to make predictions for test dates from a year ago.

I agree that in reality we don't want to do this, but I thought it could be useful to have the option. It sounds like you're saying that the extra flexibility is dangerous (besides just being unnecessary). So we need to add a check that `min(test_date) > training_end_date` to verify that this is true:

> make prediction of an old test date t0 (t0<T, T is today), we always need to train the model based on data (t<t0).

jingjtang:
> I agree that in reality we don't want to do this, but I thought it could be useful to have the option. It sounds like you're saying that the extra flexibility is dangerous (besides just being unnecessary). So we need to add a check that `min(test_date) > training_end_date` to verify that this is true:

Exactly.

jingjtang (approved):
LGTM

PR #1717, opened 2022-11-16
Use full `sklearn` name in `setup.py`

nmdefries:
### Description
Linting and tests are failing for `doctor-visits` because the `sklearn` package can't be found. This is related to the [`sklearn` deprecation](https://pypi.org/project/sklearn/), so, as recommended, use the full package name in the setup file.

This is the only instance of `sklearn` in all `covidcast-indicators` packages.

### Changelog
- `setup.py`

### DIFF
--- a/doctor_visits/setup.py
+++ b/doctor_visits/setup.py
@@ -5,7 +5,7 @@ required = [
     "numpy",
     "pandas",
     "paramiko",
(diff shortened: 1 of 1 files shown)


jingjtang (approved):
LGTM

nmdefries:

PR #1718, opened 2022-11-16
Process backfill_corrections with CI tools

korlaxxalrok:
### Description

Adds backfill_corrections to Python linting and Jenkins CI build/package/deploy process

### Changelog

- Update the lists of indicators to process via CI tooling:

  - python-ci.yml
  - Jenkinsfile


korlaxxalrok:
Ah, we probably don't need all of this.

PR #1760, opened 2023-01-12
[Backfill corrections] Align daily and rollup file formats; make dates portable

nmdefries:
### Description
Make sure `lag`  and `issue_date` fields are included in both daily and combined files. Store dates and location info as strings for better portability. These were previously `datetime64` (causing the timezone issue) and `object` types.

[`claims_hosp` backfill file generation](https://github.com/cmu-delphi/covidcast-indicators/pull/1675) was never merged, so changes to those functions are not included here.

### Changelog
- `quidel_covidtest`'s `backfill.py`
- `changehc`'s `backfill.py`

### Fixes
The original `time_value`s are meant to be plain dates ("2020-01-01") with no timestamp or timezone info.

The `parquet` format uses a schema with types. So if Python writes the `time_value`s as either pure dates (no timestamp info) or strings, R will read them in using the same types. The implicit timezone conversion (from no timezone, which R's `arrow::read_parquet` interprets as UTC, to the local timezone) only happens when the `time_value`s are saved into `parquet` form as `datetime64`s.

Python doesn't seem to have a "pure" date class (no time/timezone info). The `datetime64` type assumes the time is 00:00 even if none is given. To drop the time info, we can convert to a pure date but this [changes the type to `object`](https://stephenallwright.com/pandas-convert-datetime-to-date/). The `object` class can be a little dangerous to use. In this case, it's not clear what type `parquet` will assign to such a column or how R will read it in, and within Python `object`s can behave in unusual ways. Saving the dates to string is safer.

R will have to do an extra step to convert from string to a date, but it should avoid any weird time/timezone issues.

### DIFF
--- a/changehc/delphi_changehc/backfill.py
+++ b/changehc/delphi_changehc/backfill.py
@@ -46,6 +46,18 @@ def store_backfill_file(df, _end_date, backfill_dir, numtype, geo, weekday):
                         'num', 'den']
     backfilldata = backfilldata.loc[backfilldata["time_value"] >= _start_date,
                                     selected_columns]
+
+    backfilldata["lag"] = [(_end_date - x).days for x in backfilldata["time_value"]]
+    backfilldata["time_value"] = backfilldata.time_value.dt.strftime("%Y-%m-%d")
+    backfilldata["issue_date"] = datetime.strftime(_end_date, "%Y-%m-%d")
+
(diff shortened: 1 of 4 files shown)


jingjtang (approved):
LGTM

nmdefries:

PR #1763, opened 2023-01-17
[Backfill corrections] Restore original params if pipeline exits with an error

nmdefries:
### Description
Modify params dirs ([for convenience of working with user-provided paths](https://github.com/cmu-delphi/covidcast-indicators/blob/6e7df3424e08ec2f29b196152c66f1d1aaa0e9e4/backfill_corrections/Makefile#L98-L113)) only inside the docker container so that the original user-provided params file is preserved regardless of pipeline exit status or other unintended behavior.

### Changelog
- Makefile

### Fixes 
Previously modifications to dir paths were performed on the original `params.json` outside the docker container. Restoring it to its original state relied on a [`teardown` step](https://github.com/cmu-delphi/covidcast-indicators/blob/6e7df3424e08ec2f29b196152c66f1d1aaa0e9e4/backfill_corrections/Makefile#L118-L122) that didn't run if the pipeline exited with an error.

Instead of solving this by capturing and rethrowing errors, it was easier to limit params modifications to a copy of the mounted `params.json` inside the docker container. The copy is not linked to any file on the host, so changes to it are not propagated back to the original `params.json`.

I can't find any built-in support for mounting a file from the host but NOT propagating in-container changes back to the host, so `params.json` is mounted under another name and then copied manually.

### DIFF
--- a/backfill_corrections/Makefile
+++ b/backfill_corrections/Makefile
@@ -75,10 +75,10 @@ run:
 		-v "`realpath $(USR_EXPORT_DIR)`:/backfill_corrections/${EXPORT_DIR}" \
 		-v "`realpath $(USR_INPUT_DIR)`:/backfill_corrections/${INPUT_DIR}" \
 		-v "`realpath $(USR_CACHE_DIR)`:/backfill_corrections/${CACHE_DIR}" \
-		-v "${PWD}"/params.json:/backfill_corrections/params.json \
+		-v "${PWD}"/params.json:/backfill_corrections/params.host.json \
 		--env GRB_LICENSE_FILE=$(GRB_LICENSE_FILE) \
 		-it "${DOCKER_IMAGE}:${DOCKER_TAG}" \
-		/bin/bash -c "make gurobi.lic && make run-local OPTIONS=\"${OPTIONS}\""
+		/bin/bash -c "cp params.host.json params.json; make gurobi.lic && make standardize-dirs && make run-local OPTIONS=\"${OPTIONS}\""
 
 publish:
 	aws configure set aws_access_key_id $(AWS_KEY_ID)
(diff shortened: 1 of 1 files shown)


backfill_corrections/Makefile
```diff
@@ -75,18 +75,18 @@ run:
 		-v "`realpath $(USR_EXPORT_DIR)`:/backfill_corrections/${EXPORT_DIR}" \
 		-v "`realpath $(USR_INPUT_DIR)`:/backfill_corrections/${INPUT_DIR}" \
 		-v "`realpath $(USR_CACHE_DIR)`:/backfill_corrections/${CACHE_DIR}" \
-		-v "${PWD}"/params.json:/backfill_corrections/params.json \
+		-v "${PWD}"/params.json:/backfill_corrections/params.host.json \
 		--env GRB_LICENSE_FILE=$(GRB_LICENSE_FILE) \
 		-it "${DOCKER_IMAGE}:${DOCKER_TAG}" \
-		/bin/bash -c "make gurobi.lic && make run-local OPTIONS=\"${OPTIONS}\""
+		/bin/bash -c "cp params.host.json params.json; make gurobi.lic && make standardize-dirs && make run-local OPTIONS=\"${OPTIONS}\""
```
krivard, line 81: I can't think of a scenario where this would matter, but is it worth using && here anyway?
```suggestion
		/bin/bash -c "cp params.host.json params.json && make gurobi.lic && make standardize-dirs && make run-local OPTIONS=\"${OPTIONS}\""
```

krivard (approved):
Nice find! It's always great when you can drop a target without loss of functionality.

Suggestion optional; probably works either way.

nmdefries:

PR #1765, opened 2023-01-20
CHNG HC exception fix

M5Skid:
### Description
Fixes exception after the assertion in update_sensor.py (indicator no longer crashes):

- assert (len(multiindex) <= (len(gmpr.get_geo_values(gmpr.as_mapper_name(geo))) * len(self.fit_dates))
AssertionError: more loc-date pairs than maximum number of geographies x number of dates

New geo_ids with different amounts of digits are appearing in the source data and being counted as counties, so the multiindex created here is much larger than the list of valid geomapper fips values. 
When checking county geo_values, those with less than 5 digits will now be dropped.

Note: It is possible there are more points of failure caused by the bad geo_id values.
- May be other places in the code that are parsing the bad source data.

### Changelog
update_sensor.py
- Check the length of the county ids and find the index in the dataframe.
- Drop all rows containing bad indexes (geo_id less than 5 digits).

### Fixes 
- Fixes current chng exception


krivard (changes requested):
This is a good start -- now add tests for the following:
* verify that a data frame with only 5-char county codes is unaffected
* verify that a data frame with all geos != county is unaffected
* verify that a data frame with <5 and >5 -char county codes is modified the way we expect

[tests are located here](https://github.com/M5Skid/covidcast-indicators/tree/chnghc_fix/changehc/tests)

PR #1767, opened 2023-01-25
[Backfill corrections] Allow package to process one indicator at a time

nmdefries:
### Description
Let user pass a command line argument to the backfill corrections package to process either all indicators at once or only a single specified indicator. The `INDICATORS_AND_SIGNALS` table is filtered down to only entries matching the provided indicator. If running on a single indicator, only model files for that indicator will be removed from the cache.

Add a deduplication step into `validity_checks`. Duplicate entries cause the `pivot` steps to produce dfs full of `list` objects that don't behave with downstream calculations.

### Changelog
- Makefile
- `main.R`
- `utils.R`
- `run.R`


jingjtang (approved):
LGTM

nmdefries:

PR #1772, opened 2023-01-31
[Backfill corrections] check if any files exist in receiving before trying to publish

nmdefries:
### Description
Don't bother with connecting to AWS or trying to publish if we haven't actually generated any files, such as when doing model training.

### Changelog
- Makefile


krivard (approved):
Solid choice; verified it works

PR #1773, opened 2023-01-31
Avoid dynamic validation when no files are exported

### LINKED ISSUE #1746: Validator throws an exception when no output is produced by the indicator

M5Skid:
### Description
Fixes exceptions caused by dynamic validation, where dataframe commands were failing on empty dataframes. If no export files are found after the indicator run, dynamic validation will be skipped as the static validation will perform the check for an empty file list. 
- In the case of an empty file list, the "check_empty_filelist" check should be the only validation step performed, and can be suppressed in params.

### Changelog
validate.py
- dynamic validation included in the if statement checking the length of the frames_list. If frames list is empty, dynamic validation will not occur.

### Fixes 
- Fixes #1746


_delphi_utils_python/delphi_utils/validator/validate.py
```diff
@@ -63,5 +63,5 @@ def validate(self):
             all_frames = []
         else:
             all_frames = aggregate_frames(frames_list)
-        self.dynamic_validation.validate(all_frames, report)
+            self.dynamic_validation.validate(all_frames, report)
```
krivard, line 66: suggestion: all_frames isn't needed when frames_list is empty

could we replace the entire if block with something like this instead?

```
if len(frames_list) > 0:
    self.dynamic_validation.validate(aggregate_frames(frames_list), report)

```

M5Skid: That makes sense and would look much cleaner. I'll put that in and do some testing.

krivard (approved):
👍 good fix!

PR #1781, opened 2023-02-06
Make log file available to docker subcall so only one log is produced

nmdefries:
### Description
Pass log file name from the `make` session on the host to the `make` session inside of docker so that all messages end up in a single log file instead of in two separate files.

### Changelog
- Makefile


krivard (approved):
👍 

+verified that LOG_FILE never contains whitespace

PR #1782, opened 2023-02-07
[Backfill corrections] Support output files in `receiving` subdirectories

nmdefries:
### Description
Output files are saved to directories of `receiving` rather than directly to `receiving`. The line `if [ -f $(USR_EXPORT_DIR)/*.csv.gz ]` thus never triggers uploading the files to the S3 bucket. Switch to using `find` so we can look recursively for output files.

### Changelog
- Makefile


krivard (approved):
solid fix

nmdefries:

PR #1783, opened 2023-02-10
DSEW - include new header logic for fully vaccinated signal

M5Skid:
### Description
- Fix current DSEW CPR outage where the data drops on Fridays are not being picked up.
- On the 2023/01/12 report, the "People who are fully vaccinated..." headers were renamed to "People with a completed primary series...".
  -  Based on header descriptions, this looks to be the same data.
  -  The fully vaccinated signal was not finding data to process because the code would not pick up the new header.
- Looks similar to a past name change for this signal on 2021/03/09.
  - The code was updated to include "People with a full course".

### Changelog
delphi_dsew_community_profile_report.pull
- Updated retain_header function to include "People with a completed primary series...".
  - header is now placed in select list (list of headers).
- Updated sig_select code in  _parse_sheet function to include entries with new header name.
  - New header is now placed into sig_select for fully vaccinated signal.
  - Now len(sig_select) > 0 for fully vaccinated signal

### Fixes 
- PO-35

### DIFF
--- a/dsew_community_profile/delphi_dsew_community_profile/pull.py
+++ b/dsew_community_profile/delphi_dsew_community_profile/pull.py
@@ -298,6 +298,13 @@ class Dataset:
             # exclude "People who are fully vaccinated - 12-17" ...
             header.find("-") < 0,
         ]) or all([
+            # include "People with a completed primary series"
+            header.startswith("People with a completed primary series"),
+            # exclude "People with a completed primary series as % of adult population"
+            header.find("%") < 0,
+            # exclude "People with a completed primary series - ages 65+"
+            header.find("-") < 0,
+        ]) or all([
             # include "People with full course administered"
             header.startswith("People with full course"),
             # exclude "People with full course administered as % of adult population"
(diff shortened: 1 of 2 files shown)


M5Skid:
Some validation failures were occurring but after looking into them, they weren't from the new code but possibly some new data. These seem normal and no issues were identified when not running the validator.

krivard (changes requested):
Nice work! It took me a minute to believe in the `sig_select` change but I think it's probably right. It's a shame we don't have better coverage of the `sig_select` code; it's starting to get gnarly. I'm going to add a GitHub issue for it so we can address that at a later date.

**todo**: missing updates to `retain_header` unit test

We _do_ have good test coverage for `retain_header`, however, and this PR should include changes to that test. Please add the previous header and current header to [the test for retain_header](https://github.com/M5Skid/covidcast-indicators/blob/cad0f057e1e9bb514657d6dec066f2d2078b8378/dsew_community_profile/tests/test_pull.py#L124) to make sure this change won't prevent us from correctly processing old files.

dsew_community_profile/delphi_dsew_community_profile/pull.py
```diff
@@ -303,6 +303,8 @@ def retain_header(header):
             # exclude "People with a completed primary series as % of adult population"
             header.find("%") < 0,
             # exclude "People with a completed primary series - ages 65+"
+            header.find(" age") < 0,
```
krivard, line 306: question/praise: did the test reveal this missed corner case?

if so, nicely done! i mean nicely done regardless, but it's particularly good to find this stuff out in tests rather than get a surprise in prod.

krivard (approved):
👍 good adjustment!

PR #1784, opened 2023-02-10
[Backfill corrections] Un-log predictions before writing

nmdefries:
### Description
We model logged values so predicted values are also logged. To make output more intuitively comparable to reported values, exponentiate in output. Add a function and tests to support this.

### Changelog
- `main.R`
- `model.R`


jingjtang (approved):
LGTM

nmdefries:

PR #1786, opened 2023-02-20
Fix exception during flash phase of Runner

krivard:
### Description

Flash is crashing in currently-deployed code, though it does not prevent subsequent runner phases from executing. This PR fixes the exception.

### Changelog

- runner.py - use `flash_fn`, not its result, as the input to Timer; remove superfluous check

### Fixes 
- Fixes nonblocking exception in prod

### DIFF
--- a/_delphi_utils_python/delphi_utils/runner.py
+++ b/_delphi_utils_python/delphi_utils/runner.py
@@ -66,14 +66,12 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
     validator = validator_fn(params)
     archiver = archiver_fn(params)
 
-    if flash_fn:
-        t = threading.Timer(timer, flash_fn(params))
-        t.start()
-        t.join(timer)
-        if t.is_alive():
-            t.cancel()
-            t.join()
-
+    t = threading.Timer(timer, flash_fn, args=[params])
+    t.start()
+    t.join(timer)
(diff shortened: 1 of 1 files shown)


Ananya-Joshi (approved):
This makes more sense - just for information, what does "lambdas always evaluate to true" mean? Thank you!

krivard:
a `lambda` is a python object. python conditionals can test objects, not just boolean values and boolean expressions. The "truthiness" of an object determines whether, if you use it as the test in a conditional, that test evaluates the same way it would if you had substituted `True`.

Things that are truthy (=evaluate as if you used `True`):
* 1
* True
* "abc"
* [1]
* {1:1}
* (1,)
* lambda x: x

Things that are not-truthy (="falsy"; =evaluate as if you used `False`):
* 0
* False
* None
* ""
* []
* {}
* tuple()

Things that are truthy _even though it seems like they shouldn't be_:
* "False"
* lambda x: False
* "0"
* -1

You can test things like this out in a REPL:

```python
Python 3.8.10 (default, Nov 14 2022, 12:59:47) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> foo = lambda x: False
>>> print("yes" if foo else "no")
yes
>>> print("yes" if foo(0) else "no")
no
```

PR #1796, opened 2023-02-23
[Backfill corrections] hone AWS publish behavior

nmdefries:
### Description
- Since our AWS bucket has a policy enforcing who can upload and who can view, remove `--acl` option from the upload command.
- Only upload prediction files, not coefficient files.
- Check that AWS upload step uploaded the number of files that we expected; if not, error.

### Changelog
- Makefile

### DIFF
--- a/backfill_corrections/Makefile
+++ b/backfill_corrections/Makefile
@@ -89,7 +89,9 @@ publish:
 	if [[ $$NUM_FILES -gt 0 ]]; then \
 		aws configure set aws_access_key_id $(AWS_KEY_ID); \
 		aws configure set aws_secret_access_key $(AWS_SECRET_KEY); \
-		aws s3 cp $(USR_EXPORT_DIR) $(S3_BUCKET)/ --recursive --exclude "*" --include "*/prediction*.csv.gz"; \
+		AWS_MSG=`aws s3 cp $(USR_EXPORT_DIR) $(S3_BUCKET)/ --recursive --exclude "*" --include "*/prediction*.csv.gz"`; \
+		echo "$${AWS_MSG}"; \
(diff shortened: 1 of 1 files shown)


backfill_corrections/Makefile
```diff
@@ -85,11 +85,13 @@ run:
...
 publish:
-	NUM_FILES=`find $(USR_EXPORT_DIR) -name "*csv.gz" | wc -l`; \
+	NUM_FILES=`find $(USR_EXPORT_DIR) -name "prediction*.csv.gz" | wc -l`; \
 	if [[ $$NUM_FILES -gt 0 ]]; then \
 		aws configure set aws_access_key_id $(AWS_KEY_ID); \
 		aws configure set aws_secret_access_key $(AWS_SECRET_KEY); \
-		aws s3 cp $(USR_EXPORT_DIR) $(S3_BUCKET)/ --recursive --exclude "*" --include "*.csv.gz" --acl public-read; \
+		AWS_MSG=`aws s3 cp $(USR_EXPORT_DIR) $(S3_BUCKET)/ --recursive --exclude "*" --include "*/prediction*.csv.gz"`; \
+		echo "$${AWS_MSG}"; \
+		[[ `echo "$${AWS_MSG}" | grep "upload:" | wc -l` -eq $$NUM_FILES ]] || ( echo "ERROR: Not all $${NUM_FILES} expected files were uploaded" && exit 74 ); \
```
krivard, line 94: thought (nonblocking): it really feels like there should be a way to do this with tee or process substitution instead of variables and echo

I played with it for a while and the closest I got was teeing to stderr, which would probably be excluded from the log and that's not what we want... and your way is definitely more readable.

nmdefries: Yeah, it's awkward to have the file upload running as a side effect.

There are some suggestions online to `tee` to `/dev/tty` or `$(tty)` but I can't actually get this to work on my own computer. Something to look into.

krivard: that could work when running from a terminal, but normal pipeline operations typically run headless alas

nmdefries:
This is ready to merge

PR #1799, opened 2023-02-27
fix the bug for 02-29

jingjtang:
### Description
In backfill, for each issue_date, we only keep the data for one year. The code that we used previously does not take 02-29 into consideration which only appear once every four years.

### Changelog
- backfill.py: If the last date=XX-02-29, we go back to date (XX-1)-02-28.


nmdefries (approved):

jingjtang:
@krivard This PR is ready to be merged.

PR #1800, opened 2023-02-28
FlaSH Timeout Code in Runner not running FlaSH

Ananya-Joshi:
FlaSH threading Timer function prevents the function from running for x amt. of time, and thus doesn't function as a timeout. The following changes should have the flash module time out appropciately


_delphi_utils_python/delphi_utils/runner.py
```diff
@@ -66,13 +66,13 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
     validator = validator_fn(params)
     archiver = archiver_fn(params)
 
-    t = threading.Timer(timer, flash_fn, args=[params])
+    t = threading.Thread(target=flash_fn, args=[params])
     t.start()
     t.join(timer)
     if t.is_alive():
-        t.cancel()
         t.join()
```
krivard, line 73: question: is my understanding correct?

here's how I'm reading this, but I haven't written threading code in a while so could you check me?

```
line 70: start running flash_fn in thread t
line 71: wait `timer` seconds, or for t to finish, whichever happens first
line 72: if t is still running, then
line 73: wait for t to finish (possibly a very long time)
```

Is the intent here to do something like this? https://stackoverflow.com/questions/34562473/most-pythonic-way-to-kill-a-thread-after-some-period-of-time

Ananya-Joshi: Ah I think you're right, let me redo/test this as the folks here have, thanks Katie!

Ananya-Joshi: Ah the top suggested solution didn't work here, but I think the final solution posted on that thread is similar to the implementation provided here.

Ananya-Joshi: Do you have a preference among any of the following methods: https://www.geeksforgeeks.org/python-different-ways-to-kill-a-thread/

krivard: Short term: The [PyThreadState_SetAsyncExc](https://docs.python.org/3/c-api/init.html?highlight=pythreadstate_setasyncexc#c.PyThreadState_SetAsyncExc) approach will be fine.

Long term: their "Using a hidden function _stop()" method, which doesn't actually need to have anything to do with hidden functions -- just a subclass of Thread, some way to break the thread's work down into a bunch of little chunks, some way to iterate over them within the run() function, and an Event that can be poked when the timer is up. Happy to meet to break down how this would work in FlaSH.

Ananya-Joshi:
I added the changes we discussed about threading & downloading files from AWS!

_delphi_utils_python/delphi_utils/runner.py
```diff
@@ -66,13 +67,16 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
...
-
-
+    t1 = multiprocessing.Process(target=flash_fn, args=[params])
+    t1.start()
+    start = time.time()
+    while time.time()-start < timer:
+        if not t1.is_alive():
+            break
+        time.sleep(10)
+    if t1.is_alive():
```
krivard, line 77: suggestion: this is an ideal candidate for `while/else`

```suggestion
    else:
```

The `else` clause executes only when `while` exits without breaking; https://stackoverflow.com/a/3295949/551433

_delphi_utils_python/tests/flash_ref/confirmed_incidence_num/last_7_1.csv
krivard, line 1: question: is this file still needed? 

it's only 236KB, but it also has over 3k columns in it. if we need full representation of geos for adequate unit tests, then we should keep it, but if we can switch to pseudo-geos then we should do so.

krivard (changes requested):
Mostly documentation updates. The functionality changes look good!

krivard (approved):
👍 nice work!

PR #1803, opened 2023-03-03
generate local county mapping from CHNG spreadsheet

nmdefries:
### Description
In case CHNG makes changes to their county mappings, include the code that takes the Google spreadsheet, assigns group numbers, and saves locally. Be careful to keep from reusing group numbers within a state. Doing so would make historical data uninterpretable or difficult to use/document.

### Changelog
- `geo_data_proc.py`


_delphi_utils_python/data_proc/geomap/geo_data_proc.py
```diff
@@ -548,6 +543,99 @@ def derive_chngfips_state_crosswalk():
...
+    return county_groups
+
+
+def assign_county_groups():
+    county_groups = fetch_county_groups_spreadsheet()
+
+    # If a `lowpop_county_groups.csv` already exists in `data_proc/geomap`, we
+    # have to be careful to not reassign a group number to a different group.
+    # Group numbers must remain fixed, even if a given county group is no longer
+    # being used.
```
krivard, line 588: suggestion: add tests for this or defer to when it's needed

I'm glad you thought about how to handle updating the group mapping, but the procedure is sufficiently subtle that I'd feel better about it if we had tests for it, and that would mean figuring out how to write and run tests for code that's not in the actual python module. That sounds like work that won't be needed for a while though, so we could also decide to leave handling the already-exists case aside until we actually have a group-change situation to handle: either drop the code from this PR or throw an exception so we don't run it by mistake.

_delphi_utils_python/data_proc/geomap/geo_data_proc.py
```diff
@@ -548,6 +543,99 @@ def derive_chngfips_state_crosswalk():
...
+    # included FIPS code is also listed in its own unnamed column. County groups
+    # can contain varying numbers of counties, and if CHNG provides new or updated
+    # county groupings the number of unnamed columns listing included counties
+    # could change. Use a general approach to find all of them.
+    county_cols = [colname for colname in county_groups.columns if colname.startswith("Unnamed: ")]
+
+    new_names = {
+        "state FIPS": "state_fips",
+        "county FIPS grouping": "fips_list",
+        **{ colname: ("county" + str(i + 1)) for i, colname in enumerate(county_cols)}
```
krivard, line 572: question: are the `countyNN` columns ever used?

iirc those columns were ones I added for some analysis preprocessing, and didn't actually come from CHNG. if something downstream depends on them, we can retain the preprocessing step, but if not we may as well exclude them from the df

krivard (changes requested):
nice refactor, but it might be _too_ nice -- we should resolve the already-exists handling before merging.

_delphi_utils_python/data_proc/geomap/geo_data_proc.py
nmdefries: Oh, I didn't realize the `countyNN` fields weren't in the original spreadsheet. They're used to [turn the spreadsheet into long format](https://github.com/cmu-delphi/covidcast-indicators/blob/ndefries/geomapper/popsafe-county-level/_delphi_utils_python/data_proc/geomap/geo_data_proc.py#L495-L496), but I could also split `fips_list` if we don't want to rely on them being available.

dshemetov: I got nerd sniped, so here's split trick, if you didn't get to it yet
```py
sheet_id = "1PEce4CjjHbRM1Z5xEMNI6Xsq_b2kkCh0"
sheet_name = "groupings"
# Request sheet in CSV format via tag in URL.
url = f"https://docs.google.com/spreadsheets/d/{sheet_id}/gviz/tq?tqx=out:csv&sheet={sheet_name}"

county_groups = pd.read_csv(
        url, dtype="string", index_col=False
    ).dropna(
        how="all", axis=1
    )
county_groups = county_groups[["popsafe-id", "state FIPS", "county FIPS grouping"]]
county_groups["state FIPS"] = county_groups["state FIPS"].astype(int)
county_groups = pd.concat([county_groups, county_groups["county FIPS grouping"].str.split("|", expand=True)], axis=1)
```

_delphi_utils_python/data_proc/geomap/geo_data_proc.py
dshemetov: vote: keep the old+new groups merging code so we have a starting point if there's a change in the future, but by default throw an exception if `not county_groups.empty`. 

(agree with Katie: this code is hard to reason about without having a REPL and the data open, but unit testing this is probably not worth it, since a small change in their format could break our code)

dshemetov (changes requested):
Second Katie's comment. Should be g2g after!

_delphi_utils_python/data_proc/geomap/geo_data_proc.py
```diff
@@ -548,6 +549,103 @@ def derive_chngfips_state_crosswalk():
...
+            return
+        # Grouping spreadsheet contains rows not seen in old, on-disk county
+        # groupings file. Combining the two is delicate. While the code below
+        # appears to work, it has not been formally tested and could be
+        # invalid for even small changes to the format of the input county
+        # groupings file.
+        else:
+            raise NotImplementedError(
+                "Can't combine old and new county groupings automatically, "
+                "code is not tested or robust to changes in input format"
```
dshemetov, line 610:
```suggestion
                "code below is not tested or robust to changes in input format."
                "We recommend manually working with the code below and the new"
                "data in a REPL."
```

dshemetov (approved):
Minor nit and 🚀

nmdefries:
@korlaxxalrok This is ready to merge.

PR #1807, opened 2023-03-17
[Backfill corrections] Efficient date handling and rolling averages

nmdefries:
### Description
Various changes to make modeling/predicting faster. The main changes are:

- Keep date fields (`time_value`, `issue_date`) in string format. Explicitly convert to date when required. This reduces the (large) number of implicit date-to-string conversions happening in other functions.
- Replace `zoo::rollmeanr` with `RcppRoll::roll_mean` for speed. Unlike most other fast rolling average functions available in R (e.g. `stats::filter`), `RcppRoll::roll_mean` uses the same more-precise method as `zoo`; the speedup is only because the new function is implemented in C++.

Modeling with these changes is 20% faster than after https://github.com/cmu-delphi/covidcast-indicators/pull/1802 changes and 33% faster overall. The test example is smaller than what would be used in production.

### DIFF
--- a/backfill_corrections/Dockerfile
+++ b/backfill_corrections/Dockerfile
@@ -21,7 +21,6 @@ RUN apt-get update && apt-get install -qq -y \
 
 RUN install2.r --error \
     roxygen2 \
-    zoo \
     Rglpk \
     argparser
 


jingjtang (approved):
LGTM

jingjtang (approved):
LGTM

nmdefries:
@korlaxxalrok This is ready to merge

nmdefries:

PR #1808, opened 2023-03-21
[Backfill corrections] Run geo loops in parallel

nmdefries:
### Description
Parallelize model training/predicting over geos (states or counties). Update production params to use 2 cores by default, which speeds up modeling/training by about 2x. My local runs with more than 2 cores were memory-limited and adding additional cores didn't increase speed that much, but that's worth looking at again in staging.

### Changelog
- Production and local params template files
- DESCRIPTION
- NAMESPACE
- `main.R`


jingjtang (approved):
LGTM

nmdefries:

PR #1805, opened 2023-03-10
Google-Symptoms suppress s05 low export errors

M5Skid:
### Description
The google-symptoms indicator has been exporting minimal data for the s05 signals, which is causing validation failures and blocking acquisition. Specifically, the s05_raw and s05_smoothed signals for the MSA, HRR, and county geo levels are causing the validation failures. They are receiving critical errors for the check_missing_geo_sig_date_combo, empty_reference_data and check_min_max_date validation checks. 

Params files and templates have been updated for google symptoms to suppress errors for these geo-signal pairs. The indicator should now run even with no export from these particular geo-signal pairs.

### Changelog
Add geo-signal pairs to the suppressed errors section
- ansible/templates/google_symptoms-params-prod.json.j2
- google_symptoms/params.json.template 

### Fixes 
- PO-30 (Google-Symptoms Outage)

PR #1813, opened 2023-03-24
Revert "Google-Symptoms suppress s05 low export errors"

M5Skid:
Reverts cmu-delphi/covidcast-indicators#1805

Following the google symptoms outage, s05 signals are now behaving as expected for all geo_types. We should avoid suppressing these errors in case more issues arise.


krivard (approved):
👍

Issue #1820, opened 2023-04-04
Upgrade indicators to pandas v2.0.0


krivard:
Pandas v2.0.0 was released 2023-04-03, and includes a few breaking changes. Among them:

- `.sum()` now handles all types by default (e.g. strings, which get concatenated). There's a parameter to restrict to numeric types only.
- `.append()` not allowed for data frames
- `.iteritems()` not allowed for series

We've pinned to <2.0 for now, but we should start working on fixing our code to work with v2.

melange396:
Are we sure #1825 was fully tested for pandas v2 compatibility?  I discovered recently that we are ***still*** stuck with "`pandas<2`" for the indicators (because the `covidcast` client has that requirement, and `delphi_utils` requres `covidcast`, and `delphi_utils` is imported just about everywhere in this repo).  See https://github.com/cmu-delphi/covidcast-indicators/issues/1972#issuecomment-2197688725

melange396:
* #1994 could have been (and could yet be) helpful for this

dshemetov:
FWIW, https://github.com/cmu-delphi/covidcast/pull/618 (covidcast Pandas<2 pin PR) was merged after #1825, so at the very least, the latter was definitely tested on Pandas>2. Nat made all the tests pass and I did a number of passes through the changelog and searching the repo for changed functions, trying to be thorough. 

Side-note: it'd be nice if we had a way to easily AB test changes. Like, if we could deploy a Pandas>2 branch and compare its outputs with prod for a few weeks, we could increase our confidence in the change.

melange396:
[...]

PR #1821, opened 2023-04-04
Pin pandas below v2

krivard:
### Description
Temporary fix while we wait for #1820.

Pandas v2 was release on April 3 and includes breaking changes. This PR pins the version under v2.

### Changelog

- pin pandas version in delphi-utils (which is imported by everybody, so will be transitive in all new builds)
- add a test which fails in pandas v2. this was added while chasing down the problem, but will create a good place to start on #1820.

### Fixes

Fixes failing build in `main`

### DIFF
--- a/_delphi_utils_python/setup.py
+++ b/_delphi_utils_python/setup.py
@@ -14,7 +14,7 @@ required = [
     "mock",
     "moto",
     "numpy",
-    "pandas>=1.1.0",
+    "pandas>=1.1.0,<2",
     "pydocstyle",
     "pylint==2.8.3",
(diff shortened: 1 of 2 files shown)


dshemetov (approved):
Looks good! (Reiterating our Slack exchange):
- since delphi_utils is always installed first for every indicator, this pin takes precedence over the indicator's pin (and most of them are unpinned in the setup.py)
- the indicators on prod have not been rebuilt with the new Pandas, so they are fine
- when this is released, the indicators will be rebuilt with a <2 pin, so they should stay where they are now

PR #1825, opened 2023-04-04
changehc

nmdefries:
### Description

### Changelog
- change1
- change2

### Fixes 
- Fixes #(issue)

### DIFF
--- a/changehc/delphi_changehc/load_data.py
+++ b/changehc/delphi_changehc/load_data.py
@@ -71,7 +71,7 @@ def load_chng_data(filepath, dropdate, base_geo,
     ), "Counts must be nonnegative"
 
     # aggregate age groups (so data is unique by date and base geography)
-    data = data.groupby([base_geo, Config.DATE_COL]).sum()
+    data = data.groupby([base_geo, Config.DATE_COL]).sum(numeric_only=True)
     data.dropna(inplace=True)  # drop rows with any missing entries
 
     return data
--- a/changehc/tests/test_update_sensor.py
+++ b/changehc/tests/test_update_sensor.py
@@ -91,7 +91,7 @@ class TestCHCSensorUpdater:
                 "timestamp": [pd.Timestamp(f'03-{i}-2020') for i in range(1, 14)]})
             data_frame = su_inst.geo_reindex(test_data)
             assert data_frame.shape[0] == multiple*len(su_inst.fit_dates)
-            assert (data_frame.sum() == (4200,19000)).all()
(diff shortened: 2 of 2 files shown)


doctor_visits/delphi_doctor_visits/sensor.py
```diff
@@ -61,13 +61,15 @@ def fill_dates(y_data, dates):
...
                 pd.DataFrame(dict.fromkeys(cols, 0.0), columns=cols, index=[first_date])
-            )
+            ])
         if last_date not in y_data.index:
-            y_data = y_data.append(
+            y_data = pd.concat([
+                y_data,
                 pd.DataFrame(dict.fromkeys(cols, 0.0), columns=cols, index=[last_date])
-            )
+            ])
```
krivard, line 72: question: how big is `y_data` typically?

`pd.concat` has a bunch of expensive overhead that increases with the number of rows, so if `y_data` is big it might be worth building up a list of data frames and then concatenating just once at the end.

dsew_community_profile/tests/test_pull.py
```diff
@@ -240,8 +240,8 @@ def test_nation_from_state(self):
                 'sample_size': [None, None],
                 'publish_date': [datetime(year=2020, month=1, day=1)]*2,})
 
-        pa_pop = int(state_pop.loc[state_pop.state_id == "pa", "pop"])
-        wv_pop = int(state_pop.loc[state_pop.state_id == "wv", "pop"])
+        pa_pop = int(state_pop.loc[state_pop.state_id == "pa", "pop"].iloc[0])
+        wv_pop = int(state_pop.loc[state_pop.state_id == "wv", "pop"].iloc[0])
```
krivard, line 244: wacky. good catch!

krivard (approved):
👍 
Can merge as-is, question minor

doctor_visits/delphi_doctor_visits/sensor.py
nmdefries: Whoops, thanks for pointing that out!

krivard (approved):
🙌 stellar adjustment, I totally missed that the other appends had been inside for() loops 😕

dshemetov (approved):
I did a pass through the deprecated list of options [here](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v2.0.0.html#removal-of-prior-version-deprecations-changes) to double check if we're using any of them. Didn't find any others.

Looks good to me 👍

nmdefries:
@krivard This is ready to merge but we should wait to release until early next week in case there are any pandas v2-related issues we missed.

PR #1832, opened 2023-04-13
[Validator] Cast timestamp to date in reference vs test data up-to-dateness comparison

nmdefries:
### Description
Fix `TypeError` when comparing a `datetime.date` and a `datetime64` timestamp.

### Changelog
- `validator/dynamic.py`
- Validation tests

### Fixes
https://delphi-org.slack.com/archives/C01LZ3A2UMU/p1681391375854509

### DIFF
--- a/_delphi_utils_python/tests/validator/test_dynamic.py
+++ b/_delphi_utils_python/tests/validator/test_dynamic.py
@@ -465,3 +465,38 @@ class TestDataOutlier:
 
         assert len(report.raised_warnings) == 2
         assert report.raised_warnings[0].check_name == "check_positive_negative_spikes"
+
+class TestDateComparison:
+    params = {
+        "common": {
+            "data_source": "",
+            "span_length": 1,
+            "end_date": "2020-09-02"
+        }
+    }
+
+    def test_date_comparison_by_type(self):
+        validator = DynamicValidator(self.params)
+        report = ValidationReport([])
+
+        ref_val = [30, 30.28571429, 30.57142857, 30.85714286, 31.14285714,
+                   31.42857143, 31.71428571, 32, 32, 32.14285714,
+                   32.28571429, 32.42857143, 32.57142857, 32.71428571,
+                   32.85714286, 33, 33, 33, 33, 33, 33, 33, 33,
+                   33, 33, 33, 33.28571429, 33.57142857, 33.85714286, 34.14285714]
+        test_val = [100, 100, 100]
+
+        ref_data = pd.DataFrame({"val": ref_val, "se": [np.nan] * len(ref_val),
+                    "sample_size": [np.nan] * len(ref_val), "geo_id": ["1"] * len(ref_val),
+                    # datetime64 type
+                    "time_value": pd.date_range(start="2020-09-24", end="2020-10-23")})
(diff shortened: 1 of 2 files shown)


_delphi_utils_python/tests/validator/test_dynamic.py
```diff
@@ -465,3 +465,38 @@ def test_source_api_overlap(self):
...
+
+    def test_date_comparison_by_type(self):
+        validator = DynamicValidator(self.params)
+        report = ValidationReport([])
+
+        ref_val = [30, 30.28571429, 30.57142857, 30.85714286, 31.14285714,
+                   31.42857143, 31.71428571, 32, 32, 32.14285714,
+                   32.28571429, 32.42857143, 32.57142857, 32.71428571,
+                   32.85714286, 33, 33, 33, 33, 33, 33, 33, 33,
+                   33, 33, 33, 33.28571429, 33.57142857, 33.85714286, 34.14285714]
```
krivard, line 486: suggestion: remove unnecessary complexity

Do the exact values & length matter for this, or can we simplify for easier reading?

```suggestion
        ref_val = [30, 30, 30]
```

_delphi_utils_python/tests/validator/test_dynamic.py
```diff
@@ -465,3 +465,38 @@ def test_source_api_overlap(self):
...
+        test_val = [100, 100, 100]
+
+        ref_data = pd.DataFrame({"val": ref_val, "se": [np.nan] * len(ref_val),
+                    "sample_size": [np.nan] * len(ref_val), "geo_id": ["1"] * len(ref_val),
+                    # datetime64 type
+                    "time_value": pd.date_range(start="2020-09-24", end="2020-10-23")})
```
krivard, line 492: suggestion: reduce cognitive load

if we need all 30 values in `ref_val`, something like this:

```suggestion
                    
                    "time_value": pd.date_range(start=START, end=START + datetime.timedelta(days=30))})
```

but if we can do with <6 values in `ref_val`, can leave it as constants in both start and end

(making the reader do date math over month boundaries in their head is mean)

_delphi_utils_python/tests/validator/test_dynamic.py
```diff
@@ -465,3 +465,38 @@ def test_source_api_overlap(self):
...
+        test_data = pd.DataFrame({"val": test_val, "se": [np.nan] * len(test_val),
+                     "sample_size": [np.nan] * len(test_val), "geo_id": ["1"] * len(test_val),
+                     # datetime.date type
+                     "time_value": datetime.strptime("2020-10-26", "%Y-%m-%d").date()})
+
+        try:
+            # This should run without raising any errors.
+            validator.check_max_date_vs_reference(test_data, ref_data, "date", "state", "signal", report)
+        except TypeError:
+            assert False
```
krivard, line 502: suggestion: remove try-catch

pytest will fail if an unexpected exception is raised, so we don't need to explicitly check for exceptions.

```suggestion
        # This should run without raising any errors.
        validator.check_max_date_vs_reference(test_data, ref_data, "date", "state", "signal", report)
```

krivard (approved):
Can merge as is, but consider suggested changes for improved quality of life, especially if we ever have to come back to this test

M5Skid (approved):

PR #1838, opened 2023-04-19
Upgrade only to most recent working version of pip

korlaxxalrok:
A recent change in `pip-23.1` is causing our indicator build process to fail. This is a quick fix that upgrades only to the most recent working version of `pip-23.0.1`.

### Description
QUICK FIX: Only upgrade to `pip-23.0.1` in our initial build script.

### Changelog
- jenkins/build-and-package.sh


krivard (approved):
🪨 solid

PR #1828, opened 2023-04-11
[Backfill corrections] Pin `bettermc` to last cran-approved version

nmdefries:
### Description
The `bettermc` package, required both for backfill corrections and as a dependency of `evalcast`, was recently [removed from CRAN](https://cran.rstudio.com/web/packages/bettermc/index.html). Pin to the last working version.

### Changelog
- Dockerfile

PR #1848, opened 2023-05-02
Pin bettermc to last cran-approved version in workflow

nmdefries:
### Description
Make the same change as https://github.com/cmu-delphi/covidcast-indicators/pull/1828 in the backfill corrections workflow.

The bettermc package, required both for backfill corrections and as a dependency of evalcast, was recently [removed from CRAN](https://cran.rstudio.com/web/packages/bettermc/index.html). Pin to the last working version.

### Changelog
- `.github/workflows/backfill-corr-ci.yml`

### Fixes 
Backfill corrections CI is [failing](https://delphi-org.slack.com/archives/C013AH5N01E/p1682977732181189).

### DIFF
--- a/.github/workflows/backfill-corr-ci.yml
+++ b/.github/workflows/backfill-corr-ci.yml
@@ -29,7 +29,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Set up R ${{ matrix.r-version }}
-        uses: r-lib/actions/setup-r@v1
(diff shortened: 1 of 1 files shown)


nmdefries:
This works, but a 1-hr (uncached) runtime is annoying to work with. Will try to reduce that in a separate PR.

M5Skid (approved):
This looks good to me.

PR #1851, opened 2023-05-25
Revert "Request rcmdcheck to be installed with dependencies in backfill corrections CI"

minhkhul:
Reverts cmu-delphi/covidcast-indicators#1850 due to ci [failure](https://github.com/cmu-delphi/covidcast-indicators/actions/runs/5081193622/jobs/9129194459).


dshemetov (approved):
The CI is reporting a Python `pip` resolver error (same [message as this](https://github.com/pypa/pip/issues/10851)), so the actual issue is probably there. 

Meanwhile, we can try this revert fix:
1. If it works, then there's something about rcmdcheck that triggers some bad Python dependencies
2. If it doesn't, then the CI is failing due to bad Python dependencies

In either case, the next step will probably be to update pip.

PR #1852, opened 2023-05-26
Set backfill corrections pip to 22.0.3

minhkhul:
### Description
- Set backfill corrections pip to 22.0.3 to solve current CI [failed runs](https://github.com/cmu-delphi/covidcast-indicators/actions/runs/5086068679/jobs/9140173022).
- Restore Nat's change: #1850


dshemetov (approved):
🚀

PR #1862, opened 2023-06-22
Add API keys to indicators and sircal

neul3:
### Description
Adds API keys to Google symptoms, SIRcal, and the validator 

### Changelog
- Added api key call to each indicator 
- Added Ansible creds for each new key

### Fixes 
- Fixes sircal and indicator outages


changehc/delphi_changehc/run.py
```diff
@@ -9,6 +9,7 @@
 import time
 from datetime import datetime, timedelta
 from typing import Dict, Any
+import covidcast
```
krivard, line 12: confusion: why does changehc need an API key if it wasn't already hitting the API?

neul3: The original issue with the indicator is identical as far as I can tell to the progression of google_symptoms ((https://delphi-org.slack.com/archives/C01LZ3A2UMU/p1686834029955849)). It looks like while the indicator doesn't use it, somewhere down the line the indicator retrieves metadata (but you are right, I can't quite find where). 

Error message that led me to believe this:
 File "/mnt/data7/indicators/runtime/changehc/env/lib/python3.8/site-packages/covidcast/covidcast.py", line 299, in <lambda>
   meta_df["min_time"] = meta_df.apply(lambda x: _parse_datetimes(x.min_time, x.time_type), axis=1)
 File "/mnt/data7/indicators/runtime/changehc/env/lib/python3.8/site-packages/pandas/core/generic.py", line 5902, in __getattr__
   return object.__getattribute__(self, name)
AttributeError: 'Series' object has no attribute 'min_time'

I'm guessing that we will get the "TOO MANY REQUESTS" error on the chng pipeline tonight if it behaves the same as google_symptoms, which to my observations so far, it has. However there is always the very real possibility that I jumped to conclusions early and missed something.

krivard: Oh that's not the indicator itself, that's the validator. Validator only runs when an indicator is started using delphi_utils.runner, and that's not specific to chng (or google-symptoms)... it should probably get installed as part of [validator_from_params](https://github.com/cmu-delphi/covidcast-indicators/blob/main/_delphi_utils_python/delphi_utils/validator/run.py#L29) and not within each individual indicator's run method

changehc/delphi_changehc/run.py
```diff
@@ -119,6 +119,7 @@ def run_module(params: Dict[str, Dict[str, Any]]):
             - "types": list of str, sensor types to generate.
             - "wip_signal": list of str or bool, to be passed to delphi_utils.add_prefix.
             - "ftp_conn": dict, connection information for source FTP.
+            - "api_credentials": str, api key used
```
krivard, line 122: No longer needed

```suggestion
```

krivard (approved):
Minor cleanup then you're good to go!

PR #1787, opened 2023-02-21
Add `popsafe-fips` to geomapper

nmdefries:
### Description
Create new level in the geomapper to support changes to `chng` data. [Background and design considerations](https://docs.google.com/document/d/1JLGWiIwQLRKZo2gEbgppL6O6p-UXhQS1QH_Itekn_Uc/edit).

> CHNG would like to stop publishing county-level information for counties with low population (<20k). They have proposed a list of ~400 county groups which bundle together low-population counties until the population is above the 20k threshold. The groups are reasonably contiguous and combining data from low-population areas is a common technique to alleviate privacy concerns

To allow this, add a `popsafe-fips` level to the geomapper that codes high-population (>20k) counties with their official FIPS codes and low-population counties with a code of the format `<two digit state FIPS>g<two digit group number within the state>`, e.g. "01g05". Add mappings for `fips` -> `popsafe-fips` and `popsafe-fips` -> `state`.

### Changelog
- Original county groups file, `data_proc/geomap/lowpop_county_groups.csv`
- `geo_data_proc.py` - functions to generate `popsafe-fips` map tables from input county groups file
- 2019 and 2020 files mapping `fips` -> `popsafe-fips` and `popsafe-fips` -> `state`
- `geomap.py`
- `test_geomap.py`

Issue #1877, opened 2023-07-25
Update chng-county geomapping to use CHNG syntax


krivard:
CHNG county groupings are reported in the raw files using underscore-delimited FIPS instead of a unique identifier for each group:

```
[indicators@delphi-master-prod-01 changehc]$ zgrep _ cache/20230723_Counts_Products_Mixed.dat.gz |head
19491222,06049_06035,3 or less
20020120,08053_08111_08091_08033_08113_08085,3 or less
20091014,06049_06035,3 or less
20151023,32017_32003,3 or less
20151123,32017_32003,3 or less
20151223,32017_32003,3 or less
20160123,32017_32003,3 or less
20160223,32017_32003,3 or less
20160323,32017_32003,3 or less
20160423,32017_32003,3 or less
```

#1858 let us continue acquiring CHNG data by temporarily dropping chng county groupings from the pipeline, but we should update the geo parsing and mapping code to accept this underscore format.

nmdefries:
`chng-fips` geo code is already available in the geomapper

PR #1883, opened 2023-08-04
Support `chng-fips` county group reporting in `chng` for privacy

nmdefries:
### Description
Report signals for `chng-fips` county groups instead of counties. See the [geomapper support PR](https://github.com/cmu-delphi/covidcast-indicators/pull/1787) for more context.

Turn off sirCAL alerting for county versions of all signals using [this format](https://github.com/cmu-delphi/covidcast-indicators/pull/1233).

#### Big Note
From local testing, it looks like the new county groups are never actually reported... This doesn't seem to be a definitional issue, that is, theoretically county groups could be reported if they had enough total claims reported. But because the county groups are low-population areas in general, they tend not to have enough total claims reported to be included in the output, based on our filtering logic.

Do we want to do more investigation here? I started in on population comparisons (do counties of similar population to county groups also get filtered out? Some megacounties do, at least) but didn't do anything super formal or thorough.

Non-adjusted `county` and `chng-fips` values match exactly (although `county` files have an extra entry for a county called "\N"??); adjusted `county` and `chng-fips` values are very similar but not exactly the same.

### Changelog
- `update_sensor.py`
- Local and production template params for `chng` and sirCAL.

### DIFF
--- a/ansible/templates/changehc-params-prod.json.j2
+++ b/ansible/templates/changehc-params-prod.json.j2
@@ -23,7 +23,7 @@
     "n_waiting_days": 3,
     "se": false,
     "parallel": false,
-    "geos": ["state", "msa", "hrr", "county", "hhs", "nation"],
+    "geos": ["state", "msa", "hrr", "chng-fips", "hhs", "nation"],
     "weekday": [true, false],
     "types": ["covid","cli","flu"],
     "wip_signal": "",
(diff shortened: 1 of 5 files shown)


nmdefries:
@jingjtang This should work, but wanted to get your thoughts on the data note above. Seems like we're going to a lot of work to report data at this new county group level but it very very rarely actually gets reported (might get reported during an upswing in cases?).

nmdefries:
Need to incorporate [county groups format](https://github.com/cmu-delphi/covidcast-indicators/issues/1877).

PR #1881, opened 2023-08-03
Add multiprocessing to the Quidel indicator

rzats:
### Description
Parallelizes the long-running `generate_sensor_for_*_geo` functions in the Quidel indicator - see #1547

In a local test run, this change made the total runtime for the `pull` step of this indicator 4.2x faster, and the runtime of the `generate` functions 6x faster:
 
Indicator runtime before the parallelization

_Indicator runtime before the parallelization - note the `generate_sensor_for_parent_geo` call..._

Indicator runtime after the parallelization

_Indicator runtime after the parallelization - `generate` call replaced by multiprocessing functions._

### Changelog
- Adds multiprocessing to Quidel indicator.

PR #1885, opened 2023-08-08
bring loggers in sync and add multiproc capabilities

melange396:
#### this does a few things, including:

- bringing the `logger.py` files from https://github.com/cmu-delphi/delphi-epidata/ and https://github.com/cmu-delphi/covidcast-indicators/ back in sync.  (there will be equivalent PRs in both repositories.)

- larger functional differences are now applied to both:
  - the version in `covidcast-indicators` didnt log PIDs 
  - nor did it have the env var trick to get debug-level output
  - the version in `delphi-epidata` didnt always respect output filenames properly (though it seems we didnt actually trigger this in practice; look at the collapsible section below for how to see this for yourself).

- differentiated the uncaught exception handlers a bit
  - and reduced the log level for messages from threads exiting cleanly.  this is mostly because multiprocessing.Manager processes start and [stop threads](https://github.com/python/cpython/blob/3.8/Lib/multiprocessing/managers.py#L289-L291) often.

- added multiprocess-safe logging constructs, which should be helpful in https://github.com/cmu-delphi/covidcast-indicators/pull/1881 and in the future. (see collapsible section below for sample usage)

---

<details>
<summary>to reproduce bad outfile behavior:</summary>

- first, in your shell:

```
curl https://raw.githubusercontent.com/cmu-delphi/covidcast-indicators/main/_delphi_utils_python/delphi_utils/logger.py -o ci_logger.py

curl https://raw.githubusercontent.com/cmu-delphi/delphi-epidata/dev/src/common/logger.py -o de_logger.py
```

- then follow these instructions: toggle the commented-out-ness of the imports below, remove files foo.txt and bar.txt, paste this into your `python3` interpreter, check those files, and repeat:                       

```python
import de_logger as a_logger
#import ci_logger as a_logger

foo = a_logger.get_structured_logger('foo', filename='foo.txt')
bar = a_logger.get_structured_logger('bar', filename='bar.txt')
foo.info("hi")
bar.info("oh hello")
```

</details>

<details>
<summary>two example usages of multiprocessing-compatible logger:</summary>

- this approach uses `multiprocessing` directly, requires explicit termination of the LoggerThread (after ensuring the pool jobs have completed), and requires the `sublogger` to be instantiated before the worker and bound in an accessible scope (it must be used as though it is local to the worker, it cannot be passed as an argument to pool methods)

```python
import multiprocessing
from delphi.epidata.common.logger import get_structured_logger, LoggerThread

logger = get_structured_logger('sample_1')

lt = LoggerThread(logger)
sl = lt.get_sublogger()
sl.info("test")

def worker_1(message):
  # do something here
  sl.info(message)
  # do somethine else here

num_processes = 5
with multiprocessing.Pool(num_processes) as pool:
  pool.apply_async(worker_1, args=["abcd"])
  pool.apply_async(worker_1, args=["efgh"])
  pool.close()
  pool.join()

lt.stop()
```

- this approach uses the context manager to provide a pool and logger, and the logger can be used in arguments to pool methods.  this is a little bit less lightweight (the multiprocessing.Manager() it uses will spawn its own subprocess and a number of threads within it) but it is cleaner and more straightforward.

```python
from delphi.epidata.common.logger import get_structured_logger, pool_and_threadedlogger

logger = get_structured_logger('sample_2')

def worker_2(message, sublogger):
  # do something here
  sublogger.info(message)
  # do somethine else here

num_processes = 5  # optional argument passed to Pool constructor
with pool_and_threadedlogger(logger, num_processes) as (pool, sublogger):
  pool.apply_async(worker_2, args=["asdf", sublogger])
  pool.apply_async(worker_2, args=["jkl;", sublogger])
  pool.apply_async(worker_2, args=["qwerty", sublogger])
```

</details>

### DIFF
--- a/_delphi_utils_python/delphi_utils/logger.py
+++ b/_delphi_utils_python/delphi_utils/logger.py
@@ -76,9 +76,7 @@ def get_structured_logger(name=__name__,
         handlers=[logging.StreamHandler()])
 
     def add_pid(_logger, _method_name, event_dict):
-        """
-        Add current PID to the event dict.
-        """
+        """Add current PID to the event dict."""
         event_dict["pid"] = os.getpid()
         return event_dict
 
@@ -149,25 +147,29 @@ class LoggerThread():
 
     class SubLogger():
         """
-        multiprocessing-safe logger-like interface
-        to convey log messages to a listening LoggerThread
+        Multiprocessing-safe logger-like interface
+        to convey log messages to a listening LoggerThread.
         """
+
         def __init__(self, queue):
             self.queue = queue
+
         def _log(self, level, *args, **kwargs):
             kwargs_plus = {'sub_pid': multiprocessing.current_process().pid}
             kwargs_plus.update(kwargs)
             self.queue.put([level, args, kwargs_plus])
+
         def info(self, *args, **kwargs):
-            """log an INFO level message"""
+            """Log an INFO level message."""
             self._log(logging.INFO, *args, **kwargs)
+
         def warn(self, *args, **kwargs):
-            """log an WARN level message"""
+            """Log a WARN level message."""
             self._log(logging.WARN, *args, **kwargs)
 
 
     def get_sublogger(self):
-        """accessor method to retrieve a SubLogger for this LoggerThread"""
+        """Accessor method to retrieve a SubLogger for this LoggerThread."""
         return self.sublogger
 
     def __init__(self, logger, q=None):
@@ -203,7 +205,7 @@ class LoggerThread():
         self.running = True
 
     def stop(self):
-        """terminate this LoggerThread"""
+        """Terminate this LoggerThread."""
         if not self.running:
             self.logger.warn('thread already stopped')
             return
@@ -218,7 +220,7 @@ class LoggerThread():
 @contextlib.contextmanager
 def pool_and_threadedlogger(logger, *poolargs):
     """
-    emulates the multiprocessing.Pool() context manager,
+    Emulates the multiprocessing.Pool() context manager,
     but also provides a logger that can be used by pool workers.
     """
     with multiprocessing.Manager() as manager:
--- a/_delphi_utils_python/delphi_utils/logger.py
+++ b/_delphi_utils_python/delphi_utils/logger.py
@@ -1,21 +1,38 @@
-"""Structured logger utility for creating JSON logs in Delphi pipelines."""
+"""Structured logger utility for creating JSON logs."""
+
+# the Delphi group uses two ~identical versions of this file.
+# try to keep them in sync with edits, for sanity.
+#   https://github.com/cmu-delphi/covidcast-indicators/blob/main/_delphi_utils_python/delphi_utils/logger.py
+#   https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/common/logger.py
+
 import logging
+import os
+import structlog
 import sys
 import threading
-import structlog
+import traceback
 
 
 def handle_exceptions(logger):
     """Handle exceptions using the provided logger."""
-    def exception_handler(etype, value, traceback):
-        logger.exception("Top-level exception occurred",
-                         exc_info=(etype, value, traceback))
 
-    def multithread_exception_handler(args):
-        exception_handler(args.exc_type, args.exc_value, args.exc_traceback)
+    def exception_handler(scope, etype, value, traceback):
+        logger.exception("Top-level exception occurred", scope=scope, exc_info=(etype, value, traceback))
 
-    sys.excepthook = exception_handler
-    threading.excepthook = multithread_exception_handler
+    def sys_exception_handler(etype, value, traceback):
+        exception_handler("sys", etype, value, traceback)
+
+    def threading_exception_handler(args):
+        if args.exc_type == SystemExit and args.exc_value.code == 0:
+            # `sys.exit(0)` is considered "successful termination":
+            #   https://docs.python.org/3/library/sys.html#sys.exit
+            logger.debug(f"normal thread exit", thread=args.thread,
+                        stack="".join(traceback.format_exception(args.exc_type, args.exc_value, args.exc_traceback)))
+        else:
+            exception_handler(f"thread: {args.thread}", args.exc_type, args.exc_value, args.exc_traceback)
+
+    sys.excepthook = sys_exception_handler
+    threading.excepthook = threading_exception_handler
 
 
 def get_structured_logger(name=__name__,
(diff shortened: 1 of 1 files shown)


melange396:
paired PR: https://github.com/cmu-delphi/delphi-epidata/pull/1254

_delphi_utils_python/delphi_utils/logger.py
```diff
@@ -84,10 +120,108 @@ def get_structured_logger(name=__name__,
...
         system_logger.addHandler(logging.FileHandler(filename))
-    system_logger.setLevel(logging.INFO)
+    system_logger.setLevel(log_level)
     logger = structlog.wrap_logger(system_logger)
 
     if log_exceptions:
         handle_exceptions(logger)
 
     return logger
+
```
rzats, line 130: Does this need such a large gap?

_delphi_utils_python/delphi_utils/logger.py
```diff
@@ -84,10 +120,108 @@ def get_structured_logger(name=__name__,
...
+    def __init__(self, logger, q=None):
+        self.logger = logger
+        if q:
+            self.msg_queue = q
+        else:
+            self.msg_queue = multiprocessing.Queue()
+
+        def logger_thread_worker():
+            self.logger.info('thread started')
```
rzats, line 181: These and other thread-related messages seem like they should be `debug`-level or otherwise configurable, maybe through a boolean argument in `pool_and_threadedlogger`? For the Quidel indicator it would make sense to have them turned off by default.

_delphi_utils_python/delphi_utils/logger.py
```diff
@@ -84,10 +120,108 @@ def get_structured_logger(name=__name__,
...
+        # TODO: make this safely re-callable?
+        self.logger.info('sending stop signal')
+        self.msg_queue.put('STOP')
+        self.thread.join()
+        self.logger.info('thread stopped')
+
+
+
+@contextlib.contextmanager
+def pool_and_threadedlogger(logger, *poolargs):
```
rzats, line 214: Really elegant design! I like how this is built to be used in a `with` clause.

_delphi_utils_python/delphi_utils/logger.py
```diff
@@ -1,21 +1,44 @@
...
-    def multithread_exception_handler(args):
-        exception_handler(args.exc_type, args.exc_value, args.exc_traceback)
+    def threading_exception_handler(args):
+        if args.exc_type == SystemExit and args.exc_value.code == 0:
+            # `sys.exit(0)` is considered "successful termination":
+            #   https://docs.python.org/3/library/sys.html#sys.exit
+            logger.debug("normal thread exit", thread=args.thread,
+                        stack="".join(
+                            format_exception(
+                                args.exc_type, args.exc_value, args.exc_traceback)))
```
rzats, line 35: Lines like this seem like a good reason to loosen the line length requirement for pylint like you mentioned on Slack :)

rzats (commented):
Reviewed this PR rather than the epidata one since it has a little more commits. Looks good with a few small caveats!

_delphi_utils_python/delphi_utils/logger.py
```diff
@@ -84,10 +119,116 @@ def get_structured_logger(name=__name__,
...
+            while True:
+                msg = self.msg_queue.get()
+                if msg == 'STOP':
+                    logger.debug('received stop signal')
+                    break
+                level, args, kwargs = msg
+                if level == logging.INFO:
+                    logger.info(*args, **kwargs)
+                if level == logging.WARN:
+                    logger.warn(*args, **kwargs)
```
dshemetov, line 193: Do we need debug or exception (or others) here too? Why/why not?

_delphi_utils_python/delphi_utils/logger.py
```diff
@@ -84,10 +119,116 @@ def get_structured_logger(name=__name__,
...
+                else:
+                    logger.error('received unknown logging level!  exiting...')
+                    break
+            logger.debug('stopping thread')
+
+        self.thread = threading.Thread(target=logger_thread_worker,
+                                       name="LoggerThread__"+logger.name)
```
dshemetov, line 200: I don't quite follow the purpose of this thread... it seems like it should interact with self.sublogger in some way, no? Could you expand a bit?

dshemetov (commented):
Mostly have some questions and want to understand the multiprocessing stuff better

_delphi_utils_python/delphi_utils/logger.py
melange396: the thread doesnt really "interact" with the sublogger; the sublogger acts indirectly upon the thread.  the thread is spawned in the background with a handle to a queue and a handle to the provided Logger, and it just logs anything that shows up on the queue.  the sublogger is used by things outside the thread (or even outside the process!) to enqueue things to be logged.

_delphi_utils_python/delphi_utils/logger.py
melange396: we only need the methods we use, and right now, nobody is using debug or exception.  i had a TODO here to make it more obvious to anyone who came looking that they could add their own, but the current pylint settings disallowed that...

this is sort of a proxy or facade for a Logger, and it doesnt need to duplicate the whole interface.  someone using a SubLogger got there by providing an existing Logger object in the first place, so hopefully they have some idea that its still wrapped up in there.

_delphi_utils_python/delphi_utils/logger.py
```diff
@@ -84,10 +119,116 @@ def get_structured_logger(name=__name__,
...
+            logger.info('thread started')
+            while True:
+                msg = self.msg_queue.get()
+                if msg == 'STOP':
+                    logger.debug('received stop signal')
+                    break
+                level, args, kwargs = msg
+                if level == logging.INFO:
+                    logger.info(*args, **kwargs)
+                if level == logging.WARN:
```
rzats, line 192: `elif` I think. otherwise this will error out if `level == logging.INFO`

_delphi_utils_python/delphi_utils/logger.py
melange396: already taken care of!

_delphi_utils_python/delphi_utils/logger.py
melange396: ...i caved and added some other levels

_delphi_utils_python/delphi_utils/logger.py
melange396: yep!  but also i changed the design so this isnt needed anymore

_delphi_utils_python/delphi_utils/logger.py
```diff
@@ -84,10 +119,134 @@ def get_structured_logger(name=__name__,
...
+
+
+@contextlib.contextmanager
+def pool_and_threadedlogger(logger, *poolargs):
+    """
+    Provide (to a context) a multiprocessing Pool and a proxy to the supplied logger.
+
+    Emulates the multiprocessing.Pool() context manager,
+    but also provides (via a LoggerThread) a SubLogger proxy to logger
+    that can be safely used by pool workers.
```
dshemetov, line 240:
```suggestion
    that can be safely used by pool workers.
    The SubLogger proxy interface supports these methods: debug, info, warning, error,
    and critical.
```

dshemetov (approved):
One doc suggestion, but otherwise lgtm!

PR #1892, opened 2023-08-16
Add better logging in the indicator runner

### LINKED ISSUE #1890: Add better logging in the indicator runner
https://github.com/cmu-delphi/covidcast-indicators/blob/main/_delphi_utils_python/delphi_utils/runner.py could have some more detail in its logging.

- add timing for (ie, around) core indicator method, and validation and flash steps (archive step already has some minimal summary info but could probably be improved)
- more detail for individual validation failure log messages, like threshold and reference/live values
- maybe move log line that prints "Completed indicator run" to a core location instead of replicating in each indicator?

and less generally:
- log around update_cache_file() in delphi_quidel_covidtest because the cache file is getting huge in prod (~180MB as of 2 aug 2023)

rzats:
### Description
Adds more detail to the logging of the indicator runner, as specified in #1890.

### Changelog
- Adds timing logging around various steps of the indicator runner.
- For validation errors & warnings caused by specific thresholds not being met, specify the thresholds' values.

### Fixes 
- Fixes #1890

### DIFF
--- a/_delphi_utils_python/delphi_utils/runner.py
+++ b/_delphi_utils_python/delphi_utils/runner.py
@@ -83,7 +83,6 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
                 elapsed_time_in_seconds = elapsed_time_in_seconds)
         t1.terminate()
         t1.join()
-
     if validator:
         validation_report = validator.validate()
         validation_report.log(logger)
--- a/_delphi_utils_python/delphi_utils/runner.py
+++ b/_delphi_utils_python/delphi_utils/runner.py
@@ -67,6 +67,7 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
     validator = validator_fn(params)
     archiver = archiver_fn(params)
 
+    start_time = time.time()
     t1 = multiprocessing.Process(target=flash_fn, args=[params])
     t1.start()
     start = time.time()
@@ -77,6 +78,10 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
     else:
         t1.terminate()
         t1.join()
+    elapsed_time_in_seconds = round(time.time() - start_time, 2)
+    logger.info("Completed flash step",
+            elapsed_time_in_seconds = elapsed_time_in_seconds)
+
     if validator:
         validation_report = validator.validate()
         validation_report.log(logger)
--- a/_delphi_utils_python/delphi_utils/runner.py
+++ b/_delphi_utils_python/delphi_utils/runner.py
@@ -67,20 +67,22 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
     validator = validator_fn(params)
     archiver = archiver_fn(params)
 
-    start_time = time.time()
     t1 = multiprocessing.Process(target=flash_fn, args=[params])
     t1.start()
     start = time.time()
     while time.time()-start < timer:
         if not t1.is_alive():
+            elapsed_time_in_seconds = round(time.time() - start, 2)
+            logger.info("Completed flash step",
+                    elapsed_time_in_seconds = elapsed_time_in_seconds)
             break
         time.sleep(10)
     else:
+        elapsed_time_in_seconds = round(time.time() - start, 2)
+        logger.error(f"Flash step timed out ({timer} s), terminating",
+                elapsed_time_in_seconds = elapsed_time_in_seconds)
         t1.terminate()
         t1.join()
-    elapsed_time_in_seconds = round(time.time() - start_time, 2)
-    logger.info("Completed flash step",
-            elapsed_time_in_seconds = elapsed_time_in_seconds)
 
     if validator:
         validation_report = validator.validate()
--- a/_delphi_utils_python/delphi_utils/runner.py
+++ b/_delphi_utils_python/delphi_utils/runner.py
@@ -75,7 +75,7 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
             logger.info("Completed flash step",
                     elapsed_time_in_seconds = round(time.time() - start, 2))
             break
-        time.sleep(10)
+        time.sleep(1)
     else:
         logger.error(f"Flash step timed out ({timer} s), terminating",
                 elapsed_time_in_seconds = round(time.time() - start, 2))
--- a/_delphi_utils_python/delphi_utils/runner.py
+++ b/_delphi_utils_python/delphi_utils/runner.py
@@ -72,15 +72,13 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
     start = time.time()
     while time.time()-start < timer:
         if not t1.is_alive():
-            elapsed_time_in_seconds = round(time.time() - start, 2)
             logger.info("Completed flash step",
-                    elapsed_time_in_seconds = elapsed_time_in_seconds)
+                    elapsed_time_in_seconds = round(time.time() - start, 2))
             break
         time.sleep(10)
     else:
-        elapsed_time_in_seconds = round(time.time() - start, 2)
         logger.error(f"Flash step timed out ({timer} s), terminating",
-                elapsed_time_in_seconds = elapsed_time_in_seconds)
+                elapsed_time_in_seconds = round(time.time() - start, 2))
         t1.terminate()
         t1.join()
     if validator:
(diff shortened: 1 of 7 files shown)


_delphi_utils_python/delphi_utils/runner.py
```diff
@@ -67,6 +67,7 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
     validator = validator_fn(params)
```
rzats, line 67: The archiver already logs run timing, and each indicator has its own bit of logging which includes that; this adds logging for the flash & validation steps.

melange396: the individual indicators' run time logging looks mostly sufficient, except in `quidel_covidtest` where it happens [at program exit](https://github.com/cmu-delphi/covidcast-indicators/blob/57f5a15d9c6cb5ccbac9dd9f814045772e93bd19/quidel_covidtest/delphi_quidel_covidtest/run.py#L124), which will include anything that runs after the core indicator function and thus lead to inaccuracy.

you can refactor this so the runner does the timing and logging, with `indicator_fn` (aka each indicator's `run_module()`) returning a dict of metrics it wants logged (like csv_export_count and max_lag_in_days)

_delphi_utils_python/delphi_utils/runner.py
```diff
@@ -67,6 +67,7 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
...
     archiver = archiver_fn(params)
 
+    start_time = time.time()
```
melange396, line 70: this ~same value is already captured 3 lines below

_delphi_utils_python/delphi_utils/runner.py
```diff
@@ -77,6 +78,10 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
     else:
         t1.terminate()
         t1.join()
+    elapsed_time_in_seconds = round(time.time() - start_time, 2)
+    logger.info("Completed flash step",
```
melange396, line 82: how about making this conditional on normal flash completion, with a different error-level log message when the flash process is forcefully terminated?

melange396 (changes requested):
i havent looked at all of the individual validator messages yet, but i will in the next round.  can you also get timing for that other method mentioned in #1890 ?

_delphi_utils_python/delphi_utils/runner.py
rzats: Hm, I can actually think of one disadvantage of that - it means that the summary line will be gone from the logs if the indicator is ran individually (e.g. `env/bin/python -m delphi_quidel_covidtest` as the [README](https://github.com/cmu-delphi/covidcast-indicators/blob/main/quidel_covidtest/README.md) suggests).
Are we OK with this?

melange396: a lot of the documentation in this repo is is quite old and should be brought up to date.

is it ever desirable to run an indicator _*without*_ validation and archiving?  perhaps we can answer that in #1895.  you should at least fix the timing for the quidel indicator in the meanwhile.

rzats:
@melange396 updated, rerequesting review!

_delphi_utils_python/delphi_utils/validator/dynamic.py
```diff
@@ -237,7 +237,8 @@ def check_min_allowed_max_date(self, max_date, geo_type, signal_type, report):
                 ValidationFailure("check_min_max_date",
                                   geo_type=geo_type,
                                   signal=signal_type,
```
melange396, line 239: does it make sense to add this here too?
```suggestion
                                  geo_type=geo_type,
                                  signal=signal_type,
                                  date=max_date,
```

_delphi_utils_python/delphi_utils/validator/dynamic.py
```diff
@@ -263,7 +264,8 @@ def check_max_allowed_max_date(self, max_date, geo_type, signal_type, report):
                 ValidationFailure("check_max_max_date",
                                   geo_type=geo_type,
                                   signal=signal_type,
```
melange396, line 266: does it make sense to add this here too?
```suggestion
                                  geo_type=geo_type,
                                  signal=signal_type,
                                  date=max_date,
```

_delphi_utils_python/delphi_utils/validator/static.py
```diff
@@ -281,29 +281,35 @@ def check_bad_val(self, df_to_test, nameformat, signal_type, report):
 
         if percent_option:
             if not df_to_test[(df_to_test['val'] > 100)].empty:
+                bad_values = df_to_test[(df_to_test['val'] > 100)]['val'].unique()
```
melange396, line 284: nice touch with the `.unique()` filter!  should we do that for all the messages that print lists of values?

_delphi_utils_python/delphi_utils/validator/static.py
```diff
@@ -346,10 +352,12 @@ def check_bad_se(self, df_to_test, nameformat, report):
             report.increment_total_checks()
 
             if df_to_test["se"].isnull().mean() > 0.5:
+                bad_mean = round(df_to_test["se"].isnull().mean() * 100, 2)
                 report.add_raised_error(
                     ValidationFailure("check_se_many_missing",
                                       filename=nameformat,
-                                      message='Recent se values are >50% NA'))
+                                      message='Many recent se values are missing: '
+                                              f'{bad_mean} > 50%'))
```
melange396, line 360: keep the old message prefix:
```suggestion
                                      message=f'Recent se values are >50% NA: {bad_mean}'))
```

_delphi_utils_python/delphi_utils/validator/static.py
```diff
@@ -230,7 +230,7 @@ def find_all_unexpected_geo_ids(df_to_test, geo_regex, geo_type):
                         ValidationFailure(
                             "check_geo_id_type",
                             filename=nameformat,
-                            message="geo_ids saved as floats; strings preferred"))
+                            message=f"{len(leftover)} geo_ids saved as floats; strings preferred"))
```
melange396, line 233: keep the old message prefix, and take out the `len()` to just list the offending ids
```suggestion
                            message=f"geo_ids saved as floats; strings preferred: {leftover}"))
```

melange396 (changes requested):

_delphi_utils_python/delphi_utils/validator/static.py
```diff
@@ -178,17 +178,17 @@ def check_bad_geo_id_value(self, df_to_test, filename, geo_type, report):
             - report: ValidationReport; report where results are added
         """
         valid_geos = self._get_valid_geo_values(geo_type)
-        unexpected_geos = [geo for geo in df_to_test['geo_id']
-                           if geo.lower() not in valid_geos]
+        unexpected_geos = list(set([geo for geo in df_to_test['geo_id']
```
rzats, line 181: Made the "invalid values" list unique and in a couple other places (let me know if there's a better way to do this!)

_delphi_utils_python/delphi_utils/validator/static.py
```diff
@@ -178,17 +178,17 @@ def check_bad_geo_id_value(self, df_to_test, filename, geo_type, report):
             - report: ValidationReport; report where results are added
         """
         valid_geos = self._get_valid_geo_values(geo_type)
-        unexpected_geos = [geo for geo in df_to_test['geo_id']
-                           if geo.lower() not in valid_geos]
+        unexpected_geos = list({geo for geo in df_to_test['geo_id']
```
rzats, line 181: Made lists like this unique here and in a couple other places. lmk if there's a cleaner way to deduplicate these

melange396: the set generators in there are great!  you should be able to remove the `list()` wrapped around each of these sets as well; the only other places theyre used is in `len()` calls and f-string replacements, and those will work just as well with as set as they do with a list.

_delphi_utils_python/delphi_utils/runner.py
```diff
@@ -72,9 +72,13 @@ def run_indicator_pipeline(indicator_fn:  Callable[[Params], None],
     start = time.time()
     while time.time()-start < timer:
         if not t1.is_alive():
+            logger.info("Completed flash step",
+                    elapsed_time_in_seconds = round(time.time() - start, 2))
             break
         time.sleep(10)
```
melange396, line 78:
```suggestion
        time.sleep(1)
```
we should be able to bring this sleep time down without creating a busy wait, and itll give tighter timing measurements.

melange396 (approved):
good stuff!

PR #1899, opened 2023-09-22
Remove usage of PHP alias in the Python client

rzats:
Secondary PR to https://github.com/cmu-delphi/delphi-epidata/pull/1288 that should be merged & deployed alongside it - fixes breaking changes caused by the transition from "compatibility mode" / `api.php` in the Delphi Python client.

### DIFF
--- a/nowcast/delphi_nowcast/epidata.py
+++ b/nowcast/delphi_nowcast/epidata.py
@@ -38,8 +38,7 @@ def get_indicator_data(sensors: List[SensorConfig],
     all_combos = product(sensors, locations)
     as_of_str = as_of.strftime("%Y%m%d")
     all_params = [
-        {"source": "covidcast",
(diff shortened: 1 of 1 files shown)


melange396:
It looks like maybe you didnt get to run tests for this with the updated client from https://github.com/cmu-delphi/delphi-epidata/pull/1288 ?  Im pretty sure youll need to update the args to `assert_called_once_with()` for the `@patch`ed `async_epidata()`: https://github.com/cmu-delphi/covidcast-indicators/blob/3666fa65eac77e7db3b0bda4252140bbe91e8ed3/nowcast/tests/test_epidata.py#L37-L38

melange396:
since we changed https://github.com/cmu-delphi/delphi-epidata/pull/1288 to preserve the function signature of `async_epidata()`, shall we just abandon this PR?

PR #1908, opened 2023-11-09
Add API key to hhs run module

minhkhul:
### Description
There's an epidata covidhosp api call in hhs run module that got exposed to needing api key. The key is already available in params.json, so only need to add in code.

### Changelog
Add api key to hhs_run.

### DIFF
--- a/hhs_hosp/delphi_hhs/run.py
+++ b/hhs_hosp/delphi_hhs/run.py
@@ -80,7 +80,8 @@ def run_module(params):
             - "common":
                 - "api_credentials": str, api key to prevent hitting max number of query limit.
     """
-    Epidata.auth = ('epidata', str(params["validation"]["common"]["api_credentials"]))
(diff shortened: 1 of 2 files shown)


minhkhul:
adding tests.

hhs_hosp/delphi_hhs/run.py
```diff
@@ -77,6 +77,7 @@ def run_module(params):
             - "log_filename" (optional): str, name of file to write logs
             - "epidata" (optional): dict, extra parameters to send to Epidata.covid_hosp
     """
+    Epidata.auth = ('epidata', str(params["validation"]["common"]["api_credentials"]))
```
dshemetov, line 80: Should probably make this fail-safe, in case these keys don't exist
```suggestion
    if params.get("validation") and params["validation"].get("common") and params["validation"]["common"].get("api_credentials"):
        Epidata.auth = ('epidata', str(params["validation"]["common"]["api_credentials"]))
```

dshemetov (approved):
lgtm

PR #1912, opened 2023-11-22
Use national data for nchs-mortality signals

### LINKED ISSUE #1906: Synthesize "national" data for `nchs-mortality:deaths_covid_incidence_prop`
There are currently only "state" specific data points for NCHS Mortality series.  That is, only `geo_type=state` is provided for any of the data with `source=nchs-mortality`.  It would be advantageous for us to have this information aggregated to the national level.  We can (presumably) synthesize this by summing numbers from all states to produce a new "nation" data series (for each `reference` and `issue` date available).

In https://github.com/cmu-delphi/www-covidcast/issues/1233, on the [Covidcast Dashboard](https://delphi.cmu.edu/covidcast/), we intend to replace the old and no-longer-updated [JHU](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/jhu-csse.html) deaths signal (`jhu-csse:deaths_7dav_incidence_prop`) with one from [NCHS Mortality Data](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/nchs-mortality.html) (`nchs-mortality:deaths_covid_incidence_prop`).  The default view of the dashboard uses a geographic perspective of the United States as a whole.  Without an available "nation"-level signal for deaths,  it shows as "N/A" in plots and numeric text.

According to our signal documentation, the only currently active signals that do not provide "nation"-level data are from the sources `nchs-mortality` and `quidel`.  Do we want to construct national data for all of these?

rzats:
### Description
Currently, only state-specific data is used for any of the data with `source=nchs-mortality`. As the Covidcast dashboard uses the US as a whole by default, this leads to "N/A" values showing up in plots and numeric text.

This PR makes use of the previously discarded national data for all signals in the NCHS family, pulling it from the dataset rather than throwing it away.

### Changelog
- Alters the available geo resolutions for NCHS data from `state` to `['state', 'nation']`.
  - Excludes the `percent_of_expected_deaths` - despite the name, it contains _proportions_ of expected deaths (e.g. 1.1 instead of 110%), which are nontrivial to aggregate.
- Adds logic to process `nation` data from the dataset.
- Adds tests for whether data with the new resolutions is correctly created and exported to CSVs.

### Fixes 
- Fixes #1906


melange396:
It turns out that this dataset already has national-level data in it, we have just been throwing it away!  https://github.com/cmu-delphi/covidcast-indicators/blob/66b8ab221392915dda03f3ee10f02bc2e1b3c6e2/nchs_mortality/delphi_nchs_mortality/pull.py#L100

You can also see samples of this by browsing the dataset at:
https://data.cdc.gov/NCHS/Provisional-COVID-19-Death-Counts-by-Week-Ending-D/r8kw-7aab/data_preview

It is better to use this as-is from the source data than to rebuild it as an aggregation.

rzats:
@melange396 I've reworked the PR to use the data within the dataset, rather than synthesize new data. Of note:

- Removed one of the new tests, since the sum of state counts no longer adds up to the national count. This is due to some small cells being excluded from the data, as described in the test dataset:
```
One or more data cells have counts between 1–9 and have been suppressed in accordance with NCHS confidentiality standards.
```
- Re-included the `percent_of_expected_deaths` metric, as the data for it no longer needs to be nontrivially synthesized.

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -73,50 +73,60 @@ def run_module(params: Dict[str, Any]):
...
-            )
-            if len(dates) > 0:
-                stats.append((max(dates), len(dates)))
-        else:
-            for sensor in SENSORS:
+            for geo in ["state", "nation"]:
                 logger.info("Generating signal and exporting to CSV",
-                            metric = metric,
-                            sensor = sensor)
+                            metric = metric)
```
melange396, line 78: add `geo` to distinguish what will be otherwise identical repeated log messages
```suggestion
                            metric=metric, geo_level=geo)
```

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -73,50 +73,60 @@ def run_module(params: Dict[str, Any]):
...
                     weekly_dates=True
                 )
                 if len(dates) > 0:
                     stats.append((max(dates), len(dates)))
+        else:
+            for geo in ["state", "nation"]:
+                for sensor in SENSORS:
+                    logger.info("Generating signal and exporting to CSV",
+                                metric = metric,
+                                sensor=sensor)
```
melange396, line 105: add `geo` to distinguish what will be otherwise identical repeated log messages
```suggestion
                    logger.info("Generating signal and exporting to CSV",
                                metric=metric, sensor=sensor, geo_level=geo)
```

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -73,50 +73,60 @@ def run_module(params: Dict[str, Any]):
...
+                if geo == "nation":
+                    df = df[df["geo_id"] == "us"]
                 else:
-                    df["val"] = df[metric] / df["population"] * INCIDENCE_BASE
+                    df = df[df["geo_id"] != "us"]
+                df["val"] = df[metric]
                 df["se"] = np.nan
                 df["sample_size"] = np.nan
-                df = add_nancodes(df)
                 # df = df[~df["val"].isnull()]
```
melange396, line 87: add_nancodes() handles the nulls
```suggestion
```

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -73,50 +73,60 @@ def run_module(params: Dict[str, Any]):
...
+                        df = df[df["geo_id"] == "us"]
+                    else:
+                        df = df[df["geo_id"] != "us"]
+                    if sensor == "num":
+                        df["val"] = df[metric]
+                    else:
+                        df["val"] = df[metric] / df["population"] * INCIDENCE_BASE
+                    df["se"] = np.nan
+                    df["sample_size"] = np.nan
+                    # df = df[~df["val"].isnull()]
```
melange396, line 117: add_nancodes() handles the nulls
```suggestion
```

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -73,50 +73,60 @@ def run_module(params: Dict[str, Any]):
...
-                    df["val"] = df[metric] / df["population"] * INCIDENCE_BASE
+                    df = df[df["geo_id"] != "us"]
+                df["val"] = df[metric]
                 df["se"] = np.nan
                 df["sample_size"] = np.nan
-                df = add_nancodes(df)
                 # df = df[~df["val"].isnull()]
-                sensor_name = "_".join([SENSOR_NAME_MAP[metric], sensor])
+                df = add_nancodes(df)
+                sensor_name = "_".join([SENSOR_NAME_MAP[metric]])
```
melange396, line 89: join() does nothing here
```suggestion
```

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -73,50 +73,60 @@ def run_module(params: Dict[str, Any]):
...
                 dates = create_export_csv(
                     df,
-                    geo_res=GEO_RES,
+                    geo_res=geo,
                     export_dir=daily_export_dir,
                     start_date=datetime.strptime(export_start_date, "%Y-%m-%d"),
                     sensor=sensor_name,
```
melange396, line 95: to go with removal of join() above
```suggestion
                    sensor=SENSOR_NAME_MAP[metric],
```

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -73,50 +73,60 @@ def run_module(params: Dict[str, Any]):
...
-                export_dir=daily_export_dir,
-                start_date=datetime.strptime(export_start_date, "%Y-%m-%d"),
-                sensor=sensor_name,
-                weekly_dates=True
-            )
-            if len(dates) > 0:
-                stats.append((max(dates), len(dates)))
-        else:
-            for sensor in SENSORS:
+            for geo in ["state", "nation"]:
```
melange396, line 76: you should be able to pull this up two levels (outside of `for metric in METRICS:`) to reduce repetition

you could even do the filtering on `geo_id ==/!= "us"` there too, if you want to make another [sub]copy of `df_pull`

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -73,50 +73,60 @@ def run_module(params: Dict[str, Any]):
...
+                    geo_res=geo,
                     export_dir=daily_export_dir,
                     start_date=datetime.strptime(export_start_date, "%Y-%m-%d"),
                     sensor=sensor_name,
                     weekly_dates=True
                 )
                 if len(dates) > 0:
                     stats.append((max(dates), len(dates)))
+        else:
+            for geo in ["state", "nation"]:
```
melange396, line 101: you should be able to pull this up two levels (outside of `for metric in METRICS:`) to reduce repetition (as above)

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -73,50 +73,60 @@ def run_module(params: Dict[str, Any]):
...
                     df,
-                    geo_res=GEO_RES,
+                    geo_res=geo,
                     export_dir=daily_export_dir,
                     start_date=datetime.strptime(export_start_date, "%Y-%m-%d"),
                     sensor=sensor_name,
                     weekly_dates=True
                 )
                 if len(dates) > 0:
                     stats.append((max(dates), len(dates)))
```
melange396, line 99: you should be able to pull this up a few levels (outside of the `if metric == ... else`) to reduce repetition (removed here, indentation fixed below)
```suggestion
```

nchs_mortality/delphi_nchs_mortality/run.py
```diff
@@ -73,50 +73,60 @@ def run_module(params: Dict[str, Any]):
...
+                    dates = create_export_csv(
+                        df,
+                        geo_res=geo,
+                        export_dir=daily_export_dir,
+                        start_date=datetime.strptime(export_start_date, "%Y-%m-%d"),
+                        sensor=sensor_name,
+                        weekly_dates=True
+                    )
+                    if len(dates) > 0:
+                        stats.append((max(dates), len(dates)))
```
melange396, line 129: you should be able to pull this up a few levels (outside of the `if metric == ... else`) to reduce repetition (indentation fixed here, removed above)
```suggestion
        if len(dates) > 0:
            stats.append((max(dates), len(dates)))
```

melange396 (changes requested):
nice!  we should do a little cleanup on the inner loops of `run_module()`, and then i think it will be good to go.

melange396 (approved):
looks good!

melange396:
`nchs-mortality` national-level data went in last thursday for the first time...
- Good news: it added old dates (though not old issues, but thats probbly fine), and most signals look good to me.
- Less good news: values for "`*_prop`" national signals are all null.
  - Unfortunately, the one we were interested in for in the dashboard is a prop signal (from #1906), and it looks like zeroes in the [plots](https://delphi.cmu.edu/covidcast/indicator/?sensor=nchs-mortality-deaths_covid_incidence_prop&date=20200205).

Did this work for you in your local testing?  Since these values are coming out as "NULL", it makes me think the denominator [here](https://github.com/cmu-delphi/covidcast-indicators/blob/012bf46faca53cdcc5e0d4e293a97c36fc80284f/nchs_mortality/delphi_nchs_mortality/run.py#L108) is 0 or null, which then makes me think that the national population is not getting set properly [here](https://github.com/cmu-delphi/covidcast-indicators/blob/012bf46faca53cdcc5e0d4e293a97c36fc80284f/nchs_mortality/delphi_nchs_mortality/pull.py#L125-L127).

Also, to make [this message](https://github.com/cmu-delphi/www-covidcast/blob/8255f7a30126d3a37634dff2e8914095e524071c/src/blocks/IndicatorWarning.svelte#L33) go away, we need to mark the nchs-mortality signals as having national-level data in the spreadsheet and then get it transferred to the [csv](https://github.com/cmu-delphi/delphi-epidata/blob/8a476fd08f5aef6913fb3d231f1fcceb691953be/src/server/endpoints/covidcast_utils/db_signals.csv#L413).  We can take care of that after the signals are properly acquired.

PR #1919, opened 2023-12-15
Make SirCal send alert about nchs lag only after 19 days

minhkhul:
### Description
https://cmu-delphi.atlassian.net/browse/OKRF23-81

### DIFF
--- a/ansible/templates/sir_complainsalot-params-prod.json.j2
+++ b/ansible/templates/sir_complainsalot-params-prod.json.j2
@@ -43,7 +43,7 @@
       ]
     },
     "nchs-mortality": {
-      "max_age":16,
+      "max_age":19,
       "maintainers": []
     },
     "hhs": {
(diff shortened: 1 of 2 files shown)


nmdefries (approved):
Please update the [local/test version of the params](https://github.com/cmu-delphi/covidcast-indicators/blob/6c4c5b98be5e7b1c25293335162cd80b3e9e21f1/sir_complainsalot/params.json.template#L46), too. With that change, this looks good.

The change for `chng` will be in a different PR?

melange396 (commented):
Yeah, like @nmdefries said, be sure to change the other file to match.  This file is named "template" -- does that mean we have to do anything specialy when deploying to get it to apply?

minhkhul:
@nmdefries The chng lag is fixed now by moving sircal to running 30 minutes later than usual to give time for other processes to finish up. So no need for sircal changes.

PR #1926, opened 2024-01-10
Update archive differ S3 bucket

korlaxxalrok:
### Description
As part of an AWS migration we need to also update the archive differ bucket that certain indicators use. There are new access keys and a bucket name to supply to various templates via Ansible vars/vault tooling.

This will require some validation on staging before going to production!

### Changelog
- Update vars/vault with new assets
  - Adds new bucket name and vault references to vars
  - Adds the bucket name to vars for easier all-at-once updating of the templates that reference it
  - Adds new IAM access keys for bucket access to vault
  - Slight change to line spacing in vars to help with reading clarity
- Update various templates that use the archive differ
  - Supply templates with new variable references for bucket creds and bucket name


melange396 (approved):
im going to assume the correct things are happening in the encrypted `vault.yaml` 😹 

will there need to be a manual move of the contents of (ie, the files in) the current archive differ bucket to the new one?

korlaxxalrok:
> will there need to be a manual move of the contents of (ie, the files in) the current archive differ bucket to the new one?

Yes! We will need some coordination to stop some operations, deploy, sync the bucket, and then resume.

PR #1924, opened 2024-01-09
NWSS: post-staging fixes/updates

dsweber2:
The first commit is some changes that George recommended after the original PR was merged. This will also be a branch for any changes that need to be made to handle problems found during staging.

PR #1929, opened 2024-01-11
Additional configs after test runs on staging

minhkhul:
### Description
Some changes after test run nwss on staging.

### Changelog
- Add nchs token
- Change export dir in params template to match location with where other indicators receiving live


dsweber2:
the token is in #1924; do we want to add this to that and then merge, or wait on that and just merge this?

dsweber2 (approved):

minhkhul:
> the token is in #1924; do we want to add this to that and then merge, or wait on that and just merge this?

I'm gonna merge into nwss-staging.

PR #1934, opened 2024-01-18
Add Ansible Vault backup scripts

korlaxxalrok:
### Description
Enables better QOL for working with the Ansible vault file.

Adds:
- Script to create decrypted file for easy editing
- Script to encrypt
- Takes backups during each step
- Ignores sensitive files as appropriate


minhkhul (changes requested):
Could you change this current instruction in the decrypted vault_plain to use the scripts instead (and move instruction to the scripts because right now you can only see decrypt instruction after you have already decrypted the vault). 
image

ansible/vault-decrypt.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# Create backup directory if it doesn't exist.
+mkdir -p vault_backup
+
```
minhkhul, line 5: Since the scripts only work if you're in covidcast-indicators/ansible.
```suggestion
# Check if the current directory is not ansible
if [[ $(basename "$PWD") != "ansible" ]]; then
    # Check if ansible directory exists in the current directory
    if [[ -d "ansible" ]]; then
        cd ansible
    else
        echo "cd to covidcast-indicators/ansible to execute this script"
        exit 1
    fi
fi
```

korlaxxalrok: I think it maybe makes sense to remove the instructions from the file entirely since the scripts should be used going forward, and then add some small documentation somewhere on using the scripts.

I'm not totally against adding logic to the scripts to handle someone trying to run the script from the wrong location, but maybe it'll be simpler to add (as mentioned ☝️) some small documentation somewhere and explicitly say "you need to be in this place to use these scripts". Some small documentation for this process is probably warranted anyway.

minhkhul (approved):
So I found the current doc for vault in this repo here. I'll just update it w the new script use. https://docs.google.com/document/d/1Bbuvtoxowt7x2_8USx_JY-yTo-Av3oAFlhyG-vXGG-c/edit

ansible/vault-decrypt.sh
korlaxxalrok: @minhkhul

- I removed unnecessary instruction from vault_plain.yaml
- Added a small README to the ansible directory with basic usage of the scripts

melange396: if youre in the wrong directory, `vault.yaml` and/or `vault_plain.yaml` wont be there and the operation will fail anyway...

a couple suggestions, neither of which is truly necessary: (1) you could move the directory creation inside the `if` so it only happens when theres actually a file to back up, and (2) you could add an `echo` statement in there too to remind the user that the backup is happening.

korlaxxalrok:
@melange396 @minhkhul 

Made some changes, so now:

- There is some guarding against creating unnecessary `vault_backup` directories if you try to run the script from the wrong location, otherwise will fail naturally due to no vault/vault_plain files being found.
- If vault_plain doesn't exist then decrypting will proceed and create the file, but not a backup dir or backup file.
  - If run a second time then the backup dir and file are created. This will be a typical first run experience.
- If vault doesn't exist then encrypting will proceed and create the file, but not a backup dir or backup file
  - If run a second time then the backup dir and file are created. It would be atypical to not have a vault file to start, but maybe it got deleted for some reason.
 
I think this covers the basic needs. There is a README in the `ansible` directory now that has some additional info as well.

melange396 (approved):
noice, lgtm!

PR #1942, opened 2024-02-01
Add test for NCHS national & proportional data

rzats:
### Description
Adds an extra test for #1912 + #1938, making sure that national data is generated, and prop(ortional) signals are correctly calculated.

### DIFF
--- a/nchs_mortality/tests/test_run.py
+++ b/nchs_mortality/tests/test_run.py
@@ -5,7 +5,6 @@ import pytest
 
 import pandas as pd
 
-from delphi_utils.geomap import GeoMapper
 
 class TestRun:
     # the 14th was a Monday
--- a/nchs_mortality/tests/test_run.py
+++ b/nchs_mortality/tests/test_run.py
@@ -5,6 +5,7 @@ import pytest
 
 import pandas as pd
 
+from delphi_utils.geomap import GeoMapper
 
 class TestRun:
     # the 14th was a Monday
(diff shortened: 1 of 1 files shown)


nchs_mortality/tests/test_run.py
```diff
@@ -69,3 +70,27 @@ def test_output_file_format(self, run_as_module, date):
...
+        for output_folder in folders:
+            num = pd.read_csv(
+                join(output_folder, f"weekly_202026_nation_deaths_covid_incidence_num.csv")
+            )
+            prop = pd.read_csv(
+                join(output_folder, f"weekly_202026_nation_deaths_covid_incidence_prop.csv")
+            )
+            gmpr = GeoMapper()
+            national_pop = gmpr.get_crosswalk("nation", "pop")
+            us_pop = national_pop.loc[national_pop["nation"] == "us"]["pop"][0]
```
melange396, line 92: This is just repeating the same logic added in #1938 ...  I think its better to hardcode the US population value here in case GeoMapper starts behaving differently.

nchs_mortality/tests/test_run.py
```diff
@@ -69,3 +70,27 @@ def test_output_file_format(self, run_as_module, date):
...
+            # "assert almost equal" due to rounding down'
+            prop_value = pytest.approx(prop.iloc[0]["val"], 0.0001)
+            INCIDENCE_BASE = 100000
+            assert(prop_value == num.iloc[0]["val"] / us_pop * INCIDENCE_BASE)
```
melange396, line 96: Is this `approx` comparison necessary??  These calculations should be deterministic since the test and reference data were generated on the same machine using virtually (if not exactly) the same operations.

melange396 (changes requested):

nchs_mortality/tests/test_run.py
rzats: [They're rounded down during the export, though.](https://github.com/cmu-delphi/covidcast-indicators/blob/main/_delphi_utils_python/delphi_utils/export.py#L128)

nchs_mortality/tests/test_run.py
rzats: Will do!

nchs_mortality/tests/test_run.py
melange396: well then, it might not be a bad idea to hardcode this too

melange396 (approved):
looks good!

PR #1943, opened 2024-02-06
Patching scripts

minhkhul:
### Description
Patching scripts. Currently have available for hhs, change and quidel.


minhkhul:
OLD!

PR #1947, opened 2024-02-29
leap year fixes for claims_hosp and quidel_covidtest

melange396:
fixes for handling leap year in the backfill methods of `claims_hosp` and `quidel_covidtest`

shamlessly stolen and copypasta'd from [a similar method in `changehc`](https://github.com/cmu-delphi/covidcast-indicators/blob/a99327b4cc1ab070aea032d3d8812a32a19a1e03/changehc/delphi_changehc/backfill.py#L44-L47).


korlaxxalrok (approved):
Fire.It.Up. :shipit:

PR #1950, opened 2024-03-20
Nssp staging

minhkhul:
### Description
These are configurations to help nssp run in staging, not nssp indicator itself. This should not be merged until nssp indicator is already merged.

### Changelog
- add ansible config for params.json used in staging
- add nssp to indicator list in jenkinsfile.
- apart from changes here, also created `nssp` dir in `/common/covidcast/receiving/` on staging for .csv files to live in along with other indicators.

### DIFF
--- a/ansible/templates/nssp-params-prod.json.j2
+++ b/ansible/templates/nssp-params-prod.json.j2
@@ -1,0 +1,13 @@
+{
+  "common": {
+    "export_dir": "/common/covidcast/receiving/nssp",
+    "log_filename": "/var/log/indicators/nssp.log",
+    "log_exceptions": false
+  },
+  "indicator": {
+    "wip_signal": true,
+    "export_start_date": "2020-02-01",
+    "static_file_dir": "./static",
+    "socrata_token": "{{ nwss_wastewater_token }}"
+  }
+}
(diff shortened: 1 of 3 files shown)


ansible/templates/nssp-params-prod.json.j2
```diff
@@ -0,0 +1,13 @@
...
+  "common": {
+    "export_dir": "/common/covidcast/receiving/nssp",
+    "log_filename": "/var/log/indicators/nssp.log",
+    "log_exceptions": false
+  },
+  "indicator": {
+    "wip_signal": true,
+    "export_start_date": "2020-02-01",
+    "static_file_dir": "./static",
+    "socrata_token": "{{ nwss_wastewater_token }}"
```
dsweber2, line 11: We may want to rename `nwss_wastewater_token` if we reuse it elsewhere.

dsweber2 (approved):
The name for the token gets a bit awkward, but its not really a big deal

ansible/templates/nssp-params-prod.json.j2
minhkhul: Wait, I want to confirm that nssp uses the same cdc socrata token like with nchs and nwss correct?

PR #1951, opened 2024-03-28
Shut sircal up about historical chng signals

minhkhul:
### Description
[We're adding two historical signals to chng](https://cmu-delphi.atlassian.net/browse/OKRS24-105). For now they are not getting updates in the future. This PR makes sircal not check for new data in these signals.

### DIFF
--- a/ansible/templates/sir_complainsalot-params-prod.json.j2
+++ b/ansible/templates/sir_complainsalot-params-prod.json.j2
@@ -14,7 +14,8 @@
     },
     "chng": {
       "max_age": 6,
-      "maintainers": ["U01AP8GSWG3","U01069KCRS7"]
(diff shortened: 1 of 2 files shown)


melange396 (approved):
this looks like it should do the trick

a better way of doing it would be to check against the `active` flag in the signal metadata [here](https://github.com/cmu-delphi/covidcast-indicators/blob/089bd24f682725e3c8058512fafad7747187cf8b/sir_complainsalot/delphi_sir_complainsalot/check_source.py#L60) instead of against the enumerated list of retired signals...  implementing that shouldnt be too hard in itself, but we must make sure that all of the entries in our "retired-signals" lists are marked not active, and (more importantly!) that all the other signals for those sources ARE marked active (or we will disable the checks for them!).  i will make a gh issue for this tomorrow.

Issue #1901, opened 2023-10-03
Wastewater: NWSS SARS-Cov2 percentile dataset


willtownes:
# Publicly available signals from the National Wastewater Surveillance System (NWSS)

## Data details
 

There are basically three signals from two datasets. Both datasets are served via a Socrata API. 

In the [metrics dataset](https://data.cdc.gov/Public-Health-Surveillance/NWSS-Public-SARS-CoV-2-Wastewater-Metric-Data/2ew6-ywp6) we have the **percentile** signal which is the highest priority of the three along with the less useful **ptc_15d** signal. [Varkila et al 2023](https://jamanetwork.com/article.aspx?doi=10.1001/jamanetworkopen.2023.25591) is a useful reference showing how these two signals might be used. There are also some other possible signals pertaining to whether gene copies were detected or not. These data are normalized by flow rates and/or population (not fecal biomarkers like PMMoV), and the percentile signal is also normalized by its maximum value within each site. 

In the [concentration dataset](https://data.cdc.gov/Public-Health-Surveillance/NWSS-Public-SARS-CoV-2-Concentration-in-Wastewater/g653-rqe2) we have a smoothed, normalized metric called **pcr_conc_smoothed** . The normalization is not the same across sites, some of the values are normalized using human fecal biomarkers (usually PMMoV)

Geographic resolution: roughly county level (it is really at the wastewater treatment plant catchment basin level which may be more or less than the size of a county). Temporal resolution: intermittent, average of 1-3 observations per week. I don't think any processing is required (perhaps later we can think about mapping the numbers to the county level by dividing by population sizes etc). Note that they drop many of the locations that have small population sizes for privacy reasons. I recommend including all of the columns from both datasets. I don't anticipate there will be substantial backfill effects but it might be worth tracking it to find out.

willtownes:
@nmdefries ready for your review

dsweber2:
Looking at the datasets, its looking like there may be some gnarly geo_coding issues, at least with the metrics dataset. It looks like sample sites have situations they're tied to multiple fips (at most 2-3), and some fips having multiple `wwtp_id`'s (anywhere from 1-30). And in addition some of the sites where they're sampling upstream of the wastewater treatment plant have multiple sampling locations tied to them.
Maybe we can use the key_plot_id column we can sidestep this, but I suspect making signals at the county level will be messy, and state level too.
I'm planning on setting up a tracker locally to see if there's much revision going on.

[...]

Issue #991, opened 2021-04-07
USAFacts and JHU disagree about RI cases on March 28


krivard:
**Actual Behavior:**

The frontend shows a massive spike in RI on the 28th that's not reflected in the data from JHU. The raw USAFacts file only shows the spike in the statewide unallocated record:
```
$ cut -d, -f 1,2,3,435,436,437 usafacts.csv |sed -n '1p; /RI/p'
countyFIPS,County Name,State,2021-03-27,2021-03-28,2021-03-29
0,"Statewide Unallocated","RI",5120,12909,13553
44001,"Bristol County ","RI",4492,4492,4492
44003,"Kent County ","RI",16330,16330,16330
44005,"Newport County ","RI",5625,5625,5625
44007,"Providence County ","RI",87520,87520,87520
44009,"Washington County ","RI",8899,8899,8899
```

I haven't been able to find a news report on any case audits or otherwise large-scale discovery of previously-unreported deaths, including the JHU retrospective and back-distribution records.

We should:
* verify that this spike isn't in the JHU raw data (in case they changed their unallocated reporting and/or we introduced a bug in our code)
* poke USAFacts to fix their data

```

dfu = covidcast.signal("usa-facts","confirmed_incidence_num",geo_type="state",geo_values="ri",
start_day=date(2021,3,25),end_day=date(2021,3,31))
dfj = covidcast.signal("jhu-csse","confirmed_incidence_num",geo_type="state",geo_values="ri",
start_day=date(2021,3,25),end_day=date(2021,3,31))
pd.concat([
    dfu[["time_value","value"]].rename({"value":"usafacts"}, axis="columns").set_index("time_value"), 
    dfj[["time_value","value"]].rename({"value":"jhu"}, axis="columns").set_index("time_value")
], axis="columns").plot(y=["usafacts","jhu"])
```
![image](https://user-images.githubusercontent.com/1158666/113940268-30558b80-97cb-11eb-8cf3-abe3a32dc22f.png)

chinandrew:
This has been resolved in the latest USAfacts issues

```
[...]

PR #1952, opened 2024-04-17
nssp pipeline code

minhkhul:
### Description
Add 8 signals from source nssp


nssp/delphi_nssp/run.py
```diff
@@ -0,0 +1,133 @@
...
+        - "test_file" (optional): str, name of file from which to read test data
+        - "socrata_token": str, authentication for upstream data pull
+    - "archive" (optional): if provided, output will be archived with S3
+        - "aws_credentials": Dict[str, str], AWS login credentials (see S3 documentation)
+        - "bucket_name: str, name of S3 bucket to read/write
+        - "cache_dir": str, directory of locally cached data
+"""
+import time
+from datetime import datetime
+import pdb
```
dsweber2, line 26:
```suggestion
```

dsweber2 (approved):
Ran the pipeline and it seems to be pulling correctly, and the fields that are generated make sense. Generally looks good, I have some minor cosmetic suggestions. 

I guess the archive differ is the only part that really remains?

nssp/README.md
```diff
@@ -0,0 +1,75 @@
+# NWSS wastewater data
+
+We import the wastewater data, currently only the smoothed concentration, from the CDC website, aggregate to the state and national level from the wastewater sample site level, and export the aggregated data.
```
nmdefries, line 3: issue: Should reference NSSP, not WW data. Details of where data is coming from should also be checked.

nssp/delphi_nssp/__main__.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+"""Call the function run_module when executed.
+
+This file indicates that calling the module (`python -m delphi_NSSP`) will
```
nmdefries, line 4:
```suggestion
This file indicates that calling the module (`python -m delphi_nssp`) will
```

nssp/delphi_nssp/constants.py
```diff
@@ -0,0 +1,29 @@
...
+GEOS = [
+    "nation",
+    "state",
+    "county",
+]
+
+METRICS = ['percent_visits_covid','percent_visits_influenza',
+           'percent_visits_rsv','percent_visits_combined',
+           'percent_visits_smoothed_covid','percent_visits_smoothed_influenza',
+           'percent_visits_smoothed_rsv','percent_visits_smoothed_combined']
```
nmdefries, line 12: question: what's the difference between these smoothed signals here ("percent_visits_smoothed_covid") vs below ("smoothed_percent_visits_covid")? Should the names match?

nssp/delphi_nssp/constants.py
```diff
@@ -0,0 +1,29 @@
+"""Registry for variations."""
+
+GEOS = [
+    "nation",
+    "state",
+    "county",
```
nmdefries, line 6: suggestion: since we have county data available, there are various other geos we can easily aggregate up to (both MSAs and HRR are constructed along county boundaries). We should default to providing these for all new signals, unless there's a strong reason not to (e.g. values aren't comparable between geos). Is this handled elsewhere in the code?

nssp/delphi_nssp/pull.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+"""Functions for pulling NCHS mortality data API."""
```
nmdefries, line 2:
```suggestion
"""Functions for pulling NSSP ER data."""
```

nssp/delphi_nssp/pull.py
```diff
@@ -0,0 +1,83 @@
...
+    return f"""
+Expected column(s) missed, The dataset schema may
+have changed. Please investigate and amend the code.
+
+Columns needed:
+{NEWLINE.join(sorted(type_dict.keys()))}
+
+Columns available:
+{NEWLINE.join(sorted(df.columns))}
+"""
```
nmdefries, line 36: nit: indent for readability (unless that messes up f-strings?)

nssp/delphi_nssp/pull.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def pull_nssp_data(socrata_token: str):
+    """Pull the latest NWSS Wastewater data, and conforms it into a dataset.
```
nmdefries, line 40: nit: ditto about NWSS -> NSSP. Please check the rest of the code for this as well.

nssp/delphi_nssp/pull.py
```diff
@@ -0,0 +1,83 @@
...
+import pandas as pd
+from sodapy import Socrata
+
+from .constants import (
+    METRICS,
+    NEWLINE,
+)
+
+
+def construct_typedicts():
```
nmdefries, line 14: suggestion (optional): since this doesn't take any args, the output will be static. I'd probably move the dictionary def into constants.py instead of regenerating every time on line 59.

nssp/delphi_nssp/pull.py
```diff
@@ -0,0 +1,83 @@
...
+    client = Socrata("data.cdc.gov", socrata_token)
+    results = []
+    offset = 0
+    limit = 50000  # maximum limit allowed by SODA 2.0
+    while True:
+        page = client.get("rdmq-nq56", limit=limit, offset=offset)
+        if not page:
+            break  # exit the loop if no more results
+        results.extend(page)
+        offset += limit
```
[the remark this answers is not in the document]
nmdefries: @dsweber2 why did you choose 100 here for the limit in your rewrite? I believe 5k is the per-page item limit, not the total limit. So theoretically we could get an infinite-page result.

nssp/delphi_nssp/pull.py
```diff
@@ -0,0 +1,83 @@
...
+    df_ervisits = pd.DataFrame.from_records(results)
+    df_ervisits = df_ervisits.rename(columns={"week_end": "timestamp", 
+                                              "percent_visits_smoothed":"percent_visits_smoothed_combined",
+                                              "percent_visits_smoothed_1":"percent_visits_smoothed_influenza",})
```
nmdefries, line 75: praise: Appreciate some descriptive column names!

nssp/delphi_nssp/run.py
```diff
@@ -0,0 +1,133 @@
+# -*- coding: utf-8 -*-
+"""Functions to call when running the function.
+
+This module should contain a function called `run_module`, that is executed
+when the module is run with `python -m MODULE_NAME`.  `run_module`'s lone argument should be a
```
nmdefries, line 5:
```suggestion
when the module is run with `python -m delphi_nssp`.  `run_module`'s lone argument should be a
```

nssp/params.json.template
```diff
@@ -0,0 +1,13 @@
...
+    "export_dir": "./receiving",
+    "log_filename": "./nssp.log",
+    "log_exceptions": false
+  },
+  "indicator": {
+    "wip_signal": true,
+    "export_start_date": "",
+    "static_file_dir": "./static",
+    "socrata_token": ""
+  }
```
nmdefries, line 12: issue: Params needs a [`validation` section](https://github.com/cmu-delphi/covidcast-indicators/blob/414638c9c284c76d5416ad8388ae6fb30e69583a/changehc/params.json.template#L38-L47) as well

nssp/params.json.template
```diff
@@ -0,0 +1,13 @@
+{
+  "common": {
```
nmdefries, line 2: issue: We will also need a template params in [the `ansible/templates` directory](https://github.com/cmu-delphi/covidcast-indicators/tree/414638c9c284c76d5416ad8388ae6fb30e69583a/ansible/templates).

nssp/delphi_nssp/run.py
```diff
@@ -0,0 +1,133 @@
...
+    #     )
+    #     daily_arch_diff.update_cache()
+
+    run_stats = []
+    ## build the base version of the signal at the most detailed geo level you can get.
+    ## compute stuff here or farm out to another function or file
+    df_pull = pull_nssp_data(socrata_token)
+    sensor_i = 0
+    ## aggregate
+    for metric in METRICS:
```
nmdefries, line 104: suggestion: for clarity/maintainability, I'd recommend using `enumerate(METRICS)` instead of having a separate `sensor_i` index.

nssp/delphi_nssp/constants.py
nmdefries: suggestion: if we really do intend to rename "percent_visits_smoothed_covid" -> "smoothed_percent_visits_covid" at some point, please use a dictionary to keep the mapping straight. Relying on these two lists to have the same order will make it easy to introduce bugs, especially if these signal lists get longer in the future.

nmdefries (changes requested):
This file seems to be unused: [nssp/tests/test_data/page.txt](https://github.com/cmu-delphi/covidcast-indicators/pull/1952/files#diff-3024b1ba42e2e89d7b1d03526d691636e9dd6c269ef1a69601c8a26d35a66613)

Some nits and style suggestions. Question about which geos we're reporting.

nssp/delphi_nssp/pull.py
dshemetov: see: https://github.com/cmu-delphi/covidcast-indicators/pull/1954/files#diff-9eaea874c798327a0c3a60e5a9ac574bd5cb995d8b565fa01f26a2b3412a8743R29

nssp/params.json.template
minhkhul: yep. That one is [here](https://github.com/cmu-delphi/covidcast-indicators/pull/1950)

nssp/README.md
minhkhul: https://github.com/cmu-delphi/covidcast-indicators/pull/1952/commits/39e2cbdec58d38be3d6b0321e40fb8a4bca58932

nssp/delphi_nssp/constants.py
minhkhul: fixed. https://github.com/cmu-delphi/covidcast-indicators/pull/1952/commits/95bdac15983b7abaefee95e5093a374ecac4ee94

nssp/delphi_nssp/pull.py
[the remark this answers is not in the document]
dsweber2: 100 was maybe too low, though it would correspond to 50,000,000 items. If we're pulling more than that it should be quite a while down the road, or something has gone wrong. (or I may be misunderstanding how item counts work).

nssp/params.json.template
nmdefries: Ansible params will also need a validation section. And we'll want to add `nssp` to the sirCAL template and ansible params to make sure we're generating alerts for it.

nssp/delphi_nssp/pull.py
```diff
@@ -70,14 +71,14 @@ def pull_nssp_data(socrata_token: str):
         results.extend(page)
         offset += limit
     df_ervisits = pd.DataFrame.from_records(results)
-    df_ervisits = df_ervisits.rename(columns={"week_end": "timestamp", 
-                                              "percent_visits_smoothed":"percent_visits_smoothed_combined",
-                                              "percent_visits_smoothed_1":"percent_visits_smoothed_influenza",})
+    print(df_ervisits.columns)
```
dsweber2, line 74: guessing this is a debugging print statement?

minhkhul: yup. removed. https://github.com/cmu-delphi/covidcast-indicators/pull/1952/commits/583e24eef7a6d7e75448976b89ab3d58e3e296dc

nssp/delphi_nssp/pull.py
minhkhul: https://github.com/cmu-delphi/covidcast-indicators/pull/1952/commits/583e24eef7a6d7e75448976b89ab3d58e3e296dc

nssp/delphi_nssp/pull.py
minhkhul: https://github.com/cmu-delphi/covidcast-indicators/pull/1952/commits/583e24eef7a6d7e75448976b89ab3d58e3e296dc

nssp/delphi_nssp/pull.py
minhkhul: updated. https://github.com/cmu-delphi/covidcast-indicators/pull/1952/commits/971968e11bc69c0631e11c534df6b319598f529b

nssp/delphi_nssp/constants.py
minhkhul: https://github.com/cmu-delphi/covidcast-indicators/pull/1952/commits/38cd523571218ec2405c79cec0fc6b427ee86de8

nssp/params.json.template
minhkhul: sircal: https://github.com/cmu-delphi/covidcast-indicators/pull/1952/commits/2247e1b9fbc9152c47089a319cc56bc5d403be31
validation: https://github.com/cmu-delphi/covidcast-indicators/pull/1952/commits/2bfd5fc17f1eae6cd0bd68736e87b6927142ac9d

nssp/delphi_nssp/run.py
```diff
@@ -0,0 +1,168 @@
...
+import us
+
+def sum_all_nan(x):
+    """Return a normal sum unless everything is NaN, then return that."""
+    all_nan = np.isnan(x).all()
+    if all_nan:
+        return np.nan
+    return np.nansum(x)
+
+def weighted_geo_sum(df: pd.DataFrame, geo: str, sensor: str):
```
nmdefries, line 44: praise: Nice! Thanks for adding the aggregation in. Now we'll have an actual reference for how to aggregate non-summable signals.

nmdefries (commented):
Some remaining documentation cleanup. I also have a few questions about the data. Looks like there is data that is not getting included in the aggregation step -- we'll want to think about what to do with that.

nmdefries:
We'll also need to do a statistical review of the created data. Because there's no formal process for this, you're free to do whatever you think is reasonable/sufficient. I'd say a thorough analysis would include:

- Use [correlations notebook](https://github.com/cmu-delphi/covidcast/blob/5f15f71/R-notebooks/cor_dashboard.Rmd) (example [output](https://cmu-delphi.github.io/covidcast/R-notebooks/signal_correlations.html#)). Helps evaluate potential value in modeling. Choropleths give another way to plot the data to look for weird patterns.
- Run same analysis comparing against other relevant signals -- for signals that are ostensibly measuring the same thing, helps see issues/benefits of one versus the other and how well they agree (e.g. [JHU cases vs USAFacts cases](https://github.com/cmu-delphi/covidcast-indicators/issues/991)).
- Plot signals for all geos over time and space (via choropleth). Look for weird anomalies, missing geos, missing-not-at-random values, etc. Determine reporting behavior for DC and territories.
- Think about [limitations](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/jhu-csse.html#limitations), gotchas, and lag and backfill characteristics.
- What is the source's missing behavior? Missing due to reporting pattern (e.g. no weekend reports)? Do they censor (e.g. for low population)? Use some numeric "missing" like 9999?
- (unlikely) Do we need to do any [interpolation](https://github.com/cmu-delphi/covidcast-indicators/issues/1539)?
- (unlikely) Think about if we should do any filtering/cleaning, e.g. [low sample size](https://github.com/cmu-delphi/covidcast-indicators/issues/1513#issuecomment-1036326474) in covid tests causing high variability in test positivity rate.
- 
Example [analysis 1](https://github.com/cmu-delphi/covidcast-indicators/pull/1495#issuecomment-1039477646), [analysis 2](https://github.com/cmu-delphi/covidcast-indicators/issues/367#issuecomment-717415555). [Part of analysis for wastewater](https://github.com/cmu-delphi/covidcast-indicators/issues/1901#issuecomment-1894712989)

Get the stakeholder to look at your plots and approve if they look reasonable. Some of this investigation will be useful to include in the signal documentation.

I'd recommend putting the plots in the GitHub issue so they're easy to find later.

nssp/delphi_nssp/run.py
```diff
@@ -0,0 +1,168 @@
...
+                df["geo_id"] = "us"
+            elif geo == "state":
+                df = df[(df['county'] == "All") & (df["geography"] != "United States")]
+                df["geo_id"] = df["geography"].apply(lambda x: us.states.lookup(x).abbr.lower() if us.states.lookup(x) else 'dc')
+            elif geo == "hrr" or geo == "msa":
+                df = df[['fips', 'val', 'timestamp']]
+                df = geo_mapper.add_population_column(df, geocode_type="fips", geocode_col="fips")
+                df = geo_mapper.add_geocode(df, "fips", geo, from_col="fips", new_col="geo_id")
+                df = generate_weights(df, "val")
+                df = weighted_geo_sum(df, "geo_id", "val")
```
nmdefries, line 151: note: once the population and new geo_id columns have been added on, a simple weighted mean can be done with

```
df.groupby(["timestamp", "geo_id"]).apply(lambda x: sum(x.population * x.val) / x.population.sum() )
```

but this doesn't handle the `NaN`s appropriately. Actually I'm surprised that there isn't a built-in weighted mean function. `numpy` has an [average function that takes weights](https://numpy.org/doc/stable/reference/generated/numpy.average.html) but also doesn't let you ignore `NaN`s.

nssp/delphi_nssp/run.py
```diff
@@ -0,0 +1,168 @@
...
+            logger.info("Generating signal and exporting to CSV", metric=signal)
+            if geo == "nation":
+                df = df[df["geography"] == "United States"]
+                df["geo_id"] = "us"
+            elif geo == "state":
+                df = df[(df['county'] == "All") & (df["geography"] != "United States")]
+                df["geo_id"] = df["geography"].apply(lambda x: us.states.lookup(x).abbr.lower() if us.states.lookup(x) else 'dc')
+            elif geo == "hrr" or geo == "msa":
+                df = df[['fips', 'val', 'timestamp']]
+                df = geo_mapper.add_population_column(df, geocode_type="fips", geocode_col="fips")
```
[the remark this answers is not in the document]
dsweber2: So I found [this](https://www.cdc.gov/nchs/data/data_acces_files/County-Geography.pdf) which discusses several of these:
- 02201  former Prince of Wales-Outer Ketchikan Census Area (absorbed in 2006Alaska)
- 02232 Skagway-Hoonah-Angoon Census Area (split)
- 02270 Wade-Hampton Census Area, AK (assigned a new code 02158)
- 02280 Wrangell-Petersburg Census Area was split to create part of Wrangell City and Borough
- 51515 In 2013, Bedford City, an independent city, merged with Bedford county (FIPS code=51019)
Most of these are old fips that were kept around for some reason.
missing:
- 02261

I'm going to take a look to see if any of these are actually non-`NA` in the data, but thanks for the catch Nat.

W/out the raw data, I'm not sure how we handle not having populations for the state FIPS (e.g. `50000`; I guess overall minus the actually present counties?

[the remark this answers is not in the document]
nmdefries: Oh, correction on the unassigned state FIPS. We can't actually do anything with them because HRR and MSA are made up of counties or zipcodes. If we were aggregating FIPS -> state, we could use David's suggestion, but NSSP actually reports state values directly.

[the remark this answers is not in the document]
nmdefries: > if any of these are actually non-NA in the data

Good point, I didn't look at that.

nssp/delphi_nssp/run.py
dsweber2: yeah, I had to write these as a way to handle `NaN` correctly for nwss and they're getting reused here. I plan to eventually add them to the geomapper. It is slower than using precomputed weights, so I'm also debating adding those. There's probably also room for optimizing them though.

nmdefries: Yeah, at least an acknowledgement of the assumption would be good to include in the documentation for this data source.

nssp/delphi_nssp/run.py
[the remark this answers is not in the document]
dsweber2: TLDR: Of the non-state ones, `['02201', '02232', '02261', '02270', '02280', '51515']`, only `51515` is non-`NaN`, and it's an exact copy of `51019`.
Code for `51515`:
```
bedford = df[(df.fips =='51515') | (df.fips == '51019')].sort_values("week_end")
relevant_columns = ["percent_visits_combined", "percent_visits_covid", "percent_visits_influenza", "percent_visits_rsv", "percent_visits_smoothed", "percent_visits_smoothed_covid", "percent_visits_smoothed_1", "percent_visits_smoothed_rsv"]
bedford[bedford.fips == '51019']
bedford.transpose()
(bedford[bedford.fips == '51019'][relevant_columns].reset_index(drop = True) == bedford[bedford.fips == '51515'][relevant_columns].reset_index(drop = True)).all()
```
Example code for the alaska ones (`022**`):
```
not_present_fips = "2280"
petersburg = df[(df.fips ==not_present_fips) | (df.fips == present_fips)].sort_values("week_end")
relevant_columns = ["percent_visits_combined", "percent_visits_covid", "percent_visits_influenza", "percent_visits_rsv", "percent_visits_smoothed", "percent_visits_smoothed_covid", "percent_visits_smoothed_1", "percent_visits_smoothed_rsv"]
petersburg[relevant_columns].isnull().all()

```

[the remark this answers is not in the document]
nmdefries: Great, then we don't need to do anything here.

minhkhul:
### Signal behavior across time and geos

#### State-level:
No data for MO and SD throughout the time on all three base signals. 
pct_visits_covid_state
pct_visits_influenza_state
pct_visits_rsv_state

#### County-level:
Following states have no country-level data at all: CA, WA, AK, AZ, AL, CO, SD, ND, MO, AR, FL, OH, NH, CT, NJ
pct_visits_covid_county
pct_visits_influenza_county
pct_visits_rsv_county


There are some outliers in low population counties, creating stats like 50-100% of ER visits being covid/flu related in those places. There are around 10 of these outliers spread among each signals. This makes sense as the signals are percentages. So if county A has only one ER visit this week and that visit is about covid, then the data would show up as 100% of all ER visits in county A is covid-related. 
Screenshot 2024-05-02 at 1 41 34 AM
Screenshot 2024-05-02 at 3 21 48 AM
Screenshot 2024-05-02 at 3 22 22 AM

dsweber2:
Oh one thing I would add from our discussion: CA does have state level data, somehow. I guess the reporting pipelines must be different.

minhkhul:
### Signal behavior across time and geos (Continue)
#### MSA-level
pct_visits_covid_msa
pct_visits_influenza_msa
pct_visits_rsv_msa

minhkhul:
#### DC
DC is the only non-state territory featured in the source. Signals availability seem consistent with state-level signals.
image

nmdefries:
> There are some outliers in low population counties, creating stats like 50-100% of ER visits being covid/flu related in those places. There are around 10 of these outliers spread among each signals. This makes sense as the signals are percentages. So if county A has only one ER visit this week and that visit is about covid, then the data would show up as 100% of all ER visits in county A is covid-related.

Does it make sense to censor these?

Pros:
- privacy
- removes misleading values. Since the data doesn't report sample size (right?), it won't be obvious to a user that a given value is likely calculated from a small sample size.

Cons:
- removes data (although those stats are reported in aggregated values at higher geo levels)
- we're making an assumption about low population -> low ER visits

dsweber2:
Given we're pulling from a public datasource, I would trust them to have done the censoring correctly.

dsweber2:
Some correlation data. I'm comparing the state level data with hhs
# Flu
The overall (spearman) correlation is 0.87, so they are strongly correlated in general (which one very much expects).
## Fixed State, over time
flu_ER_admissions_state_correlations
And the summary stats
```
  cor_min cor_median cor_mean cor_std cor_q25 cor_q75 cor_max
    <dbl>      <dbl>    <dbl>   <dbl>   <dbl>   <dbl>   <dbl>
1   0.730      0.893    0.881  0.0510   0.856   0.916   0.958
```
## Fixed Time over states
flu_ER_admissions_time_correlations
And the summary stats
```
  cor_min cor_median cor_mean cor_std cor_q25 cor_q75 cor_max
    <dbl>      <dbl>    <dbl>   <dbl>   <dbl>   <dbl>   <dbl>
1   0.172      0.573    0.551   0.154   0.448   0.664   0.817
```
Which is significantly lower on average. I think we have a mild example of [Simpson's paradox](https://en.wikipedia.org/wiki/Simpson%27s_paradox) or something along those lines.
## Lag correlations
flu_ER_admissions_state_lag_cor
Also somewhat unsurprisingly, ER visits are most useful with ~a week of lag, but that's it. Given that these are not versioned, the lagging is somewhat suspect for practical purposes.


# Covid
Overall correlation is 0.81, so a bit worse than flu but still quite high.
## Covid Fixed States over time
covid_ER_admissions_state_correlations

## Covid Fixed Time over states
covid_ER_admissions_time_correlations
And the summary stats
```
# A tibble: 1 × 7
  cor_min cor_median cor_mean cor_std cor_q25 cor_q75 cor_max
    <dbl>      <dbl>    <dbl>   <dbl>   <dbl>   <dbl>   <dbl>
1 -0.0873      0.489    0.462   0.190   0.388   0.591   0.806
```
Which is significantly lower on average than both the corresponding flu, and the spatial average.
## Covid Lag analysis
covid_ER_admissions_time_correlations
Somewhat unsurprisingly, the correlation is highest immediately afterward, though its significantly lower than in the flu case.

nssp/delphi_nssp/run.py
dsweber2: Talking with @dshemetov, apparently in the HRR case, the weighted fips -> hrr map is already present, so we may not need to do this. He may add the weighted fips -> msa map in the near future, since the data is already there for it

minhkhul:
#### Quick analysis of nssp signals source backfill behavior

`36 15 * * * ~/myenv/bin/python ~/chunk.py`

I set up this cron job to run the following chunk of code on a server every day in the past 2 weeks. The chunk creates a daily snapshot of data available through socrata API.
<details>
 <summary> Code to take daily snapshot from source api </summary>

```python
import numpy as np
import pandas as pd
from sodapy import Socrata
from datetime import date

today = date.today()
socrata_token = 'sOcrAt4t0k3n' # replace with your own Socrata token
client = Socrata("data.cdc.gov", socrata_token)
results = []
offset = 0
limit = 50000  # maximum limit allowed by SODA 2.0
while True:
    page = client.get("rdmq-nq56", limit=limit, offset=offset)
    if not page:
        break  # exit the loop if no more results
    results.extend(page)
    offset += limit
df_ervisits = pd.DataFrame.from_records(results)
df_ervisits.to_csv(f'{today}.csv', index=False)
```

</details>

Grab those daily snapshots of data and put them into a dictionary of dataframes for analysis. 
<details>
 <summary> Setup for  analysis</summary>

```python
import numpy as np
import pandas as pd
from sodapy import Socrata
from datetime import date
import os
import pandas as pd

# Get the current directory
current_dir = os.getcwd()

# Initialize an empty dictionary to store the dataframes
dataframes = {}

# Iterate over each file in the current directory
for file in os.listdir(current_dir):
    if file.endswith(".csv"):
        # Get the name of the CSV file without the extension
        name = os.path.splitext(file)[0]
        
        # Read the CSV file into a dataframe
        df = pd.read_csv(file)
        
        # Store the dataframe in the dictionary with the name as the key
        dataframes[name] = df
```

</details>

We can see here that (at least recently) the api content is updated every week some time between Thursday night and Friday afternoon.
<details>
  <summary>Compare snapshot code</summary>

  ```python
from datetime import datetime, timedelta
dataframes = {key: dataframes[key] for key in sorted(dataframes)}
for date1 in dataframes:
    date2 = datetime.strptime(date1, "%Y-%m-%d")+ timedelta(days=1)
    date2 = date2.strftime("%Y-%m-%d")
    if date2 not in dataframes:
        continue
    if dataframes[date1].equals(dataframes[date2]):
        print(f"Snapshot {date1} and Snapshot {date2} are alike.")
    else:
        print(f"Snapshot {date1} and Snapshot {date2} are different.")
  ```
Output:
```
Snapshot 2024-04-17 and Snapshot 2024-04-18 are alike.
Snapshot 2024-04-18 and Snapshot 2024-04-19 are different.
Snapshot 2024-04-19 and Snapshot 2024-04-20 are alike.
Snapshot 2024-04-20 and Snapshot 2024-04-21 are alike.
Snapshot 2024-04-21 and Snapshot 2024-04-22 are alike.
Snapshot 2024-04-22 and Snapshot 2024-04-23 are alike.
Snapshot 2024-04-23 and Snapshot 2024-04-24 are alike.
Snapshot 2024-04-24 and Snapshot 2024-04-25 are alike.
Snapshot 2024-04-25 and Snapshot 2024-04-26 are different.
Snapshot 2024-04-26 and Snapshot 2024-04-27 are alike.
Snapshot 2024-04-29 and Snapshot 2024-04-30 are alike.
Snapshot 2024-04-30 and Snapshot 2024-05-01 are alike.
```
</details>

Specifically, every week the source updates with info from the week before.
For example, Friday 2024-04-19 was updated with new data from week end 2024-04-13.
<details>

```python
for datedf in dataframes:
    df = dataframes[datedf]

    # Convert the 'week_end' column to datetime
    df['week_end'] = pd.to_datetime(df['week_end'])

    # Get the row with the most recent date
    most_recent_row = df.loc[df['week_end'].idxmax()]
    
    print(f"Most recent row in {datedf}:")
    print(most_recent_row['week_end'])
```

```
Most recent row in 2024-04-17:
2024-04-06 00:00:00
Most recent row in 2024-04-18:
2024-04-06 00:00:00
Most recent row in 2024-04-19:
2024-04-13 00:00:00
Most recent row in 2024-04-20:
2024-04-13 00:00:00
Most recent row in 2024-04-21:
2024-04-13 00:00:00
Most recent row in 2024-04-22:
2024-04-13 00:00:00
Most recent row in 2024-04-23:
2024-04-13 00:00:00
Most recent row in 2024-04-24:
2024-04-13 00:00:00
Most recent row in 2024-04-25:
2024-04-13 00:00:00
Most recent row in 2024-04-26:
2024-04-20 00:00:00
Most recent row in 2024-04-27:
2024-04-20 00:00:00
Most recent row in 2024-04-29:
2024-04-20 00:00:00
Most recent row in 2024-04-30:
2024-04-20 00:00:00
Most recent row in 2024-05-01:
2024-04-20 00:00:00
```
</details>

More comparison specifically the case between data available on 20240418 and 20240419 [here](https://cmu-delphi.atlassian.net/browse/OKRS24-57).
tldr: there are around 150,000 revised rows and around 3000 new rows in the 2024-04-19 update in comparison to the day before. All new data is about week end 2024-04-13 while revised data was from various period up to 2 years ago (Oct 2022).

dsweber2:
We should get this merged and running on staging @nmdefries @melange396. Every day we wait is revisions that we're losing. What blockers remain?

melange396:
Im working through a big backlog of PRs, but this one is now at the top of my list.  @minhkhul: we should be able to get this running on staging before its merged, right?

minhkhul:
@dsweber2 @melange396 I wrote some code a month ago to call the api everyday and keep all data returned on a separate server. These can be transformed and loaded as patches later. So no worries on missing data.

melange396:
Oh, good!  I thought i remembered having a discussion about that but wasnt sure if it was for this or something else.

ansible/templates/sir_complainsalot-params-prod.json.j2
```diff
@@ -50,6 +50,11 @@
     "hhs": {
       "max_age":15,
       "maintainers": []
+    },
+    "nssp": {
+      "max_age":15,
+      "max_age":13,
```
melange396, line 56: duplicate key; which value is appropriate?

`sir_complainsalot/params.json.template` claims "13" for this, but 
`ansible/templates/nssp-params-prod.json.j2` has `"max_expected_lag":{"all": "15"}`

nssp/setup.py
```diff
@@ -0,0 +1,32 @@
...
+    "delphi-utils",
+    "sodapy",
+    "epiweeks",
+    "freezegun",
+    "us",
+]
+
+setup(
+    name="delphi_nssp",
+    version="0.0.1",
```
melange396, line 20: `nssp/delphi_nssp/__init__.py` claims version "0.1.0"...  We should probably make them agree.

melange396 (dismissed):
I'm gonna do a deeper look at the core code, but i did a first pass to see if i could narrow down what files i needed to inspect more closely; i knocked off 29/46 files that way!

Theres a few things that seem to be out of scope for this PR, like all the `max-line-length=120` in the pylint configs from other indicators...  Are they going to break the lint steps in other indicators?

ansible/templates/sir_complainsalot-params-prod.json.j2
dsweber2: my bad, this came from an apparently botched rebase. Minh specifically set it to 13

nssp/setup.py
dsweber2: switching to `0.1.0` to match other initial release indicators. Though I did notice that e.g. `claims_hosp/version.cfg` has `current _version = 0.3.54`, which seems weird given that it's `setup.py` says `0.1.0`

melange396: oof, thats a problem with our bump2version setup and/or our conventions...  ill make an issue to look at it.

melange396: https://github.com/cmu-delphi/covidcast-indicators/issues/1962

ansible/templates/sir_complainsalot-params-prod.json.j2
melange396: what about the "15" in the other location?  are those not referring to the ~same thing?

notebooks/renv/activate.R
```diff
@@ -0,0 +1,1220 @@
+
+local({
+
+  # the requested version of renv
+  version <- "1.0.7"
```
melange396, line 5: this file seems to be imported from some external source.  will we need to do anything to keep it synced/up-to-date?  i presume (while along side the `notebooks/renv.lock`) it should continue to "just work" until we need it to do something it cant already do?

ansible/templates/sir_complainsalot-params-prod.json.j2
minhkhul: fixed. They are all uniformly max 13 days now, which is based on analysis of api data in early April 2024.
```
Most recent data published on 2024-04-17:
2024-04-06 00:00:00
lag: 12
Most recent data published on 2024-04-18:
2024-04-06 00:00:00
lag: 13
Most recent data published on 2024-04-19:
2024-04-13 00:00:00
lag: 7
Most recent data published on 2024-04-20:
2024-04-13 00:00:00
lag: 8
Most recent data published on 2024-04-21:
2024-04-13 00:00:00
lag: 9
Most recent data published on 2024-04-22:
2024-04-13 00:00:00
lag: 10
Most recent data published on 2024-04-23:
2024-04-13 00:00:00
lag: 11
Most recent data published on 2024-04-24:
2024-04-13 00:00:00
lag: 12
Most recent data published on 2024-04-25:
2024-04-13 00:00:00
lag: 13
```

nmdefries:
Statistical review from Roni:

I just finished reviewing this.  Overall it looks great -- thank you all for your work on this!   I do have some comments/answers/questions:

> Nat: Looks like there is data that is not getting included in the aggregation step -- we'll want to think about what to do with that.

Was this resolved?  If so, how?

> Nat: What is the meaning of a missing value in the incoming data? Censored for privacy? 0? Too small a sample size to accurately report?

Was this resolved?  If so, how?

> Nat: Think about if we should do any filtering/cleaning, e.g. [low sample size](https://github.com/cmu-delphi/covidcast-indicators/issues/1513#issuecomment-1036326474) in covid tests causing high variability in test positivity rate.
> Minh: There are some outliers in low population counties, creating stats like 50-100% of ER visits being covid/flu related in those places. There are around 10 of these outliers spread among each signals. This makes sense as the signals are percentages. So if county A has only one ER visit this week and that visit is about covid, then the data would show up as 100% of all ER visits in county A is covid-related.

This is a known problem but one we cannot solve for now.  I am fairly sure that CDC would want us to maintain all the values, even if they are e.g. 100% or 50%.  But we should alert to this in our documentation.

RE “Covid Lag analysis”, I think you have the wrong data viz here (it’s a copy of the one above it, but should be correlation vs. lag).  Not crucial for releasing the signal, but we should fix it wherever you end up storing the analysis.  Speaking of which: where do we usually store these analyses for new signals?  Are they linked to from the documentation?

As you showed, there is a good number of discrepancies between original state-level and state-level reconstructed from lower geo-level.  Did you figure out the full reason(s) for that?  Are these possibly rounding errors?  Something else?  This too should be discussed in the documentation.

That's all I could find.  To the substance of the signal: the correlations look awesome, esp. for flu -- this was far from assured!  ED and hosp admissions are very different things, and come from very different reporting systems.   But this bodes extremely well for estimating hosp trends without hosp reporting.

nssp/delphi_nssp/run.py
```diff
@@ -0,0 +1,168 @@
...
+    return agg_df
+
+def generate_weights(df, column_aggregating="val"):
+    """
+    Weigh column_aggregating by population.
+
+    generate the relevant population amounts, and create a weighted but
+    unnormalized column, derived from `column_aggregating`
+    """
+    # set the weight of places with na's to zero
```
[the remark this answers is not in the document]
nmdefries: [From David](https://delphi-org.slack.com/archives/C06LACZ4AKE/p1716221273311199?thread_ts=1716069071.888809&cid=C06LACZ4AKE):

I couldn't find anything written on the subject. Some of them are the redundant values mentioned above. Taking another look, there's exactly the same # of locations (664) on every day that have NA values; 2335 counties have a value every day. Not clear what these mean exactly, though given that there's a fixed nubmer they don't seem too meaningful.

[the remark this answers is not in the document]
nmdefries: We should plan to include in the documentation

nmdefries (approved):
Looks good

nmdefries:
Next steps:
- check that this runs and output looks correct on staging
- loop in Brian to get this set up in production
- I will review the [docs](https://github.com/cmu-delphi/delphi-epidata/pull/1439/files)

minhkhul:
@nmdefries cool. I'll get that going.

melange396:
when merging this, please do not forget to do a squash

PR #1958, opened 2024-04-24
first attempt at fixing broken backfill build

dsweber2:
### Description
The buildprocess in https://github.com/cmu-delphi/covidcast-indicators/pull/1957 broke seemingly caused by an update to the R version. This should fix that.

### Changelog

- R CMD CHECK for this package only works on R <= 4.2, so pin Dockerfile to `rocker/tidyverse:4.2`
- Update Dockerfile to use `pak` and `rspm` to install dependencies via binaries (removed system dependencies install command because `pak` takes care of those), should speed things up
- Tested the Dockerfile build process locally and it built successfully
- Simplified `backfill-corr-ci.yml` (setup-r-dependencies already caches, so no need for the caching action)

### DIFF
--- a/.github/workflows/backfill-corr-ci.yml
+++ b/.github/workflows/backfill-corr-ci.yml
@@ -10,49 +10,28 @@ name: R backfill corrections
 
 on:
   push:
-    branches: [ main, prod ]
+    branches: [main, prod]
   pull_request:
-    types: [ opened, synchronize, reopened, ready_for_review ]
-    branches: [ main, prod ]
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches: [main, prod]
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     if: github.event.pull_request.draft == false
-    strategy:
-      matrix:
-        r-version: [4.2.1]
     defaults:
       run:
         working-directory: backfill_corrections/delphiBackfillCorrection
 
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up R ${{ matrix.r-version }}
+      - uses: actions/checkout@v4
+
+      - name: Set up R 4.2
         uses: r-lib/actions/setup-r@v2
         with:
-          r-version: ${{ matrix.r-version }}
           use-public-rspm: true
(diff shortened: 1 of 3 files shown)


.github/workflows/backfill-corr-ci.yml
```diff
@@ -10,57 +10,37 @@ name: R backfill corrections
...
+    branches: [main, prod]
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     if: github.event.pull_request.draft == false
-    strategy:
-      matrix:
-        r-version: [4.2.1]
```
nmdefries, line 24: suggestion (optional): maybe we want to keep the `matrix` way of providing `r-version` for future flexibility? In some other packages, we specify several different R versions to run tests on.

.github/workflows/backfill-corr-ci.yml
```diff
@@ -10,57 +10,37 @@ name: R backfill corrections
...
-              libgdal-dev \
-              libudunits2-dev \
-              libglpk-dev \
-              libharfbuzz-dev \
-              libfribidi-dev
-      - name: Get date
-        id: get-date
-        run: |
-          echo "::set-output name=date::$(/bin/date -u "+%Y%m%d")"
-      - name: Cache R packages
```
nmdefries, line 49: praise:

> setup-r-dependencies already caches, so no need for the caching action

Nice! Good simplification here and above.

.github/workflows/backfill-corr-ci.yml
```diff
@@ -10,57 +10,37 @@ name: R backfill corrections
...
+      - uses: actions/checkout@v4
+
+      - name: Set up R 4.2
         uses: r-lib/actions/setup-r@v2
         with:
-          r-version: ${{ matrix.r-version }}
           use-public-rspm: true
-      - name: Install linux dependencies
-        run: |
-            sudo apt-get install \
```
nmdefries, line 38: question: we don't need this section because `r-lib/actions/setup-r-dependencies@v2` installs using `pak`, which as you mentioned above also installs dependencies?

nmdefries (approved):

PR #1964, opened 2024-05-22
add warnings for unreasonable weekday effect

dsweber2:
### Description
A possible source of problems in `doctor_visits` is Weekday adjustment. This adds warnings if the parameters ever get too large, and overrides them if they're ever infinite.

### Changelog
- add warning if the weekday multiplier is ever over 1000 (this is after applying `exp(-params)` appropriately
- add warning and return a null multiplier if the values are infinite

### Fixes 
- Unclear if this fixes `doctor_visits` or not

### DIFF
--- a/_delphi_utils_python/delphi_utils/weekday.py
+++ b/_delphi_utils_python/delphi_utils/weekday.py
@@ -19,9 +19,10 @@ class Weekday:
         series column in the data.
         """
         tmp = data.reset_index()
-        denoms = tmp.groupby(date_col).sum()[denominator_col]
-        nums = tmp.groupby(date_col).sum()[numerator_cols]
-
+        denoms = tmp.groupby(date_col).sum(numeric_only = True)[denominator_col]
+        nums = tmp.groupby(date_col).sum(numeric_only = True)[numerator_cols]
+        if nums.shape[0] < 7:
+            logger.warning("Trying to handle weekday effects with fewer than 7 days worth of data. This will probably not work.")
         # Construct design matrix to have weekday indicator columns and then day
         # indicators.
         X = np.zeros((nums.shape[0], 6 + nums.shape[0]))
@@ -40,7 +41,9 @@ class Weekday:
                 logger.error("Unable to calculate weekday correction")
             else:
                 params[i,:] = result
-
+        if np.exp(-params).max() == np.inf:
+            logger.warning("largest weekday correction is infinite. Defaulting to no correction")
+            params = np.zeros((nums.shape[1], X.shape[1]))
         return params
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


_delphi_utils_python/setup.py
```diff
@@ -27,7 +27,7 @@
 
 setup(
     name="delphi_utils",
-    version="0.3.23",
+    version="0.3.24",
```
melange396, line 30: you dont need to do this manually, the release process takes care of it for you!

https://github.com/cmu-delphi/covidcast-indicators/blob/0c65bb468d9aa3a6c2f37d6f28a5490029b32354/.github/workflows/create-release.yml#L41-L48

_delphi_utils_python/delphi_utils/weekday.py
```diff
@@ -93,11 +102,12 @@ def _fit(X, scales, npnums, npdenoms):
         for scale in scales:
             try:
                 prob = cp.Problem(cp.Minimize((-ll + lmbda * penalty) / scale))
-                _ = prob.solve()
+                _ = prob.solve(solver = cp.ECOS)
```
melange396, line 105: is this the default solver?  i presume youre just making the choice explicit...

_delphi_utils_python/delphi_utils/weekday.py
```diff
@@ -40,7 +41,15 @@ def get_params(data, denominator_col, numerator_cols, date_col, scales, logger):
...
             else:
                 params[i,:] = result
-
+        # the values multiplied by in calc_adjustment
+        # Sunday, the last day here, is a sum of the other days
+        effective_multipliers = np.exp(np.concatenate([-params[:,:6], params[:,:6].sum(axis=0, keepdims = True)]))
+
+        if effective_multipliers.max() == np.inf or effective_multipliers.min() == -np.inf:
+            logger.warning("largest weekday correction is infinite. Defaulting to no correction")
+            params = np.zeros((nums.shape[1], X.shape[1]))
```
melange396, line 50: maybe make this replacement optional?  itd be nice if we could bubble up to the caller that this has happened, but im not sure how to do that nicely without throwing an exception (which defeats the purpose) or by changing the signature on the return (which might break some uses of this class)...  perhaps with a "pass by reference"-style argument that this method can modify?  thats kinda uglor but ¯\\\_(ツ)_/¯

_delphi_utils_python/delphi_utils/weekday.py
dsweber2: Yes, it was the default until recently. This code is still using the old default, and I'm including the explicit restriction to maintain the same behavior. (ECOS isn't the source of the problems, when I was running locally it was using ECOS as wel).

_delphi_utils_python/delphi_utils/weekday.py
dsweber2: in this case, the values are literally impossible to use, since it's multiplying by infinity. I suppose it should probably just kill the whole process.

note that passing all zeros is interpreted as a failure by `doctor-visits`, see here: https://github.com/cmu-delphi/covidcast-indicators/blob/0c65bb468d9aa3a6c2f37d6f28a5490029b32354/doctor_visits/delphi_doctor_visits/update_sensor.py#L136-L138

dsweber2:
closing b/c the problem is better addressed by #1966

Issue #1733, opened 2022-12-09
Deactivate USAFacts indicator


krivard:
The USA Facts source has become unreliable (only 6 issues in all of November 2022) and we should stop running it.
- [x] Turn off Task in Automation
- [x] Update [advanced metadata](https://docs.google.com/spreadsheets/d/1zb7ItJzY5oq1n-2xtvnPBiJu2L3AqmCKubrLkKJZVHs/edit#gid=852629429)
- [x] Move API docs to Inactive section
- [x] Tag final version of the indicator code in this repository
- [x] Delete indicator code

PR #1965, opened 2024-05-28
Deactivate HHS

minhkhul:
### Description
Deactivate hhs. Changes are based on [this](https://github.com/cmu-delphi/covidcast-indicators/issues/1733)

### Changelog
- [remove hhs_hosp dir](https://github.com/cmu-delphi/covidcast-indicators/commit/1a508568f6a5e438450e2f609aae38041b9e5d00)


melange396 (changes requested):
we should not yet remove the code or the ansible template with runtime parameters if we are going to [do patches on this data](https://github.com/cmu-delphi/delphi-epidata/issues/1351).

also, should #1971 be part of this?

nmdefries:
If we want to keep the code around for now, we could turn it off in automation only.

@minhkhul do you know what the timeline on the HHS patches looks like?

Also, before deactivating, we should make a tag for the final hhs code version.

minhkhul:
final tag: https://github.com/cmu-delphi/covidcast-indicators/releases/tag/final-hhs

PR #1905, opened 2023-10-25
feat: unify linter settings across indicators and add incremental formatting

dshemetov:
### Description
Related to https://github.com/cmu-delphi/delphi-epidata/pull/1328, unifying linting settings across indicators and these two repos.

### Changelog
* Use a single `pyproject.toml` for all indicators and `delphi_utils`, removed all `.pylintrc`
  * Made sure all indicators pass with the new config
  * Inlined a few pylint ignores that were previously covered by a rule in a `.pylintrc`, instead of turning them into global rules
  * Some whitespace formatting in the process (on the files I touched)
  * The line-length max is now bumped up to 120 from 100
  * Sorted the dependencies in the `setup.py` files of each indicator while I was at it (ignored in git blame)
* Add `make format` command to apply formatting to the changes on your current branch
* Update the template files as well
* Add a new darker format job to CI
* Add python package caching to CI
* Add "Lint and Format" section to README

PR #1966, opened 2024-05-30
Weekday adjustment to use Clarabel instead of ECOS

minhkhul:
### Description
`doctor_visits` signal `smoothed_adj_cli` has a bug where sometimes some percentage values are calculated to be abnormally big (90%+). We found out it's because of this ECOS solver used by default by cvxpy when calculating weekday adjustments.

Per their recent [update](https://www.cvxpy.org/updates/index.html):

> CVXPY has used ECOS as the default solver for many years; however, it has known issues with performance and numerical stability in edge cases. Recently, a new solver, Clarabel, that improves the algorithm and implementation of ECOS has been under development.

### Changelog
- Force weekday adjustment to use Clarabel in all versions of cvxpy, not just 1.5+.

### Fixes 
- Annoying doctor_visits bug


melange396:
Since the python client bug is fixed, i reran the test actions.  Now `test_weekday.py` is [failing](https://github.com/cmu-delphi/covidcast-indicators/actions/runs/9307684510/job/25660378911#step:7:64) because [the expected output values](https://github.com/cmu-delphi/covidcast-indicators/blob/6bccf68be5ff8f79e125fba201235552d2c97c04/_delphi_utils_python/tests/test_weekday.py#L16-L39) have changed slightly.

dshemetov:
You can cherrypick this commit if you want to fix just that CI issue 3559664a25bfab4c33b066a4c0d4c91441c47f60. It's included in the #1952 and (the same commit with a different hash is in) #1905.

melange396:
weird!  @dshemetov do you know what caused that?

minhkhul:
@melange396 I'm gonna guess it's related to clarabel solver being made default recently in cvxpy instead of ecos.

dshemetov:
Yea, it's coming from differences in the solvers. Clarabel is a Rust implementation and possibly with algorithmic differences in the optimization procedures as well. Some details in [this changelog](https://github.com/cvxpy/cvxpy/releases/tag/v1.5.0).

melange396:
ah yes, the default changed with the new version, and it bit them in those other PRs not long after the release.  This is an argument for using version pinning...  I will create a GH issue for that.

dsweber2 (approved):
Yep, glad to see this! Strange that they differ even in the 4th digit; I would've expected them to converge to `1e-8`

dsweber2:
That would make sense; it's a thing to do after this has gone live and been run on the old data though, right? Added as a subtask on the [Jira patch ticket](https://cmu-delphi.atlassian.net/browse/OKRS24-36)

Judging by the effect on the test case, it should cause changes on the order of ~.1%, which is probably not worth worrying about too much. If anything the values should be closer to the true optimum.

PR #1971, opened 2024-06-05
remove hhs and chng from sircal

minhkhul:
### Description
HHS and Chng are no longer active. So we're removing these items from sircal daily checks/complaints.


nmdefries:
HHS changes look fine, but I'm not sure we should be turning off alerts for CHNG. We're still expecting it to come back online at some point, right? Maybe it's safer (to keep us from forgetting about it) to keep the alerts going in Slack -- what do you think?

melange396:
@nmdefries Fair point, but it is not clear right now when CHNG will come back nor what it will look like when it does.  We also have the "ChangeHealthcare FTP Checker" in the same slack channel that should let us know if our access is restored, and the SirCAL alert just adds more noise...  Maybe just comment out the CHNG stuff for the time being?

ansible/templates/sir_complainsalot-params-prod.json.j2
```diff
@@ -12,11 +12,6 @@
       "maintainers": ["U01AP8GSWG3","U01069KCRS7"],
       "retired-signals": ["smoothed_covid19","smoothed_adj_covid19"]
     },
-    "chng": {
-      "max_age": 6,
-      "maintainers": ["U01AP8GSWG3","U01069KCRS7"],
-      "retired-signals": ["7dav_outpatient_covid","7dav_inpatient_covid"]
-    },
```
nmdefries, line 19: suggestion (optional): Maybe just comment out the CHNG stuff for the time being instead of removing it entirely, so it's easy to find/reference later

nmdefries (approved):
Looks fine.

ansible/templates/sir_complainsalot-params-prod.json.j2
minhkhul: So til you cannot comment in json files.

PR #1976, opened 2024-06-17
Reformat NSSP county zip code

minhkhul:
### Description
[Testing NSSP on staging](https://cronicle-prod-01.delphi.cmu.edu/#JobDetails?id=jlxj2cuv60v) shows some NSSP geo_ids at county level don't match what's expected in our db and so cannot be ingested. This is because our db takes 5-digit zip codes (e.g. 08059), while nssp source may note the same county down with 4-digit (8059). 

This fix adds a leading 0 all 4-digit zip codes present in nssp source.

### Changelog
This fix adds a leading 0 all 4-digit zip codes present in nssp source when pulling from API.

### DIFF
--- a/nssp/delphi_nssp/pull.py
+++ b/nssp/delphi_nssp/pull.py
@@ -68,7 +68,7 @@ def pull_nssp_data(socrata_token: str):
         raise ValueError(warn_string(df_ervisits, TYPE_DICT)) from exc
 
     # Format county fips to all be 5 digits with leading zeros
-    df_ervisits['fips'] = df_ervisits['fips'].apply(lambda x: str(x).zfill(5) if str(x) != '0' else '0')
+    df_ervisits["fips"] = df_ervisits["fips"].apply(lambda x: str(x).zfill(5) if str(x) != "0" else "0")
 
     keep_columns = ["timestamp", "geography", "county", "fips"]
     return df_ervisits[SIGNALS + keep_columns]
--- a/nssp/delphi_nssp/pull.py
+++ b/nssp/delphi_nssp/pull.py
@@ -67,5 +67,8 @@ def pull_nssp_data(socrata_token: str):
     except KeyError as exc:
         raise ValueError(warn_string(df_ervisits, TYPE_DICT)) from exc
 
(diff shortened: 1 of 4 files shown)


nssp/delphi_nssp/pull.py
```diff
@@ -67,5 +67,8 @@ def pull_nssp_data(socrata_token: str):
     except KeyError as exc:
         raise ValueError(warn_string(df_ervisits, TYPE_DICT)) from exc
 
+    # Format county fips to all be 5 digits with leading zeros
+    df_ervisits["fips"] = df_ervisits["fips"].apply(lambda x: str(x).zfill(5) if str(x) != "0" else "0")
```
dsweber2, line 71: I checked with [the docs](https://pandas.pydata.org/docs/reference/api/pandas.Series.str.zfill.html) that this is prepending. I guess "0" is supposed to just be zero? Could easily see it being "00000"

dsweber2 (approved):
pretty straightforward, sorry for the delays

nssp/delphi_nssp/pull.py
minhkhul: For all the rows with state-level data and nation-level data in the source, "0" is the value of fips column (instead of empty or nan). Something like this example in the test file:
```
    {
        "week_end": "2022-10-01T00:00:00.000",
        "geography": "United States",
        "county": "All",
        "percent_visits_combined": "2.84",
        "percent_visits_covid": "1.84",
        "percent_visits_influenza": "0.48",
        "percent_visits_rsv": "0.55",
        "percent_visits_smoothed": "2.83",
        "percent_visits_smoothed_covid": "2.07",
        "percent_visits_smoothed_1": "0.34",
        "percent_visits_smoothed_rsv": "0.44",
        "ed_trends_covid": "Decreasing",
        "ed_trends_influenza": "Increasing",
        "ed_trends_rsv": "Increasing",
        "hsa": "All",
        "hsa_counties": "All",
        "hsa_nci_id": "All",
        "fips": "0",
        "trend_source": "United States"
    },
```
So the if condition above is used to make sure we're only applying this 5-digit rule to county-level fips.

PR #1977, opened 2024-06-20
Doctor_visits patching code

minhkhul:
### Description
This change automates multi-day patching for `doctor_visits` indicators, producing `csv` files organized in batch issue format: `[name-of-patch]/issue_[issue-date]/nssp/actual_data_file.csv`, making it easier to ingest large batches of patch data.

Users start by specify in `params.json` the start issue and end issue of the patch, along with a directory where output is stored in. For example:
```{json}
  "patch": {
    "patch_dir": "/Users/minhkhuele/Desktop/delphi/covidcast-indicators/doctor_visits/AprilPatch",
    "start_issue": "2024-04-20",
    "end_issue": "2024-04-21"
  }
```
the run patch.py will create a directory 'covidcast-indicators/doctor_visits/AprilPatch' containing:

```
AprilPatch/
    issue_20240420/
        doctor-visits/
            20240408_state_smoothed_adj_cli.csv
            20240408_state_smoothed_cli.csv
            ...
    issue_20240421/
        doctor-visits/
            20240409_state_smoothed_adj_cli.csv
            20240408_state_smoothed_cli.csv
            ...
```

Issue #1972, opened 2024-06-06
replace the python covidcast client in validator


melange396:
Similar to #1931, the python `covidcast` client (which has some issues) is used in the `delphi_utils.validator` submodule and it can should come out of there:
- https://github.com/cmu-delphi/covidcast-indicators/blob/a6ea00355d79a47cc09db8090f6e5121c718de0f/_delphi_utils_python/delphi_utils/validator/run.py#L8
- https://github.com/cmu-delphi/covidcast-indicators/blob/a6ea00355d79a47cc09db8090f6e5121c718de0f/_delphi_utils_python/delphi_utils/validator/datafetcher.py#L12
- https://github.com/cmu-delphi/covidcast-indicators/blob/a6ea00355d79a47cc09db8090f6e5121c718de0f/_delphi_utils_python/delphi_utils/validator/dynamic.py#L8

melange396:
One interesting quirk of the covidcast client is that it [pins "`pandas<2`"](https://github.com/cmu-delphi/covidcast/blob/c89e4d295550ba1540d64d2cc991badf63ad04e5/Python-packages/covidcast-py/setup.py#L23), which nets [version 1.5.3](https://pandas.pydata.org/docs/whatsnew/v1.5.3.html) (from ~1.5y ago)...

I find no reason/rationale for this in either the source or comments of https://github.com/cmu-delphi/covidcast/pull/618 where the change was made :(

This also means that anything that imports the `delphi_utils` package inherits the version restriction.

The current version of Pandas is 2.2.2 -- which might offer significant performance improvements, but it risks breaking any number of things.  Pandas is not pinned anywhere in this repo (except for the now-toothless [spec in `delphi_utils` to >=1.1.0](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/_delphi_utils_python/setup.py#L19)), so completing this issue may require changes or pinning in individual indicator code.

melange396:
#1985 is an effective stop-gap for removing the dependency implications (and it also partially fulfills #1931 and #1987 ), but it doesnt address the `covidcast` client inefficiencies mentioned in #1931...

to properly satisfy that , i think it should be sufficient to:
[...]

PR #1978, opened 2024-06-24
Refactor doctor_visits for faster runtime

minhkhul:
### Description
Currently, doctor_visits source files have multiple columns with slightly different names but refer to the same kind of data:
```
"servicedate": "ServiceDate", "patCountyFIPS": "PatCountyFIPS",
"patHRRname": "Pat HRR Name", "patAgeGroup": "PatAgeGroup",
"patHRRid": "Pat HRR ID"
```

We deal with this by this process:
1. Unzip and load source file into  dataframe
2. Rename columns to a standard format 
3. Rewrote the whole source file into its original storage 
4. Reopen the source file and load into another dataframe 
5. Continue processing data.

This PR remove the first round of loading source file and rename columns. 

### Changelog
- remove modify_claims_drops related files and tests
- replace with code and tests that does the same thing in update_sensor.py


doctor_visits/delphi_doctor_visits/update_sensor.py
```diff
@@ -86,12 +87,16 @@ def update_sensor(
...
     # value cols: Denominator, Covid_like, Flu_like, Flu1, Mixed
+    filename = Path(filepath).name
     data = pd.read_csv(
         filepath,
-        usecols=Config.FILT_COLS,
         dtype=Config.DTYPES,
-        parse_dates=[Config.DATE_COL],
     )
+    logger.info(f"Starting processing {filename} ")
+    data.rename(columns=Config.DEVIANT_COLS_MAP, inplace=True)
```
melange396, line 96: There is reason to believe that [the DataFrame column renaming operation might be slow even w/ `inplace=True`](https://stackoverflow.com/questions/64264211/why-is-renaming-columns-in-pandas-so-slow), especially because all of the indicators code is stuck on Pandas 1.5.3 [for silly reasons](https://github.com/cmu-delphi/covidcast-indicators/issues/1972#issuecomment-2197688725).  An alternative to [renaming the columns](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/doctor_visits/delphi_doctor_visits/modify_claims_drops.py#L32-L41) would be to adjust [the column name "constants"](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/doctor_visits/delphi_doctor_visits/config.py#L23-L28) depending on whichever are present in the input file.

You removed one read->rename->write cycle for the source data in `modify_and_write()`, but the file is still read 12 times here ([once for each combination in the loop over 6 geo_types and 2 "weekdays"](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/doctor_visits/delphi_doctor_visits/run.py#L103-L109))...  So moving the renaming to here could be compounding the problem, thus explaining the longer runtime you saw.  Additionally, doing the column filtering and date parsing outside of `read_csv()`, and the column filtering after the `rename()` could also have performance hits.  The only way to know for sure would be to time the individual pandas operations.

Regardless, i think that we can save a lot of running time by moving stuff to outside of `update_sensor()` and outside the geo_type+weekday loop: 
- [file reading and some post-processing (including generating the Weekday params)](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/doctor_visits/delphi_doctor_visits/update_sensor.py#L90-L149) only needs to be done once
- then the [geo_id-specific stuff](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/doctor_visits/delphi_doctor_visits/update_sensor.py#L150-L176) and file outputting will be all thats left inside the loop.

Im not sure why we are [not](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/ansible/templates/doctor_visits-params-prod.json.j2#L14) using the [parallelized version of the geo_id-specific processing](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/doctor_visits/delphi_doctor_visits/update_sensor.py#L179-L208), but i have to assume that (a) it doesnt work or (b) it wasnt fully tested or (c) its slow due to having to copy the big dataframe to all of the spawned threads or (d) it mightve even caused OOMs.

For even more optimizations, [a data copy inside the `Weekday` class](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/_delphi_utils_python/delphi_utils/weekday.py#L144) can be avoided, and something tells me [`write_to_csv()`](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/doctor_visits/delphi_doctor_visits/update_sensor.py#L27-L67) could be improved a good bit.

aysim319:
### Description
continuing optimizing for doctors_visit
Main ran: 569 (
[doctor_visit_EDI_AGG_OUTPATIENT_26062024_1455CDT.log](https://github.com/user-attachments/files/16057288/doctor_visit_EDI_AGG_OUTPATIENT_26062024_1455CDT.log)
)
doctor_visit_refactor: 94 (
[doctor_visit_refactored_EDI_AGG_OUTPATIENT_26062024_1455CDT.log](https://github.com/user-attachments/files/16057348/doctor_visit_refactored_EDI_AGG_OUTPATIENT_26062024_1455CDT.log)
)


Changelog
refactored update sensor and moved the csv processing into separate module
using datetime date params instead of string
using dask to read in csv

doctor_visits/delphi_doctor_visits/process_data.py
```diff
@@ -0,0 +1,101 @@
...
+
+                if se:
+                    assert sensor > 0 and se_val > 0, "p=0, std_err=0 invalid"
+                    outfile.write(
+                        "%s,%f,%s,%s,%s\n" % (geo_id, sensor, se_val, "NA", "NA"))
+                else:
+                    # for privacy reasons we will not report the standard error
+                    outfile.write(
+                        "%s,%f,%s,%s,%s\n" % (geo_id, sensor, "NA", "NA", "NA"))
+                out_n += 1
```
dshemetov, line 50: suggestion: this whole block should use Pandas built-ins and avoid the slow `itertuples()` for loop (see the first warning box [here](https://pandas.pydata.org/docs/user_guide/basics.html#iteration)). something like this:

```py
assert not single_data_df.val.isna().all(), "sensor value is nan, check pipeline"
sensor_anomalies = single_data_df[(100 * single_data_df.val) < 90]
assert sensor_anomalies.empty(), f"strangely high percentage {sensor_anomalies['geo_id', 'val'].iloc[0]}"

# etc. etc.

single_date_df.mutate(sensor = 100 * val, se_val = 100 * se).to_csv(filename, columns = ["geo_id", "val", "se", "direction", "sample_size"], na_rep="NA")
```

doctor_visits/delphi_doctor_visits/process_data.py
```diff
@@ -0,0 +1,162 @@
...
+    filtered and formatted dataframe
+    """
+    # report in percentage
+    df["val"] = df["val"] * 100
+    df["se"] = df["se"] * 100
+
+    val_isnull = df["val"].isnull()
+    df_val_null = df[val_isnull]
+    if not df_val_null.empty:
+        logger.info("sensor value is nan, check pipeline")
```
dshemetov, line 58: question: so previously these were `assert` statements that would halt the pipeline. The pro is that kept potential bad values out of our db, the con is the pipeline could get tanked by one bad row. The new behavior is reversed. Do we want to keep the old behavior or have the new behavior? I vaguely prefer the old way, given how tough it is to patch things for us in general, though maybe that changes with #1977. thoughts?

dshemetov (changes requested):

doctor_visits/delphi_doctor_visits/process_data.py
minhkhul: 1977 makes patching faster, but it's still the same process at its core. I agree we should keep the assert.

aysim319:
output of one day of state and cprofile output with validation scripts
[profile_results.zip](https://github.com/user-attachments/files/16195406/profile_results.zip)

doctor_visits/setup.py
```diff
@@ -11,6 +11,7 @@
     "pytest-cov",
     "pytest",
     "scikit-learn",
+    "dask>=2024.6, <2024.7",
```
melange396, line 14: i think this is the syntax you want...  `dask` uses a 3-tier version format:
```suggestion
    "dask>=2024.6.*,<2024.7.0",
```
also see https://peps.python.org/pep-0440/#version-specifiers

melange396 (changes requested):

doctor_visits/delphi_doctor_visits/process_data.py
```diff
@@ -0,0 +1,181 @@
...
+        logger.info(f"========= WARNING: WRITING SEs TO {out_name} =========")
+
+    out_n = 0
+    for d in set(output_df["date"]):
+        filename = "%s/%s_%s_%s.csv" % (output_path, (d + Config.DAY_SHIFT).strftime("%Y%m%d"), geo_level, out_name)
+        single_date_df = output_df[output_df["date"] == d]
+        with open(filename, "w") as outfile:
+            outfile.write("geo_id,val,se,direction,sample_size\n")
+
+            for line in single_date_df.itertuples():
```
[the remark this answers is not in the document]
nmdefries: I see Dmitry [already commented](https://github.com/cmu-delphi/covidcast-indicators/pull/1978#discussion_r1661409852) on this with some example code to use.

nmdefries (commented):
Some style comments to get started. I'm making another pass for logic/flow.

PR #1980, opened 2024-06-28
fix doctor_visits log location & export_dir

melange396:
updated to be in shared system indicators log directory, the same way that the other indicators are configured

### DIFF
--- a/ansible/templates/doctor_visits-params-prod.json.j2
+++ b/ansible/templates/doctor_visits-params-prod.json.j2
@@ -1,6 +1,6 @@
 {
   "common": {
-    "export_dir": "./receiving",
+    "export_dir": "/common/covidcast/receiving/doctor-visits",
     "log_filename": "/var/log/indicators/doctor-visits.log"
   },
   "indicator": {
--- a/ansible/templates/doctor_visits-params-prod.json.j2
+++ b/ansible/templates/doctor_visits-params-prod.json.j2
@@ -1,7 +1,7 @@
 {
   "common": {
     "export_dir": "./receiving",
-    "log_filename": "./doctor-visits.log"
+    "log_filename": "/var/log/indicators/doctor-visits.log"
   },
   "indicator": {
     "input_file": "./input/SYNEDI_AGG_OUTPATIENT_18052020_1455CDT.csv.gz",


ansible/templates/doctor_visits-params-prod.json.j2
```diff
@@ -1,7 +1,7 @@
 {
   "common": {
     "export_dir": "./receiving",
```
minhkhul, line 3: I monkeypatched this when first setting up the move to prod, so it hasn't been a problem, but this can go here too for consistency. 
```suggestion
    "export_dir": "/common/covidcast/receiving/doctor-visits",
```

minhkhul (changes requested):

minhkhul:
Thanks for spotting this!

korlaxxalrok (commented):
Looks good to me 👍

ansible/templates/doctor_visits-params-prod.json.j2
melange396: if thats the case, then should we also change it in [quidel_covidtest](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/ansible/templates/quidel_covidtest-params-prod.json.j2#L3) and [google_symptoms](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/ansible/templates/google_symptoms-params-prod.json.j2#L3)?  i dont remember exactly how parameter is used (off the top of my head), but its concerning that we have such different values yet all of these indicators seem to be working??

theres also a similar line in the (unused?) [backfillcorr](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/ansible/templates/backfillcorr-params-prod.json.j2#L8) and a slightly different set of parameters for [nchs_mortality](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/ansible/templates/nchs_mortality-params-prod.json.j2#L3-L5)...

minhkhul: So [quidel](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/ansible/templates/quidel_covidtest-params-prod.json.j2#L65) and [google_symptoms](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/ansible/templates/google_symptoms-params-prod.json.j2#L60) use `delivery_dir` for the same purpose as this indicator and others like hhs and claims, etc. use `export_dir`. For those two indicators, indicator run will produce csv files into `export_dir`, then those files [get moved ](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/_delphi_utils_python/delphi_utils/utils.py#L91)to `delivery_dir` for acquisition to pick up. For other indicators, we just put the csv files straight into where acquisition pick up (`/common/covidcast/receiving/indicator`). 

nchs results are run daily, but results are [uploaded to covidcast api only on Monday and Thursday (weekly)](https://github.com/cmu-delphi/covidcast-indicators/blob/88e14d2bcec22e7bb4d1ba7f2f403701735ca1ab/nchs_mortality/delphi_nchs_mortality/archive_diffs.py#L29-42), hence the slightly different set of params from the rest.

minhkhul (approved):

Issue #1931, opened 2024-01-11
replace the python covidcast client in SirCAL


melange396:
https://github.com/cmu-delphi/covidcast-indicators/blob/833e81834e32d7ae91a2d4382b6e411095355290/sir_complainsalot/delphi_sir_complainsalot/check_source.py#L75-L80

SirCAL uses the python covidcast client which has some issues including inefficiency, as [it makes individual http requests for each of the days in the specified range](https://github.com/cmu-delphi/covidcast/blob/c89e4d295550ba1540d64d2cc991badf63ad04e5/Python-packages/covidcast-py/covidcast/covidcast.py#L440-L445) (and also even in the deprecated asynchronous mode, though they are ~simultaneous).

aysim319:
blocked by #1972

PR #1985, opened 2024-07-03
refactor: bundle covidcast==0.2.2 in delphi_utils

### LINKED ISSUE #1972: replace the python covidcast client in validator

dshemetov:
### Description
Tried adding a simple wrapper around `delphi_epidata.Epidata` in here to avoid `covidcast`, but it grew to become most of the `covidcast` client from the other repo. So I just bundled the `covidcast==0.2.2` in here, which removes transitive dependencies on plotting and geo libaries and unpins Pandas, but also doesn't require us to do any new testing of the client library. 

Eventually, we should replace this with epidatpy, but that is possibly months from release. 

The main thing we should pay attention to is the Pandas version changing our code behavior (both in client and in indicator pipelines) - we went from pandas-1.5.3 to pandas-2.0.3 (in CI below; not sure why it didn't go up to 2.2.x, pip might just prefer installing a cached version that satisfies the resolver). The unit tests pass below, but it's possible they missed some things. We'll need to look at what behaviors changed from 1.5.3 to 2.0.0. Or we can just pin to 1.5.3 and punt the Pandas bump to another PR.

### Changelog
- Moved covidcast.py from [here](https://github.com/cmu-delphi/covidcast/blob/main/Python-packages/covidcast-py/covidcast/covidcast.py) to this repo and brought the test file over as well. 
- Updated covidcast imports in this repo.

### Fixes 
- Partially addresses #1972
- Partially addresses #1931 
- Partially addresses #1987

### DIFF
--- a/_delphi_utils_python/delphi_utils/covidcast_port.py
+++ b/_delphi_utils_python/delphi_utils/covidcast_port.py
@@ -1,0 +1,580 @@
+"""This is the client side library for accessing the COVIDcast API.
+
+Copied (almost exactly) from covidcast==0.2.2:
+
+    https://github.com/cmu-delphi/covidcast/blob/c89e4d295550ba1540d64d2cc991badf63ad04e5/Python-packages/covidcast-py/covidcast/covidcast.py
+
+The only changes made were:
+- move NoDataWarning from errors.py to this file
+- hardcode the version to 0.2.2
+- format with black
+"""
+
+import warnings
+from datetime import date, timedelta
+from functools import reduce
+from typing import Iterable, List, Tuple, Union
+
+import numpy as np
+import pandas as pd
+from delphi_epidata import Epidata
+from delphi_epidata.delphi_epidata import _HEADERS
+from epiweeks import Week
+
+
+class NoDataWarning(Warning):
+    """Warning raised when no data is returned on a given day by covidcast.signal()."""
+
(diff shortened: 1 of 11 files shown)


melange396 (commented):
This is an effective stop-gap for removing the dependency implications of #1972 (and it also partially fulfills #1931 and #1987 ), but it doesnt address the `covidcast` client inefficiencies mentioned in #1931.  It doesnt "replace" the covidcast client as much as it just sources it from a different place...

dshemetov:
Superseded by #2004

PR #1983, opened 2024-07-01
Move pipeline manual to repo

nmdefries:
### Description
Migrating from [Gdrive](https://docs.google.com/document/d/171n814gtrt_hwSZY1ZiN-nYqo2u-nzzBLR6n0RKXeY8/edit), plus adding more detail about statistical review and documentation.

PR #1986, opened 2024-07-03
Docs on new signal deployment details

minhkhul:
### Description
Add deployment details to pipeline guide: CI/CD, staging test runs, cronicle setups.

### DIFF
--- a/_template_python/INDICATOR_DEV_GUIDE.md
+++ b/_template_python/INDICATOR_DEV_GUIDE.md
@@ -238,7 +238,7 @@ This example is taken from [`hhs_hosp`](https://github.com/cmu-delphi/covidcast-
 
 The column is described [here](https://cmu-delphi.github.io/delphi-epidata/api/missing_codes.html).
 
-#### Testing
+#### Local testing
 
 As a general rule, it helps to decompose your functions into operations for which you can write unit tests. To run the tests, use `make test` in the top-level indicator directory.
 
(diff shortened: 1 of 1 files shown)


nmdefries (approved):
You're fast! Thanks!

nmdefries:
Waiting for https://github.com/cmu-delphi/covidcast-indicators/pull/1983 to be approved before merging.

korlaxxalrok (commented):
We might consider keeping some of really specific details (hostnames, users, etc.) for our internal systems in a team-only or otherwise private location.

nmdefries:
@korlaxxalrok I was thinking about that too, but I believe we already have them up on other public repos. Will double-check and link later today.

_template_python/INDICATOR_DEV_GUIDE.md
```diff
@@ -355,25 +355,136 @@ Next, the `acquisition.covidcast` component of the `delphi-epidata` codebase doe
...
+
+Example script:
 
-https://cronicle-prod-01.delphi.cmu.edu/#Schedule?sub=edit_event&id=elr5clgy6rs
+```
+#!/bin/sh
+
+# vars
+user='automation'
+host='app-mono-dev-01.delphi.cmu.edu'
```
nmdefries, line 390: hostname publicly available [here](https://github.com/cmu-delphi/www-epicast/blob/f84314885760ea8d54a62401fe182a86e0e87e4e/docs/epicast_development.md?plain=1#L66) and [here](https://github.com/cmu-delphi/covidcast-indicators/blob/1526dc940f61e5a1ab8ff2ce795f35c965c2e48e/ansible/inventory#L6). User potentially derivable [here](https://github.com/search?q=repo%3Acmu-delphi%2Foperations%20user%20automation&type=code).

nmdefries: @korlaxxalrok which of these should be removed? Do we need to change any user or host names?

_template_python/INDICATOR_DEV_GUIDE.md
```diff
@@ -355,25 +355,136 @@ Next, the `acquisition.covidcast` component of the `delphi-epidata` codebase doe
       12. `value_updated_timestamp`: now
    2. Update the `epimetric_latest` table with any new keys or new versions of existing keys.
 
+### CI/CD
+
+* Add module name to the `build` job in `.github/workflows/python-ci.yml`.
+  This allows github actions to run on this indicator code, which includes unit tests and linting.
+* Add top-level directory name to `indicator_list` in `Jenkinsfile`.
+  This allows your code to be automatically deployed to staging after your branch is merged to main, and deployed to prod after `cocivcast-indicators` is released.
```
dsweber2, line 363:
```suggestion
  This allows your code to be automatically deployed to staging after your branch is merged to main, and deployed to prod after `covidcast-indicators` is released.
```

_template_python/INDICATOR_DEV_GUIDE.md
```diff
@@ -355,25 +355,136 @@ Next, the `acquisition.covidcast` component of the `delphi-epidata` codebase doe
...
-first the indicator run
+The indicator run code is automatically deployed on staging after your branch is merged into main. After merging, make sure you have proper access to Cronicle and staging server `app-mono-dev-01.delphi.cmu.edu` _and_ can see your code on staging at `/home/indicators/runtime/`.
 
-Then the acquisition run
+Then, on Cronicle, create two jobs: First one to run the indicator and second one to load the output csv files into database.
 
-See [@korlaxxalrok](https://www.github.com/korlaxxalrok) or [@minhkhul](https://www.github.com/minhkhul) for more information.
+#### Indicator run job
+
+This job ssh into our staging server and run the indicator, producing csv files output.
```
dsweber2, line 381:
```suggestion
This job signs into our staging server via ssh, runs the indicator, producing csv files as output.
```
verbing `ssh` gets a bit awkward in writing

_template_python/INDICATOR_DEV_GUIDE.md
```diff
@@ -355,25 +355,136 @@ Next, the `acquisition.covidcast` component of the `delphi-epidata` codebase doe
...
+```
+
+Note the staging hostname in `host`.
+
+Note that `ind_name` variable here refer to the top-level directory name where code is located, while `acq_ind_name` refer to the directory name where output csv files are located, which corresponds to the name of `source` column in our database, as mentioned in step 3.
+
+To automatically run acquisition job right after indicator job finishes successfully:
+
+1. In `Plugin` section, select `Interpret JSON in Output`.
+2. In `Chain Reaction` section, select your acquisition run job below to `Run Event on Success`
```
dsweber2, line 408: do you have to write the acquisition job first to have a valid target? If so, it may be easier to write it in the order they're done, rather than the order they're run. And/or put a note to that effect in the general Staging preamble

_template_python/INDICATOR_DEV_GUIDE.md
```diff
@@ -355,25 +355,136 @@ Next, the `acquisition.covidcast` component of the `delphi-epidata` codebase doe
...
+
+You can read more about how the `chain_data` json object in the script above can be used in our subsequent acquisition job [here](https://github.com/jhuckaby/Cronicle/blob/master/docs/Plugins.md#chain-reaction-control).
+
+#### Acquisition job
+
+The acquisition job use chained data from indicator job to determine where csv output files are, then load them into our database.
```
dsweber2, line 414:
```suggestion
The indicator job loads the location of the relevant csv output files into chained data, which the acquisition job then loads into our database.
```

_template_python/INDICATOR_DEV_GUIDE.md
```diff
@@ -355,25 +355,136 @@ Next, the `acquisition.covidcast` component of the `delphi-epidata` codebase doe
...
+```
+
+#### Staging database checks
+
+Apart from checking the logs of staging indicator run and acquisition jobs to identify potential issues with the pipeline, one can also check the contents of staging database for abnormalities.
+
+At this point, acquisition job should have loaded data onto staging mysql db, specifically the `covid` database.
+
+```
+mysql> use covid;
```
dsweber2, line 448: I would maybe mention how to log into the mysql server. I had to go digging through our chatlogs to remember that it was `mysql -u davidweb -p`

dsweber2 (commented):
just a couple of comments/questions as someone who hasn't done deployment (so would have to be just going off these notes to do it).

_template_python/INDICATOR_DEV_GUIDE.md
minhkhul: good idea.

_template_python/INDICATOR_DEV_GUIDE.md
melange396: It doesnt hurt to scrub as much of the sensitive stuff as possible...  Even if some bits of it are already "out there", reducing the number of current instances shrinks the attack surface (kinda) and makes it a little harder for a malicious actor to find useful information.

I definitely prefer having documentation like this in version control -- its easier to find, use, and see the timeline of changes.  It puts instructions and details "next to" the code assets that are being discussed, and in a format closer to whats practical for developers and operations (as opposed to the complicated wysiwyg editor of gdocs).

We could create a new private documentation-only repository or even just a top-level `/documentation` directory in the existing private https://github.com/cmu-delphi/delphi-admin repo.  It wouldnt be as nice as having the docs in the same repo as the code, but it would be protected and cross-references will be handled well by the github web ui (for privileged users).

melange396: ie: 
* https://github.com/cmu-delphi/delphi-admin/issues/204

korlaxxalrok: Something like [Docsify](https://docsify.js.org/#/) might be nice, for eventually building a docs site from Markdown files. I'd maybe push for a separate repo, to make it simpler to clone the repo without getting extra stuff (delph-admin comes with _a lot_ of extra bits), to add access control, and to eventually make it easier to add CI to build a site from it.

We tried something like this before, but with a different tool (mkdocs), ultimately building it into a small site at https://delphi.cmu.edu/systems-handbook. There is a process to clone the repo, make changes, build it locally for review, and publish changes. It never caught on though, and the wheel kept going round.

nmdefries: I agree completely, great points. The "where" is definitely a problem. I wanted to put the "how to make an indicator" manual in this repo so it's close to all the indicator code.

Alternative ideas:

- Put documentation/manuals in relevant repo, but omit host/user/login details and just say "talk to Brian"
- Ditto but link to login details stored in a separate private repo
- Put all documentation/manuals in a dedicated private repo. I like that it'd be on GH, but harder to discover for a given topic
- Put all documentation/manuals in a dedicated folder in GDrive. Also not discoverable

nmdefries: @melange396 [opened an issue](https://github.com/cmu-delphi/delphi-admin/issues/204) for this. `delphi-admin` is private and so a good place to store sensitive info.

PR #1988, opened 2024-07-10
attempt to fix jenkins builds

melange396:
the jenkins builds broke, and the logs led me to believe the [newly released `scs` version](https://pypi.org/project/scs/3.2.6/#:~:text=Released%3A%20Jul%207%2C%202024) ([a dependency of `cvxpy`](https://pypi.org/project/cvxpy/#:~:text=ECOS%20%3E%3D%202-,SCS,-%3E%3D%203.2.4.post1)) might be to blame.  then i found [a VERY recent `scs` gh issue](https://github.com/cvxgrp/scs/issues/283) that looks similar to what we are seeing.

i tried to test this fix in a few ways but couldnt get them to work easily -- i think they'd need deeper changes to our jenkins config, which probbly isnt worth it.  praps we can merge this into `main` and see if it does the trick?

### DIFF
--- a/_delphi_utils_python/setup.py
+++ b/_delphi_utils_python/setup.py
@@ -8,6 +8,7 @@ required = [
     "boto3",
     "covidcast",
     "cvxpy",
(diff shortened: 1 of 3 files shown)


korlaxxalrok (approved):
Looks good to me :shipit:

melange396:
🤞

PR #1995, opened 2024-07-11
switch from input_file to input_dir

minhkhul:
### Description
Originally when doctor_visits was running on bigchunk 2, the series of scripts back then used `input_file` to specify location of one source file extracted from ftp server at a time. Now we use `input_dir` to specify a whole directory where all relevant source files can be extracted into.

### Changelog
- params.json


melange396 (approved):
cool!  you already manually changed this in the `.../doctor_visits/params.json` file on the running system too, right?

PR #2000, opened 2024-07-23
nssp patching code

### LINKED ISSUE #1998: NSSP patching script and NSSP patch april 2024 to july 2024
**Actual Behavior:**
Earliest issue for nssp signals is 30-06-2024 when we query from the api.

**Expected behavior**

Earliest issue at county level could be mid April instead, since we do have source data stashed in big chunk dev 2 for these 3 months.

**Context**
TODO: write code that take these stashed csv files, run nssp indicator on each, and output them in batch issue upload csv orderings.

minhkhul:
### Description

### Changelog
- change1
- change2

### Fixes 
- Fixes #1998

### DIFF
--- a/nssp/delphi_nssp/pull.py
+++ b/nssp/delphi_nssp/pull.py
@@ -27,7 +27,7 @@ def warn_string(df, type_dict):
     return warn
 
 
-def pull_nssp_data(socrata_token: str):
+def pull_nssp_data(socrata_token: str, issue_date: str = None, source_dir: str = None) -> pd.DataFrame:
     """Pull the latest NSSP ER visits data, and conforms it into a dataset.
 
     The output dataset has:
@@ -39,6 +39,11 @@ def pull_nssp_data(socrata_token: str):
     ----------
     socrata_token: str
         My App Token for pulling the NWSS data (could be the same as the nchs data)
+    issue_date: Optional[str]
+        (patching mode only) YYYY-MM-DD formatted date of the issue to pull data for.
+    source_dir: Optional[str]
+        (patching mode only) Path to the directory containing the source data csv files. 
+        The files in source_dir are expected to be named yyyy-mm-dd.csv
     test_file: Optional[str]
         When not null, name of file from which to read test data
 
@@ -47,18 +52,21 @@ def pull_nssp_data(socrata_token: str):
     pd.DataFrame
         Dataframe as described above.
     """
-    # Pull data from Socrata API
-    client = Socrata("data.cdc.gov", socrata_token)
-    results = []
-    offset = 0
-    limit = 50000  # maximum limit allowed by SODA 2.0
-    while True:
-        page = client.get("rdmq-nq56", limit=limit, offset=offset)
-        if not page:
-            break  # exit the loop if no more results
-        results.extend(page)
-        offset += limit
-    df_ervisits = pd.DataFrame.from_records(results)
+    if not issue_date:
+        # Pull data from Socrata API
+        client = Socrata("data.cdc.gov", socrata_token)
+        results = []
+        offset = 0
+        limit = 50000  # maximum limit allowed by SODA 2.0
+        while True:
+            page = client.get("rdmq-nq56", limit=limit, offset=offset)
+            if not page:
+                break  # exit the loop if no more results
+            results.extend(page)
+            offset += limit
+        df_ervisits = pd.DataFrame.from_records(results)
+    else:
+        df_ervisits = pd.read_csv(f"{source_dir}/{issue_date}.csv")
     df_ervisits = df_ervisits.rename(columns={"week_end": "timestamp"})
     df_ervisits = df_ervisits.rename(columns=SIGNALS_MAP)
 
--- a/nssp/delphi_nssp/run.py
+++ b/nssp/delphi_nssp/run.py
@@ -62,7 +62,7 @@ def logging(start_time, run_stats, logger):
     )
 
 
-def run_module(params):
+def run_module(params, logger=None):
     """
     Run the indicator.
 
@@ -72,18 +72,21 @@ def run_module(params):
         Nested dictionary of parameters.
     """
     start_time = time.time()
-    logger = get_structured_logger(
-        __name__,
-        filename=params["common"].get("log_filename"),
-        log_exceptions=params["common"].get("log_exceptions", True),
-    )
+    issue_date = params.get("patch", {}).get("current_issue", None)
+    source_dir = params.get("patch", {}).get("source_dir", None)
+    if not logger:
+        logger = get_structured_logger(
+            __name__,
+            filename=params["common"].get("log_filename"),
+            log_exceptions=params["common"].get("log_exceptions", True),
+        )
     export_dir = params["common"]["export_dir"]
     socrata_token = params["indicator"]["socrata_token"]
 
     run_stats = []
     ## build the base version of the signal at the most detailed geo level you can get.
     ## compute stuff here or farm out to another function or file
-    df_pull = pull_nssp_data(socrata_token)
+    df_pull = pull_nssp_data(socrata_token, issue_date, source_dir)
     ## aggregate
     geo_mapper = GeoMapper()
     for signal in SIGNALS:
--- a/nssp/README.md
+++ b/nssp/README.md
@@ -73,3 +73,33 @@ with the percentage of code covered by the tests.
 
 None of the linting or unit tests should fail, and the code lines that are not covered by unit tests should be small and
 should not include critical sub-routines. 
+
+## Running Patches:
+A daily backup of from source in the form of csv files can be found on bigchunk-dev-02. Ask @minhkhul for more details.
+
+You can also generate your own backup from source by setting up a cron job that runs the following .py every day when a pipeline outtage is going on on our side but aource api is still available:
+```
+import numpy as np
+import pandas as pd
+from sodapy import Socrata
+from datetime import date
+
+today = date.today()
+socrata_token = 'FILL_YOUR_OWN_TOKEN_HERE'
+client = Socrata("data.cdc.gov", socrata_token)
+results = []
+offset = 0
+limit = 50000  # maximum limit allowed by SODA 2.0
+while True:
+    page = client.get("rdmq-nq56", limit=limit, offset=offset)
+    if not page:
+        break  # exit the loop if no more results
+    results.extend(page)
+    offset += limit
+df_ervisits = pd.DataFrame.from_records(results)
+df_ervisits.to_csv(f'~/{today}.csv', index=False)
+```
+When you're ready to create patching data for a specific date range output in batch issue format, adjust `params.json` in accordance with instructions in `patch.py`, move these backup csv files into `source_dir`, then run
+```
+env/bin/python -m delphi_nssp.patch
+```


nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,83 @@
...
+To use this module, you need to specify the range of issue dates in params.json, like so:
+
+{
+  "common": {
+    ...
+  },
+  "validation": {
+    ...
+  },
+  "patch": {
```
aysim319, line 13: As I got feedback from Nat from my google-symptom patch, I realized that if we don't remove the patch parameter, wonky things could happen. So for me, I decided to add another parameter in common called custom_run as a safety check
line 58-60
```
    # safety check for patch parameters exists in file, but not running custom runs/patches
    custom_run_flag = False if not params["common"].get("custom_run") else params["common"].get("custom_run", False)
 ```
 I thought it would be better have a new flag rather than to remove back and forth. 
 
 second note for me too; we should have a section for params.json in the DEV guide of the readme @nmdefries thought? oh and also add patch in the Step 2, development sub section where we list out the core modules that an indicator has

File: nssp/tests/test_patch.py [hunk 2]
```diff
@@ -0,0 +1,38 @@
+import unittest
+from unittest.mock import patch as mock_patch, call
+from delphi_nssp.patch import patch
+import os
+import shutil
+
+class TestPatchModule(unittest.TestCase):
+    def test_patch(self):
```
aysim319, line 8: Can we extend the dates where you're testing right here? I recall you mentioned and also see in the patch code that there's this 
```
        current_issue_week = Week.fromdate(current_issue)
        current_issue_dir = f"""{params["patch"]["patch_dir"]}/issue_{current_issue_week}/nssp"""
 ```
 that's different from just incrementing the issue_date for the directory, the test should check if for 2-3 weeks if the date format is populated properly.
 
 I suggest looking into freezegun python package; it's already used in the google symptom tests, specifically `test_pull.py`  and ~should~ be fairly simple (famous last words)

File: nssp/delphi_nssp/patch.py (hunk 1 above)
Comment (nmdefries) [reply nssp/delphi_nssp/patch.py:13]: There's an issue for [adding patching info](https://github.com/cmu-delphi/covidcast-indicators/issues/1993) to the dev guide. Just added an [issue for adding params info](https://github.com/cmu-delphi/covidcast-indicators/issues/2002) to the dev guide, too.

Comment (minhkhul) [reply nssp/delphi_nssp/patch.py:13]: @aysim319 so for the custom run flag, I wanna make sure the behavior makes sense:
| custom_run_flag  | "patch" and others exist in params | -> run patch?|
| ------------- | ------------- |------------- |
| true  | true | true |
| false  | false  | false |
| true  | false | error?/false |
| false  | true | false |
Is this approach reasonable?

File: nssp/tests/test_patch.py (hunk 2 above)
Comment (minhkhul) [reply nssp/tests/test_patch.py:9]: So I changed the directory mechanism a bit [here](https://github.com/cmu-delphi/covidcast-indicators/blob/8f0bb320c5ede9bb4bc641ea94a3a544f04f1a3d/nssp/delphi_nssp/patch.py#L73-L79) due to how our db handle weekly data coming in being quite different from daily-ingesting sources.

Basically, all weekly data coming in to our db has the `issue_date` column always marked as the first date of the epiweek that the reporting date falls into, rather than the reporting date itself. For example, I query nssp data from our api and got something like this:
```
geo_value                  signal source geo_type time_type time_value  \
0      54079  pct_ed_visits_combined   nssp   county      week 2022-09-25
**issue**  lag  missing_value  missing_stderr  missing_sample_size  value  \
0  **2024-06-30**   93              0               1                    1   3.27
```
Note the issue is `2024-06-30` despite this data coming in on July 5th. It's because `2024-06-30` is the start date of epiweek `202427`, so any data coming in that week will be marked under that issue date. 

Furthermore, the reason why we use this directory format is because of this `batch_issue_format` [implementation in acquisition](https://github.com/cmu-delphi/delphi-epidata/blob/72b4050993fa2080cae1f0570bd9f8febc013428/src/acquisition/covidcast/csv_to_database.py#L43), which demands folder format to be `issue_yyyymmdd` including for patches of weekly data.

Hence, I changed the implementation to fit with these expectations. lmk if all this makes sense.

File: nssp/delphi_nssp/patch.py (hunk 1 above)
Comment (aysim319) [reply nssp/delphi_nssp/patch.py:13]: 1st, 2nd, last is correct. I didn't realize at least for nssp, it doesn't have a start_date, end_date parameter like some of the other ones. 

I think having a warning saying that this flag is on, but no custom date to refer and don't run patch, but just treat it as regular run is probably the way to go. Because we don't want to exit and throw an error when it's not something that could break the program.

File: nssp/tests/test_patch.py (hunk 2 above)
Comment (aysim319) [reply nssp/tests/test_patch.py:9]: That makes sense; I just wanted to see if how it would work if the outage was like couple months, where you would have multiple directories for patching. Does the date really line up? Sanity checks like that

File: nssp/delphi_nssp/run.py [hunk 3]
```diff
@@ -72,18 +72,21 @@ def run_module(params):
         Nested dictionary of parameters.
     """
     start_time = time.time()
-    logger = get_structured_logger(
-        __name__,
-        filename=params["common"].get("log_filename"),
-        log_exceptions=params["common"].get("log_exceptions", True),
-    )
+    issue_date = params.get("patch", {}).get("current_issue", None)
+    source_dir = params.get("patch", {}).get("source_dir", None)
```
aysim319, line 76: We should throw a error and exit the program if there isn't a source_dir

nssp/delphi_nssp/pull.py
```diff
@@ -47,18 +52,21 @@ def pull_nssp_data(socrata_token: str):
...
+        limit = 50000  # maximum limit allowed by SODA 2.0
+        while True:
+            page = client.get("rdmq-nq56", limit=limit, offset=offset)
+            if not page:
+                break  # exit the loop if no more results
+            results.extend(page)
+            offset += limit
+        df_ervisits = pd.DataFrame.from_records(results)
+    else:
+        df_ervisits = pd.read_csv(f"{source_dir}/{issue_date}.csv")
```
aysim319, line 69: What if this file doesn't exist? what should be the behavior while patching if the csv doesn't exist?

nssp/delphi_nssp/run.py
aysim319, line 76: I feel like this should throw an error if there is no source_dir param filled out, because in pull.py line 69
```
df_ervisits = pd.read_csv(f"{source_dir}/{issue_date}.csv")
```
it'll throw an error then

nssp/tests/test_patch.py
```diff
@@ -0,0 +1,47 @@
+import unittest
+from unittest.mock import patch as mock_patch, call
+from delphi_nssp.patch import patch
+import os
+import shutil
+from datetime import datetime, timedelta
+
+class TestPatchModule(unittest.TestCase):
```
aysim319, line 8: I think we need another test that given the patch params it will run the else condition, maybe in test_pull?

you can use the existing test to build the test with the parameters with source_dir and issue_date

nssp/delphi_nssp/patch.py
melange396: Putting patch info into the params file has always bothered me.  It wastes time by complicating the patching process (ie: edit params file, run, remove edits from file), and it is fragile and dangerous (what if someone forgets to remove the file changes?), and probably other downsides that i cant think of right now.  Patches are pretty much always run by hand, so IMNSHO the patch-specific parameters should be specified by commandline arguments.  It will make things easier and more portable.

melange396: If you want the patching process to use parameters from a file instead of CLI (there are good reasons for it), i would recommend creating a separate runner that does something like read a `patch.params.json` file and combining that with the regular `params.json` before passing the result to the regular runner.

That being said, i havent had a chance to review this PR yet, so im not up to speed on what youre doing here...  I was directed here via a comment in #2002 .

aysim319: @melange396 made changes to run as a cli for #1999 is this something you were looking for?

minhkhul: @melange396 I support the extra `patch.params.json` path over the cli path. Last year I when I wrote the early versions of these patching scripts, I actually tried adding and using cli. I ended up going back to params for a few reasons:
- The number and length of input becomes a bit too long and not readable, specifically when long dir/paths are involved.
- When running patches, I notice that I most times I do open up params.json to make other configurations specific to the need of the patch (geos, credentials, archive, etc.). So the process would involve editing params.json anyway, and now I gotta both edit params.json and think of cli configuration.
- Which is why I only run patches on staging (and then put results on prod for acquisition), hence the accidental prod run using patch params won't happen.

nmdefries: I like the idea of having a separate `patch.params.json` file. I like having a saved file available that already has the right format defined (e.g. so you can see that the date needs to be YYYYMMDD, not another format). It decreases time spent chasing down docs and command line arg names.

nssp/delphi_nssp/run.py
nmdefries: If files don't exist, I figure we'll either get an error that we raise ourselves, or an automatic error when the file isn't found, so I don't think it matters too much either way.

nssp/tests/test_patch.py
minhkhul: I see. yep will add that test.

nssp/delphi_nssp/patch.py
nmdefries: > most times I do open up params.json to make other configurations specific to the need of the patch

Hm, if you have to do this a lot it could be inconvenient to have a separate `patch.params.json`, since you'd end up needing to modify two separate files instead of one.

minhkhul: yea that's true lmao. I think @aysim319 suggestion about the flag put us in a good direction for now in terms of preventing accidental patch run.

aysim319: So... are we keeping with having a custom_run flag and having patch params within the original params.json or having a new one for patch.params.json?

minhkhul: For now I'll add the custom run flag for this pr. I think if we want to do patch.params.json it should be a separate effort anyway.

nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,139 @@
...
+from datetime import datetime, timedelta
+from os import makedirs, path
+
+from delphi_utils import get_structured_logger, read_params
+from epiweeks import Week
+
+from .run import run_module
+
+
+def good_patch_config(params, logger):
```
aysim319, line 37: I wonder if we can bake this in the read_params() in delphi_utils.... @nmdefries thoughts?
pydantic doesn't a full pledged support for json schema validation, but seeing how the read_params has no validation what so ever, we could look into it sometime in the future.

nssp/tests/test_patch.py
```diff
@@ -0,0 +1,130 @@
+import unittest
+from unittest.mock import patch as mock_patch, call
+from delphi_nssp.patch import patch, good_patch_config
+import os
+import shutil
+from datetime import datetime, timedelta
+
+class TestPatchModule(unittest.TestCase):
```
aysim319, line 8: Any particular reason for using unittest? Personally I've been using pytest and I think one of the older tickets somewhere said wanted to standardize into using pytest; more pythonic compated to unittests, which is based on Junit and such

nssp/tests/test_patch.py
```diff
@@ -0,0 +1,130 @@
...
+                }
+            }
+            self.assertFalse(good_patch_config(patch_config, mock_logger))
+            mock_logger.error.assert_has_calls([
+                call("Issue dates must be in YYYY-MM-DD format."),
+                call("Source directory bad_source_dir does not exist.")
+            ])
+
+        # Case 4: start_issue after end_issue
+        with mock_patch('logging.Logger') as mock_logger:
```
aysim319, line 58: can you break down the cases into individual tests? I find it easier to understand especially the test name is to test good config but you're also testing all the bad ones; 
the cases can be a part of the test names like test_bad_start_issue_config and etc (also helps when reading the build because it shows the individual tests)

nssp/tests/test_patch.py
minhkhul: I'm a fan of `mock`, hence the use of `unittest`. `mock` is in `unittest`, which is in the Python standard library in recent versions of python. It's the [same mock on pypi](https://docs.python.org/3/library/unittest.mock.html):

> Mock is designed for use with [unittest](https://docs.python.org/3/library/unittest.html#module-unittest) and is based on the ‘action -> assertion’ pattern instead of ‘record -> replay’ used by many mocking frameworks.
> 
> There is a backport of [unittest.mock](https://docs.python.org/3/library/unittest.mock.html#module-unittest.mock) for earlier versions of Python, available as [mock](https://pypi.org/project/mock/) on PyPI.

But ye lemme rewrite the rest to use pytest.

nssp/delphi_nssp/patch.py
minhkhul: I agree. I wrote this method and half way through was like hmm this should probably be generalized since it may be applicable to other indicators too.

nssp/delphi_nssp/pull.py
minhkhul: For the `issue_date.csv`, I wrote [this](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/files#diff-8fce7ae14a0e2097dd3da98817cd6258af5a7596b2a7117339da5ce8abffc875R118-R121) so no next steps (including `pull.py`) will be called with csv that don't exists because the script skip the date right at the start.
Similarly, I just added guardrails in `patch.py` for the source dir [here](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/files#diff-8fce7ae14a0e2097dd3da98817cd6258af5a7596b2a7117339da5ce8abffc875R73-R75).

nssp/tests/test_patch.py
minhkhul: [Added](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/files#diff-9bc87d511e1d29db37345fe99f5a82ad0aac79053f05ea57d303b18569fbfc62R25-R34).

nssp/tests/test_patch.py
minhkhul: yup. added here [e29e07e](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/commits/e29e07e715f3a3854f00fca89fa0568e4b7490ea)

nssp/delphi_nssp/patch.py
aysim319: yeahh I'm like oh I also would like that for mine  👀 and also saw that the current read params isn't doing any validations at all

nssp/delphi_nssp/run.py
minhkhul: The check was [here](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/files#diff-8fce7ae14a0e2097dd3da98817cd6258af5a7596b2a7117339da5ce8abffc875R118-R121) in `patch.py`, so if a date doesn't have source file, we just log out that there's no file for that date, and skip it. `run.py` would not be called at all in this case.
The `custom_run` flag also take it a step further [here](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/files#diff-0f52ac57508ccf1c326cb006bdbc426344d14d0dcdbc79838a67fa0f6999ff90R89-R94). So I think this is very safe.

nssp/tests/test_patch.py
```diff
@@ -0,0 +1,179 @@
...
+            "common": {}
+        }
+        assert not good_patch_config(patch_config, mock_logger)
+        mock_logger.error.assert_has_calls([
+            call("Calling patch.py without custom_run flag set true."),
+            call("Custom flag is on, but patch section is missing."),
+        ])
+
+    @mock_patch('logging.Logger')
+    def test_good_patch_config_case2(self, mock_logger):
```
aysim319, line 23: can you renameing these to describe the the behavior like test_bad_config_missing_end

nssp/tests/test_patch.py
```diff
@@ -0,0 +1,179 @@
...
+                "source_dir": "source_dir"
+            }
+        }
+        assert not good_patch_config(patch_config, mock_logger)
+        mock_logger.error.assert_has_calls([
+            call("Patch section is missing required key(s): end_issue"),
+        ])
+
+    @mock_patch('logging.Logger')
+    def test_good_patch_config_case3(self, mock_logger):
```
aysim319, line 41: likewise for these

nssp/tests/test_patch.py
```diff
@@ -0,0 +1,179 @@
...
+                "source_dir": "source_dir"
+            }
+        }
+        assert good_patch_config(patch_config, mock_logger)
+        mock_logger.info.assert_called_once_with("Good patch configuration.")
+
+    @mock_patch('delphi_nssp.patch.run_module')
+    @mock_patch('delphi_nssp.patch.get_structured_logger')
+    @mock_patch('delphi_nssp.patch.read_params')
+    def test_patch(self, mock_read_params, mock_get_structured_logger, mock_run_module):
```
aysim319, line 97: I see 2 test_patch (here and starting from line 135) and doesn't seem like there's much of a difference. This one is doing an extra assertion for current_issue?

nssp/tests/test_pull.py
```diff
@@ -22,6 +22,16 @@
...
+        source_dir = "source_dir"
+        issue_date = "2021-01-02"
+        test_source_data = pd.read_csv(f"{source_dir}/{issue_date}.csv")
+        result = pull_nssp_data("test_token", issue_date, source_dir)
+
+        assert test_source_data.shape[0] == result.shape[0] # Check if the number of rows are the same
+        mock_socrata.assert_not_called()
+
     @patch("delphi_nssp.pull.Socrata")
     def test_pull_nssp_data(self, mock_socrata):
```
aysim319, line 36: wild print further down;

nssp/delphi_nssp/patch.py
aysim319: I mentioned it the #2002 and I think for right now it's better to remove this and deal with this in another issue

nssp/delphi_nssp/pull.py
```diff
@@ -47,18 +52,21 @@ def pull_nssp_data(socrata_token: str):
...
+        offset = 0
+        limit = 50000  # maximum limit allowed by SODA 2.0
+        while True:
+            page = client.get("rdmq-nq56", limit=limit, offset=offset)
+            if not page:
+                break  # exit the loop if no more results
+            results.extend(page)
+            offset += limit
+        df_ervisits = pd.DataFrame.from_records(results)
+    else:
```
aysim319, line 68: I'm not sure if would be benefictual, but I think it would be nice to add a log saving grabbing from socrata vs grabbing from source

nssp/delphi_nssp/run.py
```diff
@@ -72,18 +70,28 @@ def run_module(params):
...
+            custom_run = False
     export_dir = params["common"]["export_dir"]
     socrata_token = params["indicator"]["socrata_token"]
-
     run_stats = []
     ## build the base version of the signal at the most detailed geo level you can get.
     ## compute stuff here or farm out to another function or file
-    df_pull = pull_nssp_data(socrata_token)
+    if custom_run:
+        issue_date = params.get("patch", {}).get("current_issue", None)
```
aysim319, line 90: I think we want to fail if the issue date or the current issue doesn't exist right? because if we explicitly set the flag to be a custom run, but if current issue is None, then it's going to run a normal run...which would make things confusing since the logger would be for patching, but still doing a regular run.

aysim319: Furthermore if you plan to kepp the patch param validation; you already did the check ahead of time so it should be fine to change it to the [] operator instead of get

nssp/delphi_nssp/pull.py
```diff
@@ -47,18 +52,21 @@ def pull_nssp_data(socrata_token: str):
...
-    while True:
-        page = client.get("rdmq-nq56", limit=limit, offset=offset)
-        if not page:
-            break  # exit the loop if no more results
-        results.extend(page)
-        offset += limit
-    df_ervisits = pd.DataFrame.from_records(results)
+    if not issue_date:
+        # Pull data from Socrata API
+        client = Socrata("data.cdc.gov", socrata_token)
```
aysim319, line 57: I noticed in the logging that it doesn't have any message for when it's grabbing from socrata just the message from create csv. can we also add logging here?

nssp/delphi_nssp/run.py
```diff
@@ -72,18 +70,28 @@ def run_module(params):
...
-    )
+    custom_run = params["common"].get("custom_run", False)
+    # logger doesn't exist yet means run_module is called from normal indicator run (instead of patching)
+    if not logger:
+        logger = get_structured_logger(
+            __name__,
+            filename=params["common"].get("log_filename"),
+            log_exceptions=params["common"].get("log_exceptions", True),
+        )
+        if custom_run:
```
aysim319, line 81: Are you explicitly using the custom_run as patch vs regular run?

nssp/tests/test_patch.py
minhkhul: I'l just remove the shorter test. Basically the original intention is that the short one is a unit test specifically for that function (all calls to anything outside the function is mocked), while the long one actually run a patch from start to finish and check the output result.

nssp/tests/test_patch.py
minhkhul: Added. [6c0abad](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/commits/6c0abad7af8b01d79aded2fec06864ba67a9f7e0)

nssp/tests/test_patch.py
minhkhul: OK so I decided to restructure the test_patch tests. Now the short one (with mocked patch run behavior) will be used to confirm directory structures created, while the long one actually runs the patch and check for correct values. [6c0abad](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/commits/6c0abad7af8b01d79aded2fec06864ba67a9f7e0)

nssp/delphi_nssp/pull.py
minhkhul: good idea.

nssp/delphi_nssp/run.py
minhkhul: If we're in normal patching behavior, which means:
1.`custom_run` is True 
2.`patch.py` was called to do patching
but 
3. `current_issue` is None 
Then the check for current issue being none is already done [here](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/files#diff-8fce7ae14a0e2097dd3da98817cd6258af5a7596b2a7117339da5ce8abffc875R118-R121), so the code won't get to `run.py` in the first place if there's no current issue. It will just skip to the next day's data.
In the above case, logger will be from patching, as [this](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/files#diff-0f52ac57508ccf1c326cb006bdbc426344d14d0dcdbc79838a67fa0f6999ff90R75-R83) block won't get triggered

If we're doing cursed behavior:
1.`custom_run` is True 
but
2. `run.py` was called instead of patch.py
Then [this chunk](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/files#diff-0f52ac57508ccf1c326cb006bdbc426344d14d0dcdbc79838a67fa0f6999ff90R81-R83) will stop the cursed run and assume this is a normal run, set the `custom_run` flag to false automatically, and call `pull_nssp_data(socrata_token)` with no issue date or source dir, which means we're pulling the latest data. 
The logger will be normal run in this case, because [this](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/files#diff-0f52ac57508ccf1c326cb006bdbc426344d14d0dcdbc79838a67fa0f6999ff90R75-R83) block will get triggered.

nssp/delphi_nssp/run.py
minhkhul: As of now, yes.

nssp/delphi_nssp/run.py
aysim319: So you intentionally want the behavior for the "cursed" run to run as a normal run? I think if that's the case, you should have make a comment when you change the custom run flag.

nssp/delphi_nssp/run.py
aysim319: Can you make a comment about it? Something like custom flag only supported for patching, then it makes more sense for the next comment that I had before.

aysim319: Is there a reason? Like wanting to make things more simple/ unnecessary extra work for non patch custom runs that wouldn't be used? A small comment of the reason would be nice. Like if this was because of a technical limitation or non essential/necessary

nssp/delphi_nssp/run.py
aysim319: Okay I see that you were intentionally setting convert to regular run if it's not a patch. If that's the case, then this is fine

aysim319: Ah yup; you are right about the logger! I forgot that it's passed in...

That was a bad example...I still think it's better to explicitly use the [] where it's clear that if the custom flag is true; we expect the issue and the source dir to exist. I was trying to give a weird behavior where it would be confusing/unclear but 🙃

nssp/tests/test_patch.py
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        patch()
+
+        start_date = datetime(2021, 1, 1)
+        end_date = datetime(2021, 1, 16)
+        date = start_date
+
+        # Only Sundays should be issue dirs.
+        while date <= end_date:
```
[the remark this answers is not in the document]
minhkhul: [7078dd0](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/commits/7078dd04efddf254a87e1892482af4c4af49eea3)

nssp/delphi_nssp/run.py
minhkhul: no reason. When I read your suggestion to add the custom_run flag, I just thought of the patch use case for it, so I wrote the flag as that. I would imagine if we do other kinds of custom run in the future, we can change it then.

nssp/delphi_nssp/run.py
nmdefries: I don't particularly care how we handle this and [the `custom_flag`](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/files#r1718538840), but this logic should be the same between all of our patch modules. Please consult how we do this in the `google-symptoms` and `doctor-visits` patchers.

nmdefries (commented):
Initial comments

nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,139 @@
...
+    ...
+  },
+  "validation": {
+    ...
+  },
+  "patch": {
+    "patch_dir": "/Users/minhkhuele/Desktop/delphi/covidcast-indicators/nssp/AprilPatch",
+    "start_issue": "2024-04-20",
+    "end_issue": "2024-04-21",
+    "source_dir": "/Users/minhkhuele/Desktop/delphi/covidcast-indicators/nssp/source_data"
```
nmdefries, line 19: question: what is the `source_dir` used for? Isn't NSSP data fetched from an API?

nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,139 @@
...
+                valid_config = False
+
+    if valid_config:
+        logger.info("Good patch configuration.")
+        return True
+    logger.info("Bad patch configuration.")
+    return False
+
+
+def patch():
```
nmdefries, line 84: thought: it seems like would be easy to turn this into a shared fn in utils. We would just need to pass it the package name, maybe the source name, and the package's specific `run_module` fn. Would make it easier to roll out. The issue-handling happens in `run_module` and data-fetch functions, so I think this could all stay as-is.

nssp/delphi_nssp/run.py
```diff
@@ -72,18 +70,28 @@ def run_module(params):
...
+            logger.warning("custom_run flag is on despite direct indicator run call. Normal indicator run continues.")
```
nmdefries, line 82: question: is this the default behavior we want? How does this work in the other existing patchers? Maybe we should just raise an error so, e.g., the user doesn't accidentally use normal output as patch data.

nssp/delphi_nssp/pull.py
```diff
@@ -39,26 +38,38 @@ def pull_nssp_data(socrata_token: str):
...
+        while True:
+            page = client.get("rdmq-nq56", limit=limit, offset=offset)
+            if not page:
+                break  # exit the loop if no more results
+            results.extend(page)
+            offset += limit
+        df_ervisits = pd.DataFrame.from_records(results)
+        logger.info(f"Grabbed {len(df_ervisits)} records from Socrata API")
+    else:
+        df_ervisits = pd.read_csv(f"{source_dir}/{issue_date}.csv")
```
nmdefries, line 71: question: ah, we create a local copy of the data set using the "backup" script provided in the README, and then read a single issue-specific CSV from that archive? A few questions/comments

- Is it possible to include the archive-generation in the code here? We could even turn the "backup" script into a fn and call it as part of the patching module. I want to avoid having patching be a multi-step process. So I want to make sure the person patching doesn't need to remember to run the backup code first -- it should be automatic.
- I don't understand how the backup code is generating issue-specific datasets. Or are we doing post-processing somewhere?

minhkhul:
@nmdefries 
I realized how confusing I wrote the readme to be. So I just made some major changes to that, then added code to auto-download the source backup files [here]([4356494](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/commits/4356494246e2defc1c4a5118bcf2eddce8ca8028)). With this, patcher will only need to adjust params then run `env/bin/python -m delphi_nssp.patch` to create patch data. 

None of any other extra steps. 

The code chunk in the readme was not generating issue-specific dataset all at once, but just for one issue. I use that code and daily scheduling to made a daily copy of what the source API returns and store that copy as in a csv on `bigchunk-dev-02.delphi.cmu.edu`. So we have a history since march for this indicator. This means on the server there are these daily csv where each csv is source data for one issue:
```
[user@bigchunk-dev-02 nssp]$ pwd
/common/source_backup/nssp
[user@bigchunk-dev-02 nssp]$ ls
2024-04-17.csv  2024-05-21.csv  2024-06-25.csv  2024-08-03.csv
2024-04-18.csv  2024-05-22.csv  2024-06-26.csv  2024-08-04.csv
2024-04-19.csv  2024-05-23.csv  2024-06-27.csv  2024-08-05.csv
...
```
I originally put the code chunk in readme to show how if there'a a patching need, one can set up their own daily/weekly backup starting from the start of an outage to the end. But it's really not needed now that I made the relevant csv files on bigchunk-dev-02.delphi.cmu.edu all readable to anyone with access to that server. 

Then I added the auto-download source backup csv files.

So now we avoid having patching be a multi-step process.

nssp/delphi_nssp/run.py
minhkhul: @nmdefries Existing patcher in google symptom doesn't handle this case yet. 

My thought process when I wrote this is, the accidental use of normal output as patch data won't happen because

Case 1: intention is to do normal run, just forgot to turn off flag:
Since user call the normal run (`env/bin/python -m delphi_nssp`), chances are they do want to do normal run and they just forgot to turn the custom flag off. 

Case 2: intention is to do patch run, just mistakenly trigger normal run:
If someone want to patch things but accidentally call normal run, they would find no batch issue format dirs or csv created in the patch_dir if we handle it this way. They will then check log and find their mistake. This is because output of the run under this circumstance will go into normal receiving dir instead of patch dir. 
If they stubbornly try to ingest these normal run result into our database, without the correct dir format created by correct patch run call, patch acquisition will fail anyway.

lmk if I should just error it out and halt.

nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,195 @@
...
+    Run the nssp indicator for a range of issue dates.
+
+    The range of issue dates is specified in params.json using the following keys:
+    - "patch": Only used for patching data
+        - "source_backup_credentials": (Optional)
+        Add this object to download source data from a backup server to source_dir.
+        Remove to assume source csv are already in source_dir.
+            - "host": str, hostname of the server where source data is backed up
+            - "user": str, username to log in to the server
+            - "path": str, path to the directory containing backup csv files
```
minhkhul, line 143: I keep this optional because downloading stashed source files take _awhile_, so if for some reason you're re-runing patch for the same period again and again, removing this config chunk would save time.

nssp/delphi_nssp/run.py
aysim319: I made the option to let custom run be a possible behavior if someone want to run with custom dates that's not patching. If that's not a really a use case that needs to be considered I can also fix things on the other indicators to match this behavior

nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,195 @@
...
+
+import pandas as pd
+import paramiko
+from delphi_utils import get_structured_logger, read_params
+from epiweeks import Week
+
+from .run import run_module
+
+
+def get_source_data(params, logger):
```
aysim319, line 44: I'm not quite sure what's the purpose of this function here. And furthermore, it doesn't seem like it's cleaned up at the end. Do we need to need to keep the csv?

nssp/delphi_nssp/patch.py
aysim319: Probably a good idea to check if the files/directory is not empty at the beginning of the get_source_data. Then even if it was called, it won't make the calls over and over agin

nssp/delphi_nssp/pull.py
aysim319: following up with that how big is the csv? Do we see any issues opening/processing into a dataframe?

nssp/delphi_nssp/run.py
aysim319: Also the case that Nat was mentioning of running normal when wanting to do patch instead shouldn't be a concern since you have to run as `python -m delphi_indicator.patch` specifically in order to do patches and the output directory would be an indicator that you didn't run the right code

nssp/delphi_nssp/patch.py
minhkhul: yup. It's intentional to keep the csv. 
The purpose of the function is to automate grabbing source data. Before adding this, whoever does patching has to scp the source files into source_dir before running patch.py.

nssp/delphi_nssp/pull.py
```diff
@@ -39,26 +38,40 @@ def pull_nssp_data(socrata_token: str):
...
+        limit = 50000  # maximum limit allowed by SODA 2.0
+        while True:
+            page = client.get("rdmq-nq56", limit=limit, offset=offset)
+            if not page:
+                break  # exit the loop if no more results
+            results.extend(page)
+            offset += limit
+        df_ervisits = pd.DataFrame.from_records(results)
+        logger.info("Number of records grabbed from Socrata API", num_records=len(df_ervisits), source="Socrata API")
+    elif custom_run and logger.name == "delphi_nssp.patch":
```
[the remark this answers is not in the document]
minhkhul: yep. It will work. As long as the logger object passed there is originaly created in patch.py, it will work. So technically can move this to pull.py

nssp/delphi_nssp/patch.py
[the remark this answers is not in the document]
minhkhul: nssp pull.py doesn't use paramiko like some other indicators btw, because for normal runs, nssp grabs source data from this soda API rather than a server somewhere. But I agree get_source_data pulls data so it should be in pull.py.

nssp/delphi_nssp/patch.py
minhkhul: fixed at [7e06f94](https://github.com/cmu-delphi/covidcast-indicators/pull/2000/commits/7e06f94d7b6a9668ad8e9843e6838f5c8c369a63).

nssp/delphi_nssp/patch.py
aysim319: Hm... How often do we have to re run patches like so? I understand it's time consuming to pull, but I'm not quite convinced that we would need to keep them after the run. I feel like it's going to get cluttered and messy.

minhkhul: That's super fair. Imma add something to delete them after each run then.

minhkhul: added.

nssp/delphi_nssp/run.py
aysim319: Since it seems like people use custom start and end date can we change the logic to more align with the behavior that google symptom has (custom flag is on, but on regular run) 

For custom regular run, we could also log the start and end date and like a messaging saying custom run

nssp/delphi_nssp/pull.py
minhkhul: Each csv is just under 60mb, so no issue so far.

minhkhul:
Note: About custom_run flag, non-patch custom run doesn't really exist in nssp like it does in google-symptoms due to the differences in how the two sources handle revision data and how we grab those data. We'll keep the flag in nssp still to maintain consistency and to disambiguate params.json.

nssp/delphi_nssp/pull.py
```diff
@@ -1,14 +1,60 @@
...
+        try:
+            local_file_path = path.join(params["patch"]["source_dir"], remote_file_name)
+            sftp.stat(remote_file_name)
+            sftp.get(remote_file_name, local_file_path)
+            num_files_transferred += 1
+        except IOError:
+            logger.warning(
+                "Source backup for this date does not exist on the remote server.", missing_filename=remote_file_name
+            )
+    sftp.close()
```
aysim319, line 51: I don't think it needs to be closed like that, I checked simliar code and it seems that only the paramiko client needs to be closed.

nssp/delphi_nssp/pull.py
```diff
@@ -1,14 +1,60 @@
...
+    csv_file_names = [date.strftime("%Y-%m-%d") + ".csv" for date in dates]
+
+    # Download source files
+    sftp = ssh.open_sftp()
+    sftp.chdir(params["patch"]["source_backup_credentials"]["path"])
+    num_files_transferred = 0
+    for remote_file_name in csv_file_names:
+        try:
+            local_file_path = path.join(params["patch"]["source_dir"], remote_file_name)
+            sftp.stat(remote_file_name)
```
aysim319, line 44: What's the stat command for?

nssp/delphi_nssp/pull.py
```diff
@@ -1,14 +1,60 @@
...
+            sftp.get(remote_file_name, local_file_path)
```
aysim319, line 45: Can we change this call so it also logs the progress similar to the other downloads

https://github.com/cmu-delphi/covidcast-indicators/blob/d52dcd5a6b84083fb5cbca4164ebdd8429ac81d3/claims_hosp/delphi_claims_hosp/download_claims_ftp_files.py#L100-L102

nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,155 @@
...
+        current_issue_date = Week.fromdate(current_issue).startdate()
+        current_issue_dir = f"""{params["patch"]["patch_dir"]}/issue_{current_issue_date.strftime("%Y%m%d")}/nssp"""
+        makedirs(f"{current_issue_dir}", exist_ok=True)
+        params["common"]["export_dir"] = f"""{current_issue_dir}"""
+
+        run_module(params, logger)
+        current_issue += timedelta(days=1)
+
+    if downloaded_source:
+        rmtree(source_dir)
```
aysim319, line 151: Add logging saying it's removing the files

nssp/delphi_nssp/patch.py
aysim319: Yup, that's the general idea of patches and generally how I would do patches and such with airflow...but there needs some work to be done in the run modules to make it possible.

nssp/delphi_nssp/patch.py
minhkhul: nssp data is fetched from an API when it is run normally. The API does not have versioning, so we keep a daily stash of what the api returns every day on a separate server, which is used here as source data for issue dates in patching.

nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,155 @@
...
+            current_issue += timedelta(days=1)
+            continue
+
+        params["patch"]["current_issue"] = current_issue.strftime("%Y-%m-%d")
+
+        # current_issue_date can be different from params["patch"]["current_issue"]
+        # due to weekly cadence of nssp data. For weekly sources, issue dates in our
+        # db matches with first date of epiweek that the reporting date falls in,
+        # rather than reporting date itself.
+        current_issue_date = Week.fromdate(current_issue).startdate()
```
nmdefries, line 142: issue: Since our `current_issue` `while` loop goes day-by-day, we will get the same `current_issue_date` for 7 `current_issue`s in a row (Sunday-Saturday of the week starting with a given `current_issue_date`). That means that the `run_module` call below will generate the same data.

This can fixed by keeping track of which `current_issue_date`s we've seen before (e.g. in a set) and skipping loop iterations/`current_issue`s where the corresponding `current_issue_date` is in the set.

nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,155 @@
...
+
+    logger.info(start_issue=start_issue.strftime("%Y-%m-%d"))
+    logger.info(end_issue=end_issue.strftime("%Y-%m-%d"))
+    logger.info(source_dir=source_dir)
+    logger.info(patch_dir=params["patch"]["patch_dir"])
+
+    makedirs(params["patch"]["patch_dir"], exist_ok=True)
+
+    current_issue = start_issue
+    while current_issue <= end_issue:
```
nmdefries, line 127: nit (optional): I have a slight preference to loop over a date range like `dates = pd.date_range(start=params["patch"]["start_issue"], end=params["patch"]["end_issue"])` rather than using a while loop.

nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,155 @@
...
+    params = read_params()
+    logger = get_structured_logger("delphi_nssp.patch", filename=params["common"]["log_filename"])
+    if not good_patch_config(params, logger):
+        sys.exit(1)
+
+    source_dir = params["patch"]["source_dir"]
+    downloaded_source = False
+    if not path.isdir(source_dir) or not listdir(source_dir):
+        get_source_data(params, logger)
+        downloaded_source = True
```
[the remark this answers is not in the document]
nmdefries: Ah, this actually downloads the full `source_dir` from the server? I saw the definition of `get_source_data` below.

nmdefries (commented):
One issue with generating duplicate data, and a couple clarifying questions.

nssp/tests/test_patch.py
```diff
@@ -0,0 +1,162 @@
...
+                           "20210115", "20210116"]
+        for date_str in not_issue_dates:
+            assert not os.path.isdir(f'./patch_dir/issue_{date_str}/nssp')
+
+        # Clean up the created directories after the test
+        shutil.rmtree(mock_read_params.return_value["patch"]["patch_dir"])
+
+    @mock_patch('delphi_nssp.patch.get_structured_logger')
+    @mock_patch('delphi_nssp.patch.read_params')
+    def test_patch_confirm_values_generated(self, mock_read_params, mock_get_structured_logger):
```
aysim319, line 130: this test fails in good_patch_config for patch(). Was that intentional? There should be a test for patch that actually goes through the full patch code

nssp/delphi_nssp/patch.py
```diff
@@ -0,0 +1,153 @@
...
+    if not custom_run:
+        logger.error("Calling patch.py without custom_run flag set true.")
+        valid_config = False
+
+    patch_config = params.get("patch", {})
+    if patch_config == {}:
+        logger.error("Custom flag is on, but patch section is missing.")
+        valid_config = False
+    else:
+        required_patch_keys = ["start_issue", "end_issue", "patch_dir", "source_dir", "user"]
```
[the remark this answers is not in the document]
minhkhul: good point. I'll make the change.

nssp/delphi_nssp/pull.py
```diff
@@ -1,20 +1,83 @@
...
+    It uses "backup_dir" config under "common" section to locate backup files on remote server.
+    It then searches for CSV files that match the inclusive range of issue dates
+    specified by 'start_issue', and 'end_issue' config.
 
+    These CSV files are then downloaded and stored in the local 'source_dir' directory.
+    """
+    makedirs(params["patch"]["source_dir"], exist_ok=True)
+    ssh = paramiko.SSHClient()
+    ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+    host = "delphi-master-prod-01.delphi.cmu.edu"
```
nmdefries, line 45: suggestion: Hostnames should probably be stored in `params` for security.

nssp/delphi_nssp/pull.py
```diff
@@ -1,20 +1,83 @@
...
+    # Generate file names of source files to download
+    dates = pd.date_range(start=params["patch"]["start_issue"], end=params["patch"]["end_issue"])
+    remote_source_files = [f"{date.strftime('%Y%m%d')}.csv.gz" for date in dates]
+
+    # Download source files
+    sftp = ssh.open_sftp()
+    try:
+        sftp.stat(params["common"]["backup_dir"])
+    except IOError:
+        logger.error("Source backup directory does not exist on the remote server.")
```
nmdefries, line 58: question: I'm confused about this section and below. I _think_ that "*.stat" checks if a file/dir exists.

So here, we're "stat"ing the whole remote backup dir. Below, we're "stat"ing each individual data file that fits in the issue range. And then `get` does the actual downloading. So we're only downloading data files that fit into the issue range.

Is that correct?

nssp/delphi_nssp/pull.py
```diff
@@ -1,20 +1,83 @@
...
+            sftp.stat(remote_file_name)
+        except IOError:
+            logger.warning(
+                "Source backup for this date does not exist on the remote server.", missing_filename=remote_file_name
+            )
+            continue
+        sftp.get(remote_file_name, local_file_path, callback=callback_for_filename)
+        logger.info("Transfer finished", remote_file_name=remote_file_name, local_file_path=local_file_path)
+        num_files_transferred += 1
+    ssh.close()
```
nmdefries, line 76: suggestion (optional): You could wrap this section in a `with ssh.open_sftp() as sftp:` to avoid this `close()`. It won't change the behavior but it might make maintenance/change less error-prone.

nssp/delphi_nssp/pull.py
```diff
@@ -76,9 +146,25 @@ def pull_nssp_data(socrata_token: str, backup_dir: str, custom_run: bool, logger
...
         Dataframe as described above.
     """
-    socrata_results = pull_with_socrata_api(socrata_token, "rdmq-nq56")
-    df_ervisits = pd.DataFrame.from_records(socrata_results)
-    create_backup_csv(df_ervisits, backup_dir, custom_run, logger=logger)
+    if not custom_run:
+        socrata_results = pull_with_socrata_api(socrata_token, "rdmq-nq56")
+        df_ervisits = pd.DataFrame.from_records(socrata_results)
+        create_backup_csv(df_ervisits, backup_dir, custom_run, logger=logger)
+        logger.info("Number of records grabbed", num_records=len(df_ervisits), source="Socrata API")
```
nmdefries, line 153: nit: this makes it look like `logger` isn't actually optional (e.g. if logger is not provided `None.info()` or `None.warning()` will cause an error). Should we change that arg description?

nmdefries (approved):
This looks good to me. The only required change before merging is moving the hostname to a param and not including it in the actual template.

nssp/delphi_nssp/pull.py
minhkhul: yep.

nssp/delphi_nssp/pull.py
minhkhul: so I didn't change the arg description, but added `if logger:` before each time a logger method is called in `create_backup_csv`. That should help avoid those errors.

aysim319 (approved):

nssp/tests/test_patch.py
```diff
@@ -0,0 +1,216 @@
+import pytest
+from unittest.mock import patch as mock_patch, call, MagicMock
+from delphi_nssp.patch import patch, good_patch_config, get_source_data, get_patch_dates
```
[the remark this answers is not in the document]
minhkhul: unit test for `get_source_data` is in test_pull. 
For sftp vs local files it's in `test_patch_from_local_source` and `test_patch_download_remote_source`

PR #2004, opened 2024-07-26
1972 replace covidcast

aysim319:
### Description
refactored covidcast that under the hood uses a for loop for each day to grab signals

### Changelog
- replaced instances of covidcast.signal and covidcast.metadata with respective epidata api calls

### Associated Issue(s) 
- Addresses #(1972)
- Addresses #(1931)
- Addresses #(1987)

### DIFF
--- a/google_symptoms/tests/conftest.py
+++ b/google_symptoms/tests/conftest.py
@@ -7,7 +7,6 @@ import copy
 import pytest
 import mock
 import pandas as pd
-from freezegun import freeze_time
 
 from os import listdir, remove, makedirs
 from os.path import join, exists


aysim319:
The refactoring takes more time than the original, possibly because the refactored grabs the whole data instead of truncated data, so longer processing time to go through the validator

[cprofile_google_symptoms.txt](https://github.com/user-attachments/files/16432308/cprofile_google_symptoms.txt)
[refactored_cprofile_google_symptoms.txt](https://github.com/user-attachments/files/16432309/refactored_cprofile_google_symptoms.txt)

melange396:
@aysim319 can you explain your previous comment a little more?  what do each of those files represent?  the first two files appear to be telling me that this branch runs a little bit slower (281s) than the dev branch (267s)...  but what operation are they performing (ie, what commands and arguments did you use to run these samples)?

_delphi_utils_python/delphi_utils/covidcast_wrapper.py
```diff
@@ -0,0 +1,251 @@
...
+def metadata() -> Union[pd.DataFrame, None]:
+    """
+    Make covidcast metadata api call.
+
+    Returns
+    -------
+    pd.DataFrame of covidcast metadata.
+    """
+    # pylint: disable=W0212
+    response = Epidata._request("covidcast_meta")
```
melange396, line 59:
```suggestion
    response = Epidata.covidcast_meta()
```
it does ~exactly the same thing, but you might as well use the public method; see https://github.com/cmu-delphi/delphi-epidata/blob/8746ff2ef7a936bb93628bc1358471d7c6c4f5f8/src/client/delphi_epidata.py#L611-L615

_delphi_utils_python/delphi_utils/covidcast_wrapper.py
```diff
@@ -0,0 +1,251 @@
...
+    signal: str,  # pylint: disable=W0621
+    start_day: date = None,
+    end_day: date = None,
+    geo_type: str = "county",
+    geo_values: Union[str, Iterable[str]] = "*",
+    as_of: date = None,
+    lag: int = None,
+    time_type: str = "day",
+) -> Union[pd.DataFrame, None]:
+    """Download a Pandas data frame for one signal.
```
melange396, line 83: This docstring looks like it was copied from https://github.com/cmu-delphi/covidcast/blob/main/Python-packages/covidcast-py/covidcast/covidcast.py ...  I dont think we need this much detail here.  We arent necessarily trying to duplicate the exact functionality of the `covidcast` client, we just wanna replace it with something else that does a similar job but with a simpler interface and less dependency bloat.

_delphi_utils_python/delphi_utils/covidcast_wrapper.py
```diff
@@ -0,0 +1,251 @@
...
+    for more details on how values and standard errors are calculated for
+    specific signals.
+
+    """
+    if start_day > end_day:
+        raise ValueError(
+            "end_day must be on or after start_day, but " f"start_day = '{start_day}', end_day = '{end_day}'"
+        )
+
+    time_values = list(date_generator(start_day, end_day))
```
melange396, line 228: You can get rid of this generator for a few reasons: its not going to work properly for "week"-type data, and its gonna be inefficient...
```suggestion
    time_values = f"{start_day}-{end_day}"
```
I think this suggestion should work without exploding out into a big comma-separated list, which will be slightly gentler on the API server and probably also better for the DB optimizer.

_delphi_utils_python/delphi_utils/covidcast_wrapper.py
```diff
@@ -0,0 +1,251 @@
...
+        time_type=time_type,
+        geo_type=geo_type,
+        time_values=time_values,
+        geo_value=geo_values,
+        as_of=as_of,
+        lag=lag,
+    )
+    if response["result"] != 1:
+        # Something failed in the API and we did not get real metadata
+        raise RuntimeError("Error when fetching metadata from the API", response["message"])
```
melange396, line 242: copypasta!
```suggestion
        raise RuntimeError("Error when fetching signal data from the API", response["message"])
```

_delphi_utils_python/delphi_utils/validator/datafetcher.py
```diff
@@ -159,9 +163,8 @@ def fetch_api_reference(data_source, start_date, end_date, geo_type, signal_type
     Formatting is changed to match that of source data CSVs.
     """
     with warnings.catch_warnings():
-        warnings.simplefilter("ignore")
-        api_df = covidcast.signal(
-            data_source, signal_type, start_date, end_date, geo_type)
+        api_df = covidcast_wrapper.signal(data_source, signal_type, start_date, end_date, geo_type)
```
melange396, line 166: I dont think this context manager will still ignore warnings if you remove the `.simplefilter()` call.  I also suspect that the `delphi-epidata` lib will not produce a warning where the `covidcast` lib did, so you can probably remove the c.m. entirely.  Do you know what the text of the warning from `covidcast.signal()` actually is?  Is it something we need to worry about?

_delphi_utils_python/delphi_utils/validator/dynamic.py
```diff
@@ -79,7 +81,7 @@ def validate(self, all_frames, report):
         outlier_lookbehind = timedelta(days=14)
 
         # Authenticate API
-        covidcast.use_api_key(self.params.api_key)
+        # Epidata.auth = ("epidata", api)
```
melange396, line 84: This can be removed.
```suggestion
```

_delphi_utils_python/tests/test_covidcast_wrapper.py
```diff
@@ -0,0 +1,35 @@
...
+TEST_DIR = Path(__file__).parent
+class TestCovidcastWrapper:
+    def test_metadata(self):
+        expected_df = covidcast.metadata()
+        df = covidcast_wrapper.metadata()
+        assert_frame_equal(expected_df, df)
+
+    @freeze_time("2024-07-29")
+    def test_signal(self):
+        meta_df = covidcast_wrapper.metadata()
```
melange396, line 18: This is making an actual request over the network, but unit tests should be self-contained.  You could move these tests to `integrations/` but heavy use of external resources is discouraged there too...  The `delphi-epidata` client has a [debug mode](https://github.com/cmu-delphi/delphi-epidata/blob/8746ff2ef7a936bb93628bc1358471d7c6c4f5f8/src/client/delphi_epidata.py#L46) where it can tell you exactly what requests its making -- can you list all of the requests made by this test?  That can help give me an idea of how big of a lift this is doing.

_delphi_utils_python/tests/test_covidcast_wrapper.py
```diff
@@ -0,0 +1,35 @@
...
+        data_filter = ((meta_df["max_time"] >= datetime(year=2024, month=6, day=1)) & (meta_df["time_type"] == "day"))
```
melange396, line 19: The way you are filtering this looks like it will include new signals that are added...  `@freeze_time` only applies to the local scope of this test, not what youre gonna get back from the live API!

_delphi_utils_python/tests/test_covidcast_wrapper.py
```diff
@@ -0,0 +1,35 @@
...
+        signal_df = meta_df[data_filter].groupby("data_source")["signal"].agg(['unique'])
+        enddate = datetime.today()
+        startdate = enddate - timedelta(days=15)
+        for data_source, row in signal_df.iterrows():
+            signals = list(row[0])
+            for signal in signals:
+                # expected_df = covidcast.signal(data_source, signal, start_day=startdate, end_day=enddate, geo_type="state")
+                expected_df = pd.read_pickle(f"{TEST_DIR}/test_data/{data_source}_{signal}.pkl")
```
melange396, line 27: The dates you picked for this are kinda recent, and im not sure we can guarantee the associated data wont change.  Maybe try something from a few years ago?

melange396 (changes requested):

_delphi_utils_python/delphi_utils/covidcast_wrapper.py
melange396: You could also try using the `Epidata.range()` method (like [this](https://github.com/cmu-delphi/delphi-epidata/blob/8746ff2ef7a936bb93628bc1358471d7c6c4f5f8/integrations/client/test_delphi_epidata.py#L152-L154)).  Its more verbosity and obfuscation, but its kinda nicer from an object-oriented perspective

aysim319:
> @aysim319 can you explain your previous comment a little more? what do each of those files represent? the first two files appear to be telling me that this branch runs a little bit slower (281s) than the dev branch (267s)... but what operation are they performing (ie, what commands and arguments did you use to run these samples)?


The first time I ran the comparison the profiler was taking more time because of the issues param that I thought I also needed to figure out to format to make the call to epidata. I later found out I didn't need to pass along the issues param which fixed both the speed and the difference between the validator result.

_delphi_utils_python/delphi_utils/covidcast_wrapper.py
aysim319: Ah I missed the public method for metadata for some reason. Making change

_delphi_utils_python/delphi_utils/validator/datafetcher.py
aysim319: the warnings were either no data or too much data:
in def _fetch_epidata():
```
        # Two possible error conditions: no data or too much data.
        if day_data["message"] == "no results":
            warnings.warn(f"No {data_source} {signal} data found on {day_str} "
                          f"for geography '{geo_type}'",
                          NoDataWarning)
        if day_data["message"] not in {"success", "no results"}:
            warnings.warn(f"Problem obtaining {data_source} {signal} data on {day_str} "
                          f"for geography '{geo_type}': {day_data['message']}",
                          RuntimeWarning)
```                          

 if the response for that date was too big, I recall Dmitry mention how back in the day the limit used to be 5k per response and that was the main reason behind the covidcast api having a for loop for each date and truncating the rest or something

_delphi_utils_python/delphi_utils/covidcast_wrapper.py
aysim319: The reason I created a generator is that when I looked into the epidata.covidcast() call it was building a query to call the db and for the time values. I tried to use the something similar and it didn't work.

And I just tried using the Epidata.range and the result didn't work. tried setting the range with str and datetime.

```
from delphi_epidata import Epidata
startdate = datetime(year=2024, month=7, day=19)
enddate = datetime(year=2024, month=7, day=24)
sdate_str = startdate.strftime('%Y-%m-%d')
edate_str = enddate.strftime('%Y-%m-%d')

def date_generator(startdate, enddate):
  while startdate <= enddate:
    yield startdate.strftime('%Y-%m-%d')
    startdate = startdate + timedelta(days=1)

response = Epidata.covidcast(source, signal, time_type="day",
                          geo_type=geo, time_values=Epidata.range(sdate_str, edate_str),
                          geo_value="*", as_of=None, lag=None)
response2 = Epidata.covidcast(source, signal, time_type="day",
                          geo_type=geo, time_values=Epidata.range(startdate, enddate),
                          geo_value="*", as_of=None, lag=None)
response3 = Epidata.covidcast(source, signal, time_type="day",
                          geo_type=geo, time_values=list(date_generator(startdate, enddate)),
                          geo_value="*", as_of=None, lag=None)
```

_delphi_utils_python/delphi_utils/covidcast_wrapper.py
aysim319: oops! caught and fixed

_delphi_utils_python/delphi_utils/validator/dynamic.py
aysim319: removed but are we not registering the call or because it's running in the same environment it won't be rate limited or just it's not something we need to worry about?

_delphi_utils_python/delphi_utils/covidcast_wrapper.py
aysim319: I realized that the Epidata.range was used specifically for issues, but I still need to get the data based on the time value and in epidata the time value needs to be filled out, and from the previous comment the epidata.range or string does not work.

in the example you showed the time_value was set to one date, where we would see rows where the time value is a specific date 2020_02_02 associated with multiple issue_dates.

So this does give me data back:
```
response6 = Epidata.covidcast(source, signal, time_type="day",
                          geo_type=geo,  time_values=sdate_str, issues=Epidata.range("20240629", "20240719"),
                          geo_value="*", as_of=None, lag=None)
```
but I would still need to call for multiple days unless I use the exploded comma dates

dshemetov: The `time_values` param to the `/covidcast` can definitely handle time ranges, for instance

```
https://api.delphi.cmu.edu/epidata/covidcast/?data_source=fb-survey&signals=smoothed_cli&time_type=day&time_values=20210405-20210410&geo_type=state&geo_values=ca
```

 so there's something going wrong client side. May the string isn't formatted correctly?

_delphi_utils_python/tests/test_covidcast_wrapper.py
```diff
@@ -0,0 +1,38 @@
+from datetime import datetime, timedelta
+from pathlib import Path
+import pandas as pd
+from delphi_utils import covidcast_wrapper
+import covidcast
+from freezegun import freeze_time
+from delphi_epidata import Epidata
+from pandas.testing import assert_frame_equal
+
+TEST_DIR = Path(__file__).parent
```
dshemetov, line 10: I don't think these tests should be permanent because (a) they will slow down our tests with expensive network calls and (b) they will not give us new information once finalized (unless covidcast_wrapper code changes, which I expect to be rare). I propose that we think of this as more of a testing script that we use while iterating on this PR; once we convince ourselves that the PR code is correct, we disable it. A simple way to do this is to add this skip decorator

```suggestion
TEST_DIR = Path(__file__).parent
import pytest
# Uncomment this to run these tests, leave commented once done.
# @pytest.mark.skip(reason="Expensive external requests to verify covidcast_wrapper.")
```

_delphi_utils_python/tests/test_covidcast_wrapper.py
```diff
@@ -0,0 +1,38 @@
...
+class TestCovidcastWrapper:
+    Epidata.debug = True
+    def test_metadata(self):
+        expected_df = covidcast.metadata()
+        df = covidcast_wrapper.metadata()
+        assert_frame_equal(expected_df, df)
+
+    @freeze_time("2022-01-29")
+    def test_signal(self):
+        meta_df = pd.read_pickle(f"{TEST_DIR}/test_data/covidcast_metadata.pkl")
```
dshemetov, line 20: - What's the benefit of freeze_time here? It only affects calls like `datetime.today()` and I think you could just use `enddate = datetime(...)` instead. freeze_time makes me think that there's some internal calls using `today()` calls, but I don't think we rely on those anywhere?
- What's the benefit of pickling the metadata here? Since these tests are already making a lot of network calls, I'm leaning towards just "rounding" to making this script self-contained and not reliant on any pre-computed files. The benefit of making it self-contained is that if we ever need to test this wrapper later, we can just comment out the skip at the top, and run this test to compare the wrapper with the actual output of `metadata()` instead of a pre-computed one.

_delphi_utils_python/delphi_utils/covidcast_wrapper.py
aysim319: yup -- the strings aren't formatted correctly. Didn't realized that the string format for the epidata api were %Y%m%d

melange396: If youre using the range notation properly, you shouldnt need the comma-separated list, and you should be able to make just one request.  You could try turning on the client's DEBUG mode to see what its actually requesting.

_delphi_utils_python/delphi_utils/validator/datafetcher.py
melange396: The current limit is 10M rows returned per request, we shouldnt have problems with it, but it is still worth catching the error condition.

_delphi_utils_python/delphi_utils/validator/dynamic.py
melange396: You took care of adding the credentials at https://github.com/cmu-delphi/covidcast-indicators/pull/2004/files#diff-de79eb5faa25bf35238e1d85065a72bf9b7a264f92357b64b44b411423703f64R23

Also, this line was commented out, so it wasnt doing anything anyway.

_delphi_utils_python/tests/test_covidcast_wrapper.py
melange396: TBQH, i dont think the covidcast wrapper is necessary -- in fact, i think its overkill.  getting rid of it entirely will negate any need for this test.

dshemetov: Getting rid of the wrapper seems possible, but I think these tests are still useful for ensuring correctness of this migration. Once we trust the migration, I definitely think we should disable/delete them, though having a test record is nice, whether that takes the form of a link to a commit hash or a disabled unit test, doesn't matter to me.

_delphi_utils_python/tests/test_covidcast_wrapper.py
aysim319: moved as a script in testing_utils, hopefully that's good enough? also if this is something that's going to be one and done, I just used api calls everywhere and made the process quite expensive...should I make it bit less or trying to get the absolute truth?

_delphi_utils_python/delphi_utils/validator/datafetcher.py
```diff
@@ -3,18 +3,45 @@
...
+import requests
+from delphi_epidata import Epidata
+from epiweeks import Week
+
 from .errors import APIDataFetchError, ValidationFailure
 
 FILENAME_REGEX = re.compile(
     r'^(?P<date>\d{8})_(?P<geo_type>\w+?)_(?P<signal>\w+)\.csv$')
 
+def _parse_datetimes(date_int: int, time_type: str, date_format: str = "%Y%m%d") -> Union[pd.Timestamp, None]:
```
aysim319, line 22: we don't need to have this function just yet since the validator only support day time type, but it was in the backlog of eventually supporting week time type, should I keep this or not?

dshemetov: I think we can get rid of it because:
- it's only used by metadata fetching to parse min_time, max_time, and last_update columns from metadata and those columns aren't even used (we only use `signal` and `geo_type` for each source)
- supporting the week time type isn't easy to do and should probably be a separate PR

_delphi_utils_python/delphi_utils/validator/datafetcher.py
```diff
@@ -158,10 +196,30 @@ def fetch_api_reference(data_source, start_date, end_date, geo_type, signal_type
...
+    )
+    if response["result"] != 1:
+        # Something failed in the API and we did not get real metadata
+        raise RuntimeError("Error when fetching signal data from the API", response["message"])
+
+    api_df = None
+    if len(response["epidata"]) > 0:
+        api_df = pd.DataFrame.from_dict(response["epidata"])
+        api_df["issue"] = pd.to_datetime(api_df["issue"], format="%Y%m%d")
+        api_df["time_value"] = pd.to_datetime(api_df["time_value"], format="%Y%m%d")
```
dshemetov, line 219: You said validator only runs on daily signals right? Because this will fail on signals where `time_value` is in epiweek `YYYYWW` format. Should be fine if we don't need to handle those now. Can be adjusted when we expand later.

aysim319: Ah good catch! I'll make a comment for future reference to fix them

testing_utils/check_covidcast_port.py
```diff
@@ -0,0 +1,169 @@
...
+
+def check_metadata():
+    expected_df = covidcast.metadata()
+    df = ported_metadata().metadata()
+    assert_frame_equal(expected_df, df)
+
+def check_signal():
+    meta_df = covidcast.metadata()
+    startdate = datetime(year=2022, month=2, day=1)
+    data_filter = (meta_df["max_time"] >= startdate)
```
dshemetov, line 149: Can you add a check that this yields a non-empty dataframe?

testing_utils/check_covidcast_port.py
```diff
@@ -0,0 +1,169 @@
...
+    signal_df = meta_df[data_filter].groupby("data_source")["signal"].agg(['unique'])
+    enddate = startdate + timedelta(days=3)
```
dshemetov, line 151: Can we make this a longer window? Like 30 days? I want to make sure we're actually getting some data and not empty dataframes downstream.

testing_utils/check_covidcast_port.py
```diff
@@ -0,0 +1,169 @@
...
+
+    for data_source, row in signal_df.iterrows():
+        signals = list(row[0])
+        for signal in signals:
+            expected_df = covidcast.signal(data_source, signal, start_day=startdate, end_day=enddate, geo_type="state")
+            if expected_df is None:
+                print("%s %s %s %s not existing", data_source, signal, startdate, enddate)
+
+            df = ported_signal(data_source, signal, start_day=startdate, end_day=enddate, geo_type="state")
+
```
dshemetov, line 161: Can we add a check that both these dataframes are not empty?

testing_utils/check_covidcast_port.py
```diff
@@ -0,0 +1,169 @@
+"""
+script to check converting covidcast api calls with Epidata.covidcast Epidata.covidcast_meta
+"""
```
dshemetov, line 3: Do these tests pass locally for you? How long do they take to run?

google_symptoms/delphi_google_symptoms/run.py
```diff
@@ -66,8 +62,7 @@ def run_module(params):
         )
 
         # Fetch metadata to check how recent each signal is
-        covidcast.use_api_key(params["indicator"]["api_credentials"])
```
dshemetov, line 69: I think we need Epidata.auth to be set here now.

dshemetov (changes requested):
Want to make sure the tests we have are solid, so made some a few change requests.

testing_utils/check_covidcast_port.py
aysim319: I'm trying to be through since sircomplainalot does take into account for all the signals, it's been some amount of minutes (mostly on the covidcast because of the for loop)

I'm working on grabbing like a few days from their respective min time

_delphi_utils_python/tests/test_covidcast_wrapper.py
dshemetov: The script looks pretty good to me. I think we shouldn't worry about how expensive it is while testing and try to get a decent amount of data and compare it all. I'd rather spend an hour or two on a testing script run once than to spend hours deploying and reverting.

google_symptoms/delphi_google_symptoms/run.py
```diff
@@ -60,7 +60,7 @@ def run_module(params):
             "_".join([metric, smoother, "search"])
             for metric, smoother in product(COMBINED_METRIC, SMOOTHERS)
         )
-
+        Epidata.auth = ("epidata", params["api_credentials"])
```
dshemetov, line 63: This should be `params["indicator"]["api_credentials"]`.

aysim319:
I ran the first half where it's calling with covidcast api and then saved the resulted into parquet then ran the whole thing to get the logs
[covidcast_signal.log](https://github.com/user-attachments/files/16555282/covidcast_signal.log)
[epidata_signal.log](https://github.com/user-attachments/files/16555283/epidata_signal.log)

testing_utils/check_covidcast_port.py
aysim319: the covidcast one takes about takes about an hour to run and epidata take about 30 minutes, so toghet about 1.5 hours

dshemetov (commented):
As far as correctness goes, this is looking good to me. I tested locally and the new port is parsing things identically to the old covidcast function (FWIW, the client logs aren't that important here, we just really want to make sure the DataFrame outputs from our API calls are absolutely identical). 

I convinced myself that we don't need to test every signal: the covidcast schema is the same for every signal, so testing a single signal API query gets you 99% coverage (as long as the query returns a representative amount data subset). The only snag is that time_value can be in two different formats (date or epiweek), so as long as we test NCHS along with the other sources, we get full coverage. I tweaked the test to test a single signal per source, so it runs much faster now (thank you for doing the comprehensive runs nonetheless, it's nice to have that extra safety!).

I also found and fixed some anti-patterns in the covidcast code you ported over, specifically `_parse_datetimes`. Should be a bit faster.

TODO:
- [ ] fix conflicts
- [ ] test with CI (for some reason CI isn't running?)

sir_complainsalot/delphi_sir_complainsalot/check_source.py
```diff
@@ -64,27 +66,38 @@ def check_source(data_source, meta, params, grace, logger):  # pylint: disable=t
...
+            geo_type=row["geo_type"],
+            time_values=Epidata.range(_date_to_api_string(start_date), _date_to_api_string(end_date)),
+            geo_value="*",
         )
 
         current_lag_in_days = (now - row["max_time"]).days
         lag_calculated_from_api = False
+        latest_data = None
+
+        if response["result"] == 1:
```
aysim319, line 94: capture when message is not successful or empty

_delphi_utils_python/delphi_utils/validator/datafetcher.py
```diff
@@ -115,7 +117,20 @@ def get_geo_signal_combos(data_source, api_key):
...
     source_signal_mappings = {i['source']:i['db_source'] for i in
         meta_response.json()}
-    meta = covidcast.metadata()
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+
+        response = Epidata.covidcast_meta()
+
+        if response["result"] != 1:
```
aysim319, line 126: @melange396 would this do the trick? or still add the conditional anyway?

aysim319:
Continuation from last comment https://github.com/cmu-delphi/covidcast-indicators/pull/2056

PR #2005, opened 2024-07-29
Refactor Jenkins pipeline

korlaxxalrok:
### Description
Bug fix and feature

(This Slack [thread](https://delphi-org.slack.com/archives/C013AH5N01E/p1721845491120579) is the starting point for this work.)

**Main change:**

Make the pipeline less likely to produce hard to debug errors by removing usage of `parallel`. Though not 100% diagnosed, it _seems_ like parallel can lead to some strange behavior that manifests as files and programs not being found, or Python package installation steps failing, etc. These are transient and often go away after a retry, but they are confusing and sometimes send people on unnecessary troubleshooting exercises.

Ex issue:

- [Failure](https://jenkins-ci-prod-01.delphi.cmu.edu/blue/organizations/jenkins/covidcast-indicators/detail/main/567/pipeline)
- [Retry](https://jenkins-ci-prod-01.delphi.cmu.edu/blue/organizations/jenkins/covidcast-indicators/detail/main/568/pipeline)

A potentially significant downside will be the extra time the pipeline will take to complete. This won't be too bad for now, but time will scale with the number of indicators. There will be additional work done to migrate the pipeline to GH Actions in the short to medium term, where we should be able to run in parallel again.

**Additional change(s):**

Run the build script on dev/feature branches. This is useful to catch issues before the necessity of merging to `main`.

### Changelog

- Remove usage of parallel in build and package stages
- Create separate build and package scripts
- Add build stage for dev/feature branches, eg anything other than main and prod


melange396 (approved):
The deploy scripts are safe to keep `parallel`?

Do you wanna also remove the old https://github.com/cmu-delphi/covidcast-indicators/blob/main/jenkins/build-and-package.sh script?

Otherwise, LGTM!

korlaxxalrok:
Yeah, re deploy stages: I'd considered changing them as well, but thought we could see how they fare. They _do_ have some infrequent issues, but they are typically minor and iirc always resolve on a retry, so maybe we keep 'em? I'll keep an eye out for failure. The build stage is the one that, at least in the context of this latest issue we ran into, would not resolve on any subsequent retry and is most necessary to address to unblock work.

I'll yank out that original script as well 👍

minhkhul (approved):
To sum up, we're checking env and installations on all branches now, not just main and prod. Got it.

PR #2007, opened 2024-07-30
`google-symptoms` patching fn should not return data for the 4 days immediately before the issue date

nmdefries:
### Description
Plus simplify some tests. Add in "gold standard" data from our actual API to verify patching behavior. Some formatting cleanup.


aysim319:
line 14 the parameter return changed to Dict[date, Dict[str, Union[date, int]] date_utils
I'll change it after the merge

aysim319 (approved):
I was so caught up with the "replicating" the logic and made it more complicated; also I should've asked about the lag; felt like I thought about it for a second and then forgot and spent waaay too much time trying to replicate the date range

PR #2001, opened 2024-07-24
Revert "Clean up delphi_utils "

aysim319:
Reverts cmu-delphi/covidcast-indicators#1984
bugs in the jenkin builds :(

PR #1984, opened 2024-07-01
Clean up delphi_utils 

aysim319:
### Description
migrating setup.py to pyproject.toml and cleaning up dependencies

### Changelog
Itemize code/test/documentation changes and files added/removed.
- added pyproject.toml

### Fixes 
- Fixes #1973

Issue #1973, opened 2024-06-07
Clean up delphi_utils


melange396:
`delphi_utils` is a big hodge-podge of a package with a lot of dependencies, thus it has a relatively large net footprint and takes a while to install.  Lets see if we can lessen that.

Move [packages in setup.py](https://github.com/cmu-delphi/covidcast-indicators/blob/a6ea00355d79a47cc09db8090f6e5121c718de0f/_delphi_utils_python/setup.py#L8-L26) that are unnecessary for a downstream user but still used in test/lint/ci, like: `darker`, `freezegun`, `mock`, `moto`, `pydocstyle`, `pylint`, `pytest-cov`, `pytest`, `requests-mock`.  They can be put into something like a "`requirements.test.txt`" file that gets installed via [the CI](https://github.com/cmu-delphi/covidcast-indicators/blob/a6ea00355d79a47cc09db8090f6e5121c718de0f/.github/workflows/python-ci.yml#L53-L70) and/or [Makefile](https://github.com/cmu-delphi/covidcast-indicators/blob/a6ea00355d79a47cc09db8090f6e5121c718de0f/_delphi_utils_python/Makefile#L11-L25).  

Remove from `setup.py` any unused packages, as `gitpython` appears to be.

The `covidcast` package has its own [wacky imports](https://github.com/cmu-delphi/covidcast/blob/main/Python-packages/covidcast-py/setup.py#L23-L33), which includes a bunch of imaging/graphics and geospatial libraries.  It is only used in the `delphi_utils.validator` module for pulling data from the API and can be replaced with the lighter-weight [`delphi_epidata` client](https://pypi.org/project/delphi-epidata/) instead.  Doing that is not quite as easy as moving requirements entries around, and it is covered in in #1972.

melange396:
We could even put some of the esoteric usage and/or larger installation dependencies behind [package "extras"](https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-extras).

dshemetov:
If we combine this with migrating over to `pyproject.toml`, we could use the [`optional-dependencies`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-optional-dependencies) interface to separate out the `dev` dependencies. See [here](https://github.com/cmu-delphi/delphi-epidata/pull/1470) for an example of migrating over to `pyproject.toml`.

melange396:
[...]

PR #2014, opened 2024-08-05
Clean up delphi_utils (take 2)

melange396:
### Description
migrating setup.py to pyproject.toml and cleaning up dependencies

### Changelog
- added pyproject.toml

### Associated Issue(s) 
- Addresses #1973 

(this is a redo of https://github.com/cmu-delphi/covidcast-indicators/pull/1984, which was reverted in #2001 due to problems with Jenkins builds)


changehc/Makefile
```diff
@@ -13,7 +13,7 @@ install: venv
 install-ci: venv
 	. env/bin/activate; \
 	pip install wheel ; \
-	pip install ../_delphi_utils_python ;\
+	pip install '../_delphi_utils_python[dev]' ;\
```
dshemetov, line 16:
```suggestion
	pip install ../_delphi_utils_python ;\
```

claims_hosp/Makefile
```diff
@@ -13,7 +13,7 @@ install: venv
 install-ci: venv
 	. env/bin/activate; \
 	pip install wheel ; \
-	pip install ../_delphi_utils_python ;\
+	pip install '../_delphi_utils_python[dev]' ;\
```
dshemetov, line 16:
```suggestion
	pip install ../_delphi_utils_python ;\
```

doctor_visits/Makefile
```diff
@@ -13,7 +13,7 @@ install: venv
 install-ci: venv
 	. env/bin/activate; \
 	pip install wheel ; \
-	pip install ../_delphi_utils_python ;\
+	pip install '../_delphi_utils_python[dev]' ;\
```
dshemetov, line 16:
```suggestion
	pip install ../_delphi_utils_python ;\
```

google_symptoms/Makefile
```diff
@@ -13,7 +13,7 @@ install: venv
 install-ci: venv
 	. env/bin/activate; \
 	pip install wheel ; \
-	pip install ../_delphi_utils_python ;\
+	pip install '../_delphi_utils_python[dev]' ;\
```
dshemetov, line 16:
```suggestion
	pip install ../_delphi_utils_python ;\
```

hhs_hosp/Makefile
```diff
@@ -13,7 +13,7 @@ install: venv
 install-ci: venv
 	. env/bin/activate; \
 	pip install wheel ; \
-	pip install ../_delphi_utils_python ;\
+	pip install '../_delphi_utils_python[dev]' ;\
```
dshemetov, line 16:
```suggestion
	pip install ../_delphi_utils_python ;\
```

nchs_mortality/Makefile
```diff
@@ -13,7 +13,7 @@ install: venv
 install-ci: venv
 	. env/bin/activate; \
 	pip install wheel ; \
-	pip install ../_delphi_utils_python ;\
+	pip install '../_delphi_utils_python[dev]' ;\
```
dshemetov, line 16:
```suggestion
	pip install ../_delphi_utils_python ;\
```

nssp/Makefile
```diff
@@ -13,7 +13,7 @@ install: venv
 install-ci: venv
 	. env/bin/activate; \
 	pip install wheel ; \
-	pip install ../_delphi_utils_python ;\
+	pip install '../_delphi_utils_python[dev]' ;\
```
dshemetov, line 16:
```suggestion
	pip install ../_delphi_utils_python ;\
```

nwss_wastewater/Makefile
```diff
@@ -13,7 +13,7 @@ install: venv
 install-ci: venv
 	. env/bin/activate; \
 	pip install wheel ; \
-	pip install ../_delphi_utils_python ;\
+	pip install '../_delphi_utils_python[dev]' ;\
```
dshemetov, line 16:
```suggestion
	pip install ../_delphi_utils_python ;\
```

quidel_covidtest/Makefile
```diff
@@ -13,7 +13,7 @@ install: venv
 install-ci: venv
 	. env/bin/activate; \
 	pip install wheel ; \
-	pip install ../_delphi_utils_python ;\
+	pip install '../_delphi_utils_python[dev]' ;\
```
dshemetov, line 16:
```suggestion
	pip install ../_delphi_utils_python ;\
```

quidel_covidtest/setup.py
```diff
@@ -2,6 +2,7 @@
 from setuptools import find_packages
 
 required = [
+    "boto3",
```
dshemetov, line 5:
```suggestion
	pip install ../_delphi_utils_python ;\
```

sir_complainsalot/Makefile
```diff
@@ -13,7 +13,7 @@ install: venv
 install-ci: venv
 	. env/bin/activate; \
 	pip install wheel ; \
-	pip install ../_delphi_utils_python ;\
+	pip install '../_delphi_utils_python[dev]' ;\
```
dshemetov, line 16:
```suggestion
	pip install ../_delphi_utils_python ;\
```

dshemetov (changes requested):
I didn't catch in the previous PR that the individual indicators don't actually need the dev version of the delphi utils, so made those suggestions. If the indicators have dependency issues because of these changes, we should add the dependencies to the indicator's dependency list (seems like good form not to have packages depend on other packages dev dependencies).

_delphi_utils_python/pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+    "cvxpy",
+    "epiweeks",
+    "gitpython",
+    "importlib_resources>=1.3",
+    "numpy",
+    "pandas>=1.1.0",
+    "slackclient",
+    "scs<3.2.6", # TODO: remove this ; it is a cvxpy dependency, and the excluded version appears to break our jenkins build. see: https://github.com/cvxgrp/scs/issues/283
+    "structlog",
+    "xlrd",
```
dshemetov, line 30:
```suggestion
    "pandas[excel]>=1.1.0",
    "requests",
    "slackclient",
    "scs<3.2.6", # TODO: remove this ; it is a cvxpy dependency, and the excluded version appears to break our jenkins build. see: https://github.com/cvxgrp/scs/issues/283
    "structlog",
```
Suggestion from [comment here](https://github.com/cmu-delphi/covidcast-indicators/issues/1973#issuecomment-2272154545).

google_symptoms/Makefile
aysim319: Has an interesting error; when I don't do the [dev] it's saying that covidcast doesn't have usi_api_key as a module???

`delphi_google_symptoms/date_utils.py:77:4: E1101: Module 'covidcast' has no 'use_api_key' member (no-member)`

https://github.com/cmu-delphi/covidcast-indicators/actions/runs/10285726380/job/28464696528

sir_complainsalot/Makefile
aysim319: Here too: the linter fails when I took the 'dev' off

https://github.com/cmu-delphi/covidcast-indicators/actions/runs/10285993553/job/28465604873?pr=2014

aysim319:
> I didn't catch in the previous PR that the individual indicators don't actually need the dev version of the delphi utils, so made those suggestions. If the indicators have dependency issues because of these changes, we should add the dependencies to the indicator's dependency list (seems like good form to isolate dev dependencies).

most of these need for mock, the only other package that has mock as a requirement in setup.py is google symptoms.

quidel_covidtest/setup.py
aysim319: ? is this a typo?

aysim319:
Do we also want to convert the rest of the indicator to pyproject.toml? maybe putting in the template_python to starting thing off?

dshemetov:
> most of these need for mock

Right, I'm saying we should add mock to each indicator's individual setup.py instead of grabbing it from _delphi_utils' dev dependencies. If you feel up for migrating the indicators to pyproject.toml, then each indicator can have its own "dev" set of packages.

Also, please add `requests` to the _delphi_utils `pyproject.toml`, it was in my suggestion above.

quidel_covidtest/setup.py
dshemetov: ah yea, my bad

aysim319:
Oh I didn't see it before 🤦‍♀️Thanks for pointing it out

dshemetov:
FYI, [this Jenkins script](https://github.com/cmu-delphi/covidcast-indicators/blob/main/jenkins/build-indicator.sh) will need to be adjusted to check for the presence of `pyproject.toml` instead.

aysim319:
> > most of these need for mock
> 
> Right, I'm saying we should add mock to each indicator's individual setup.py instead of grabbing it from _delphi_utils' dev dependencies. Sounds good on migrating the indicators to pyproject.toml, then each indicator can have its own "dev" set of packages and we can put `mock` in there.

Okay just making sure I'm understanding this properly; do we want to convert the existing indicators also to pyproject.toml with the same setup as delphi_utils that has an optional requirement in this PR?

dshemetov:
I don't have a strong preference. Getting delphi_utils cleaned is the key task here and this PR is close to taking care of that (I still think we should remove the ".[dev]" from the individual indicators and debug this CI). Porting the indicators to `pyproject.toml` is nice to have for consistency, but not critical.

dshemetov (approved):
This is good to go. Let's make an issue to migrate the indicators to pyproject.toml and link to this thread. Let's remember to squash merge, so the commit log gets cleaned.

PR #2015, opened 2024-08-09
Fix new date for google symptoms

aysim319:
### Description
When validating the patch, there was an issue where the lag would be taken for custom runs where it would not be applicable
and some clean ups
### Changelog
- added test data that would properly check the signal date ranges
- conditionally adding the extra lag
- minor cleanup mentioned in PR for 1996

### Associated Issue(s) 
- Addresses #(1996) bug


google_symptoms/tests/conftest.py
```diff
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
 import logging
 from pathlib import Path
+import re
 
 import copy
 import pytest
 import mock
 import pandas as pd
+from freezegun import freeze_time
```
nmdefries, line 10: issue: I don't think you're actually freezing time in this file.

google_symptoms/tests/conftest.py
```diff
@@ -38,17 +40,21 @@
 #     end as open_covid_region_code,
 #     *
 # from `bigquery-public-data.covid19_symptom_search.counties_daily_2020` # Counties by day; includes state and county name, + FIPS code
-# where timestamp(date) between timestamp("2020-07-26") and timestamp("2020-08-11")
+# where timestamp(date) between timestamp("2020-07-15") and timestamp("2020-08-22")
+
+# good_input = {
+#     "state": f"{TEST_DIR}/test_data/small_states_daily.csv",
+#     "county": f"{TEST_DIR}/test_data/small_counties_daily.csv"
+# }
```
nmdefries, line 48: suggestion: remove commented code

```suggestion
```

nmdefries (commented):
One question/issue in a test

nmdefries (approved):

PR #2019, opened 2024-08-09
build+ci: make dev envs identical with gh build envs

dshemetov:
### Description
I noticed that the GH actions were throwing extra pip install commands in, which would install either redundant packages or packages necessary to building and publishing. I removed that command and made sure that the necessary packages are accounted for in pyproject.toml so the `make install` and `make install-ci` commands should be sufficient.

### Changelog

* make sure `make venv` upgrades pip right after creating the env
* removed redundant install of pylint pytest pydocstyle
* install twine in [dev] section of _delphi_utils dependencies
* install build in [dev] section of _delphi_utils dependencies
* only needed wheel for bdist_wheel command, which setuptools>=70.1 provides, so remove wheel (see [historical note](https://github.com/pypa/wheel?tab=readme-ov-file#historical-note))

### Associated Issue(s) 
- Addresses #1973


aysim319:
@minhkhul wasn't this somewhat solved by rebuilding the virtualenv in croncile

_delphi_utils_python/pyproject.toml
```diff
@@ -0,0 +1,59 @@
...
+version = "0.3.24"
+description = "Shared Utility Functions for Indicators"
+readme = "README.md"
+requires-python = "== 3.8.*"
+license = { text = "MIT License" }
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Programming Language :: Python :: 3.8",
+    "License :: MIT",
```
aysim319, line 16: the curse of my past that caused error in the release back in september "License :: OSI Approved :: MIT License"

aysim319:
need to add changes nhsn since nhsn was added in the new release

_template_python/pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+requires = ["setuptools", "setuptools-scm>=8.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "delphi_NAME"
+version = "0.1.0"
+description = "SHORT DESCRIPTION"
+readme = "README.md"
+requires-python = "== 3.8.*"
+license = { text = "MIT License" }
```
aysim319, line 11: need to change to     "License :: OSI Approved :: MIT License",

_delphi_utils_python/setup.py
aysim319, line 1: I wonder if we should keep this... I know previously we wanted to convert all the setup.py into pyproject.toml, but I rememeber the release where we converted delphi_utils setup.py into pyproject.toml has some hiccups....

aysim319: I had some issues using the pyproject.toml for nhsn and ended up just using the setup.py. since this codebase will be replaced, I think it's better to keep setup.py and have it consistent with the other indicators

PR #2020, opened 2024-08-09
Ds/makefiles

dshemetov:
### Description
The Makefiles have this pattern of activating Python environments for a single line command, which bloats the file. This directly calls the environment instead. Personal preference, but figured I'd put this up, in case others agree that this is clearer.

### Changelog
* Reduce lines in makefiles
* Ignore blame on this, since it's essentially just a formatting commit


nmdefries (approved):
This looks fine and works for me.

Issue #1268, opened 2021-09-26
Refactor claims_hosp and doctor_visits to use create_export_csv


chinandrew:
They currently use custom write_to_csv functions instead of the shared utils version. CHNG looks very similar to them and can probably be used as a reference.

dshemetov:
This may be useful for a list of diffs #1012

melange396:
reminder to update logging in `create_export_csv()` as mentioned in https://github.com/cmu-delphi/covidcast-indicators/pull/2029#discussion_r1733124034

PR #2032, opened 2024-08-26
refactor hospital admission to use delphi_utils create_export_csv

aysim319:
### Changelog
kept the existing method for testing/review purposes
- update_indicator function returns dataframe
- new function filter output to preprocess before feeding into create_export_csv
- run.py now uses create_export_csv

### Associated Issue(s) 
- Partially addresses #1268


claims_hosp/tests/test_update_indicator.py
```diff
@@ -289,3 +295,46 @@ def test_write_to_csv_wrong_results(self):
             updater.write_to_csv(res3, td.name)
 
         td.cleanup()
+
+    def test_prefilter_results(self):
```
dshemetov, line 299: praise: nice test, thanks!

question: how big is the dataset we're comparing here? do you think it's representative and gets a lot of code coverage?

suggestion: this seems like another migration test we can remove once this PR is ready for merge. 

suggestion: if we want to be especially careful, we could run this same kind of test but compare staging and prod output CSVs.

claims_hosp/delphi_claims_hosp/update_indicator.py
```diff
@@ -133,14 +133,76 @@ def geo_reindex(self, data):
...
+        )
+        output_df = pd.DataFrame()
+        if not self.parallel:
+            for geo_id, sub_data in data_frame.groupby(level=0):
+                sub_data.reset_index(inplace=True)
+                if self.weekday:
+                    sub_data = Weekday.calc_adjustment(wd_params, sub_data, ["num"], Config.DATE_COL)
+                sub_data.set_index(Config.DATE_COL, inplace=True)
+                res = ClaimsHospIndicator.fit(sub_data, self.burnindate, geo_id)
+                output_df = output_df.append(pd.DataFrame(res))
```
dshemetov, line 172: suggestion (non-blocking): the [append method is deprecated](https://pandas.pydata.org/docs/whatsnew/v1.4.0.html#deprecated-dataframe-append-and-series-append) in favor of `pd.concat(list_of_dataframes)` in pandas 2 (for various reasons, performance being one), so if we instead build a list of dataframes and do a single concat after the for-loop, we can future proof this code for our migration to pandas 2.

claims_hosp/delphi_claims_hosp/update_indicator.py
```diff
@@ -133,14 +133,76 @@ def geo_reindex(self, data):
         data_frame.fillna(0, inplace=True)
         return data_frame
 
-    def update_indicator(self, input_filepath, outpath, logger):
+    def update_indicator_to_df(self, input_filepath, logger):
```
dshemetov, line 136: question: i compared this function to update_indicator. it looks mostly the same, except for building a dataframe. any other changes?

dshemetov (commented):
thanks for doing this, made a few suggestions and a few questions

claims_hosp/delphi_claims_hosp/update_indicator.py
```diff
@@ -215,10 +277,30 @@ def update_indicator(self, input_filepath, outpath, logger):
...
+            if self.write_se:
+                assert np.all(group.val > 0) and np.all(group.se > 0), "p=0, std_err=0 invalid"
+            else:
+                group["se"] = np.NaN
+            group.drop("incl", inplace=True, axis="columns")
+            group["direction"] = np.NaN
+            output_df = output_df.append(group)
+
+        return output_df
     def write_to_csv(self, output_dict, output_path="./receiving"):
```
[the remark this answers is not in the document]
aysim319: yes :) I kept it for review/ comparing purpose

claims_hosp/delphi_claims_hosp/update_indicator.py
aysim319: nope that's the only change, since the create_export_csv takes in a dataframe, but I didn't want to remove the old way until I got the approval.

claims_hosp/tests/test_update_indicator.py
aysim319: A1: I need to double check, but I believe I got an actual file from a one-off run and should be about a gig, but I would need to double check. Do you think I should add another file that's more recent?

Response to S1: that's the idea

Response to S2: that seems like a good idea; I would need to poke around how staging is and see what happens

dshemetov: > A1: I need to double check, but I believe I got an actual file from a one-off run and should be about a gig, but I would need to double check. Do you think I should add another file that's more recent?

side-note: very important that we squash merge this PR, so the gig-sized file doesn't make it into the commit history.

> Response to S2: that seems like a good idea; I would need to poke around how staging is and see what happens

I think it would be worthwhile, so let's do that at some point. I also think that your experience with doing prod/staging comparisons will help us streamline this process in the future and make something that does branch comparisons with the press of a button.

aysim319: in staging: `text_hosptial_admission_test_export_20240903` 

and running 
```
    def test_compare_run(self):
        expected_path = "../from_staging/test_export"
        actual_path = "../receiving"
        expected_files = sorted(glob.glob(f"{expected_path}/*.csv"))
        actual_files = sorted(glob.glob(f"{actual_path}/*.csv"))
        for expected, actual in zip(expected_files, actual_files):
            with open(f"{expected_path}/{expected}", "rb") as expected_f, \
                 open(f"{actual_path}/{actual}", "rb") as actual_f:
                expected_df = pd.read_csv(expected_f)
                actual_df = pd.read_csv(actual_f)
                pd.testing.assert_frame_equal(expected_df, actual_df)
```

passed.

_delphi_utils_python/delphi_utils/export.py
```diff
@@ -113,10 +113,12 @@ def create_export_csv(
             "geo_id",
             "val",
             "se",
+            "direction",
             "sample_size",
             "missing_val",
             "missing_se",
-            "missing_sample_size"
+            "missing_sample_size",
+            "missing_direction",
```
dshemetov, line 121: question: did this break without `direction` here? I am skeptical that it is necessary. iirc, direction is a deprecated column that isn't used by any indicator.

suggestion: we never produce a `missing_direction` column anywhere, so you can remove that.

claims_hosp/tests/test_update_indicator.py
dshemetov: how many export csvs are produced by the staging run? `/common/text_hosptial_admission_test_export_20240903`?

claims_hosp/tests/test_update_indicator.py
```diff
@@ -35,6 +36,9 @@ class TestClaimsHospIndicatorUpdater:
     weekday = False
     write_se = False
     prefix = "foo"
+    start_date = "02-01-2020"
+    end_date = "06-01-2020"
+    drop_date = "2020-06-12"
```
dshemetov, line 41: To match the old code and for consistency
```suggestion
    start_date = "02-01-2020"
    end_date = "06-01-2020"
    drop_date = "06-12-2020"
```

_delphi_utils_python/delphi_utils/export.py
aysim319: It did break without directions. The csv for both hospital admissions and doctor visits uses direction

dshemetov: alright, i see, seems fine, though i'm pretty sure `missing_direction` can be removed

claims_hosp/tests/test_update_indicator.py
aysim319: 20076 or so. hospital-admission creates all geos starting from 2020-02-01 till 2024-08-31 (there's some lag)

claims_hosp/tests/test_update_indicator.py
```diff
@@ -83,31 +87,52 @@ def test_update_indicator(self):
...
+                              start_date=self.start_date,
+                              end_date=self.end_date,
+                              geo_res=geo,
+                              write_empty_days=True,
+                              sensor=Config.signal_name)
+            # output date range is half exclusive (2020-02-01 to 2020-05-31)
+            # while the export function takes the dates fully inclusive to start and end_date
             assert len(os.listdir(td.name)) == len(
-                updater.output_dates), f"failed {geo} update_indicator test"
+                updater.output_dates) + 1, f"failed {geo} update_indicator test"
```
minhkhul, line 116: question: for half exclusive, is it (2020-02-01 to 2020-05-31] or [2020-02-01 to 2020-05-31)? Export function takes dates fully [2020-02-01 to 2020-05-31], correct?

_delphi_utils_python/delphi_utils/export.py
```diff
@@ -113,10 +113,11 @@ def create_export_csv(
             "geo_id",
             "val",
             "se",
+            "direction",
```
minhkhul, line 116: question: Do you know how not having `direction` as an expected column so far hasn't been a problem? I actually don't know what `direction` describes at all, or if/how it shows up in our database. 
Additionally, how does adding `direction` as an expected column affect other indicators that don't have this column in their current implementation (nchs_mortality, etc. indicators that aren't claims_hosp or doctor_visits essentially)?

claims_hosp/tests/test_update_indicator.py
dshemetov: fwiw, i like right-inclusive and left-inclusive for (2020-02-01 to 2020-05-31] or [2020-02-01 to 2020-05-31), respectively.

_delphi_utils_python/delphi_utils/export.py
aysim319: it's because doctor-visits and hospital claims were using their own functions instead of create_export_csv. The other indicators don't use direction

dshemetov: have the same essential question about how it wasn't needed before

my understanding of `direction` is that it was meant to be a classification of which direction the indicator is trending: up, down, steady.

as for your last question, i tested the pandas filter function and it doesn't complain if you have extra columns in there, so for instance
```py
df = pd.DataFrame({"a": [0, 1], "c": [1, 2]})
df.filter(items=["a", "b"])
```
just returns the df with the "a" column back.

claims_hosp/tests/test_update_indicator.py
aysim319: Yeah me too...but since create_export_csv is fully inclusive and I'm not sure how it would affect the downstream....¯\_(ツ)_/¯

dshemetov: makes sense. in this case half exclusive means left inclusive right? since we're adding a 1 on there.

claims_hosp/delphi_claims_hosp/update_indicator.py
```diff
@@ -188,83 +182,53 @@ def update_indicator(self, input_filepath, outpath, logger):
...
+            if np.any(group.val > 90):
+                for sus_val in np.where(group.val > 90):
+                    logging.warning("value suspicious, %s: %d", geo_id, sus_val)
+            if self.write_se:
+                assert np.all(group.val > 0) and np.all(group.se > 0), "p=0, std_err=0 invalid"
+            else:
+                group["se"] = np.NaN
+            group.drop("incl", inplace=True, axis="columns")
+            group["direction"] = np.NaN
+            output_df = output_df.append(group)
```
dshemetov, line 232: again, better to use pd.concat on a list of dfs at the end

claims_hosp/delphi_claims_hosp/update_indicator.py
```diff
@@ -188,83 +182,53 @@ def update_indicator(self, input_filepath, outpath, logger):
...
+            assert not group.se.isnull().any()
+            assert np.all(group.se < 5), f"se suspicious, {geo_id}: {np.where(group.se >= 5)[0]}"
+            if np.any(group.val > 90):
+                for sus_val in np.where(group.val > 90):
+                    logging.warning("value suspicious, %s: %d", geo_id, sus_val)
+            if self.write_se:
+                assert np.all(group.val > 0) and np.all(group.se > 0), "p=0, std_err=0 invalid"
+            else:
+                group["se"] = np.NaN
+            group.drop("incl", inplace=True, axis="columns")
```
dshemetov, line 230: is this necessary here? create_export_csv will drop it anyway.

claims_hosp/delphi_claims_hosp/update_indicator.py
aysim319: good catch! ty

claims_hosp/delphi_claims_hosp/update_indicator.py
```diff
@@ -188,83 +182,53 @@ def update_indicator(self, input_filepath, outpath, logger):
...
+            group["direction"] = np.NaN
```
dshemetov, line 230: comment(non-blocking): ah the always-nan column...

claims_hosp/delphi_claims_hosp/update_indicator.py
dshemetov: could you take a look at the comment above again, i updated it

claims_hosp/delphi_claims_hosp/update_indicator.py
aysim319: yeah... I noticed that before T.T I think it was a feature that didn't make it, but it's baked into the code. I'll investigate the usage, because I think when we also grab data from the api, I'm mostly sure we drop it there too....

But that's something I don't want to get involved in this particular PR

dshemetov: i agree, keeping the indicator functioning exactly as before is reasonable. just remnants of old never-done features.

_delphi_utils_python/delphi_utils/export.py
melange396: @dshemetov 's understanding is correct, but "direction" is now very deprecated...  [It doesnt even exist in the database and is just stubbed out as null](https://github.com/cmu-delphi/delphi-epidata/blob/4e130d26c6c6fa393eae94d5cc31a29460d1897e/src/ddl/v4_schema.sql#L102) in case some user is still depending on it.  We should be removing instances of it wherever we can instead of adding new ones.

melange396 (commented):
I wish there werent so many formatting changes here, it makes it hard to see the real meat of the functionality changes.  :(

Is this all worth it?  It looks like there is a lot of stuff happening here just to shoehorn in the usage of a fairly simple ~45 line file utility method...  Does this give us an efficiency boost?  Do you have timing comparisons of runs of the old vs the new code?  Did you consider creating a smaller diff by just replacing calls to (or within) `write_to_csv()` with calls to `create_export_csv()`?

There are some logging lines that are being removed -- can we replace them with similar stuff in `create_export_csv()`?

claims_hosp/delphi_claims_hosp/update_indicator.py
aysim319: That's a good point! I actually missed sample_size if it wasn't for your comment. Hopefully this should fix the issue

aysim319:
> I wish there werent so many formatting changes here, it makes it hard to see the real meat of the functionality changes. :(
I believe good chunk of it is auto generated with darker, so that's out of my hand

dshemetov:
Honestly my bad for not documenting the darker workflow better -- make your code changes, commit them, then do `make format` in the indicator you're changing, let it make its formatting changes, and then commit those separately. I didn't think this would be necessary, since its whole premise is that it only formats the lines you touch, but it turned out to be more aggressive about resorting our imports than I thought it would be.

aysim319:
> Honestly my bad for not documenting the darker workflow better -- make your code changes, commit them, then do `make format` in the indicator you're changing, let it make its formatting changes, and then commit those separately. I didn't think this would be necessary, since its whole premise is that it only formats the lines you touch, but it turned out to be more aggressive about resorting our imports than I thought it would be.

I've been doing running the linter more frequently so some of the commit also have formatting changes, since the ci run checks for lints before the tests and I like to check that all the checks pass within my comments, but I can work on isolating the comments for functional changes and linting for future reference.

But in terms of  the file changed page the default shows all the changes that was made and unless you manually filter out the lint commits. At least that's how I understand how the file changed page works... Is there a workaround that doesn't involve manually filtering?

dshemetov:
> Is there a workaround that doesn't involve manually filtering?

Not that I know of, but often times looking through the individual commits is good enough for me.

claims_hosp/delphi_claims_hosp/update_indicator.py
dshemetov: > That's a good point! I actually missed sample_size if it wasn't for your comment. Hopefully this should fix the issue.

Glad that helped!

> Yes, we do need the incl at least until the preprocess_output that filters out with incl column being true.

I meant is it necessary to drop it in like 230, since create_export_csv will ignore it when writing the csv. But it's a minor thing, not a big deal.

claims_hosp/delphi_claims_hosp/update_indicator.py
```diff
@@ -188,83 +182,57 @@ def update_indicator(self, input_filepath, outpath, logger):
...
+            assert np.all(group.se < 5), f"se suspicious, {geo_id}: {np.where(group.se >= 5)[0]}"
+            if np.any(group.val > 90):
+                for sus_val in np.where(group.val > 90):
+                    logging.warning("value suspicious, %s: %d", geo_id, sus_val)
+            if self.write_se:
+                assert np.all(group.val > 0) and np.all(group.se > 0), "p=0, std_err=0 invalid"
+            else:
+                group["se"] = np.NaN
+            group["sample_size"] = np.NaN
+            df_list.append(group)
```
dshemetov, line 233: Rereading this bit, I'm wondering: why do we need to break apart filtered_df into groups, make a df_list, and then reappend it all together again? These are all just asserts that terminate the pipeline if any of the values is off, so we should just run the asserts, but there's no need to re-build the df. The only transformations here are `group["se"] = np.NaN` and `group["sample_size"] = np.NaN`, but those can just be handled outside, regardless of group.

claims_hosp/delphi_claims_hosp/update_indicator.py
```diff
@@ -164,22 +161,19 @@ def update_indicator(self, input_filepath, outpath, logger):
             if self.weekday
             else None
         )
-        # run fitting code (maybe in parallel)
-        rates = {}
-        std_errs = {}
-        valid_inds = {}
+        df_lst = []
+        output_df = pd.DataFrame()
```
dshemetov, line 165: Don't need to initialize this
```suggestion
```

claims_hosp/delphi_claims_hosp/update_indicator.py
dshemetov: Initially, I'd just try removing the df_list and output_df complete though and do something like this after the for loop with all the assert statements. 

```py
if not self.write_se:
    filtered_df["se"] = np.NaN
filtered_df["sample_size"] = np.NaN
filtered_df.drop(columns=["incl"], inplace = True)
assert sorted(list(filtered_df..columns)) == ["geo_id", "sample_size", "se", "timestamp", "val"]
return filtered_df
```

dshemetov: Curious (a) if that works, (b) how much that speeds things up. I'd guess that this for loop is really expensive because it runs over all counties. There may be a way to avoid the for loop using DataFrame methods, but we can get there later.

aysim319: actually the create_export_csv is what's slowing down the runs. the preprocessing takes about half the time as the previous write_to_csv. It's the delphi utils create_export_csv that's slower than the previous versions.
Still moved the nan columns outside of the loop though.

claims_hosp/delphi_claims_hosp/update_indicator.py
```diff
@@ -188,83 +181,57 @@ def update_indicator(self, input_filepath, outpath, logger):
...
+            assert not group.se.isnull().any()
+            assert np.all(group.se < 5), f"se suspicious, {geo_id}: {np.where(group.se >= 5)[0]}"
+            if np.any(group.val > 90):
+                for sus_val in np.where(group.val > 90):
+                    logging.warning("value suspicious, %s: %d", geo_id, sus_val)
+            df_list.append(group)
+            if self.write_se:
+                assert np.all(group.val > 0) and np.all(group.se > 0), "p=0, std_err=0 invalid"
+
+        output_df = pd.concat(df_list)
```
dshemetov, line 231:
```suggestion
            if self.write_se:
                assert np.all(group.val > 0) and np.all(group.se > 0), "p=0, std_err=0 invalid"

        output_df = filtered_df
```

claims_hosp/delphi_claims_hosp/update_indicator.py
dshemetov: Gotcha, I see what you mean. Really surprised create_export_csv is that much slower, that's a bit unfortunate.

aysim319:
optimziation not really panning out; closing

Issue #2027, opened 2024-08-20
Add aggregated nssp data for hhs regions


minhkhul:
**Context**

NSSP source has data at state level. We should aggregate data for hhs regions based on this state-level data. Note that nssp data is in percentage, and should be weighted accordingly at hhs regions level.

PR #2036, opened 2024-08-26
Add hhs geo level aggregation for NSSP

minhkhul:
### Description
This PR adds hhs geo level aggregation for NSSP indicator. It allows us to generate hhs data from state data. 

### Associated Issue(s) 
- Addresses #2027

### DIFF
--- a/_delphi_utils_python/delphi_utils/geomap.py
+++ b/_delphi_utils_python/delphi_utils/geomap.py
@@ -443,7 +443,7 @@ class GeoMapper:
         ---------
         data: pd.DataFrame
             The dataframe with a FIPS code column.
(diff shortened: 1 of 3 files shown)


dsweber2 (approved):
Glad to see this included quickly! This will definitely work, though it would be better if this were implemented as a method in the geomapper itself, for 2 reasons: reuse, and speed. `aggregate_by_weighted_sum` has to create the weights at every single `time_value`, which means it's definitely slower. Otoh, I doubt its enough of a speed penalty to be worth putting off having this geo level available. lgtm!

I think [this](https://github.com/cmu-delphi/delphi-epidata/pull/1535) is all we need for doc updates?

PR #2042, opened 2024-08-29
add testing environment indicator to the logger

dsweber2:
### Description
Following a recommendation from the [structlog docs on testing](https://www.structlog.org/en/stable/testing.html). Needed so that I can actually test the log output from various warnings and errors in developing the Wastewater endpoint https://github.com/cmu-delphi/delphi-epidata/pull/1513. Marked as a draft because I am still testing that this is all that's needed to be able to test the logger.
### Changelog
- Adds another option to `get_structured_logger(name, filename, log_exceptions, *is_real=True*)` indicating whether to log as if in a testing environment
- A bunch of formatting things that `ruff` decided were Correct. Can remove if they're a problem.

### DIFF
--- a/_delphi_utils_python/delphi_utils/logger.py
+++ b/_delphi_utils_python/delphi_utils/logger.py
@@ -23,8 +23,7 @@ def handle_exceptions(logger):
     """Handle exceptions using the provided logger."""
 
     def exception_handler(scope, etype, value, traceback):
-        logger.exception("Top-level exception occurred",
-                         scope=scope, exc_info=(etype, value, traceback))
+        logger.exception("Top-level exception occurred", scope=scope, exc_info=(etype, value, traceback))
 
     def sys_exception_handler(etype, value, traceback):
         exception_handler("sys", etype, value, traceback)
@@ -33,21 +32,19 @@ def handle_exceptions(logger):
         if args.exc_type == SystemExit and args.exc_value.code == 0:
             # `sys.exit(0)` is considered "successful termination":
             #   https://docs.python.org/3/library/sys.html#sys.exit
-            logger.debug("normal thread exit", thread=args.thread,
-                        stack="".join(
-                            format_exception(
-                                args.exc_type, args.exc_value, args.exc_traceback)))
+            logger.debug(
+                "normal thread exit",
+                thread=args.thread,
+                stack="".join(format_exception(args.exc_type, args.exc_value, args.exc_traceback)),
+            )
         else:
-            exception_handler(f"thread: {args.thread}",
-                              args.exc_type, args.exc_value, args.exc_traceback)
+            exception_handler(f"thread: {args.thread}", args.exc_type, args.exc_value, args.exc_traceback)
 
     sys.excepthook = sys_exception_handler
     threading.excepthook = threading_exception_handler
 
 
-def get_structured_logger(name=__name__,
-                          filename=None,
-                          log_exceptions=True):
+def get_structured_logger(name=__name__, filename=None, log_exceptions=True):
     """Create a new structlog logger.
 
     Use the logger returned from this in indicator code using the standard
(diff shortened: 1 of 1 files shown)


_delphi_utils_python/delphi_utils/logger.py
```diff
@@ -115,7 +112,7 @@ def add_pid(_logger, _method_name, event_dict):
         # Use a standard wrapper class for utilities like log.warning()
         wrapper_class=structlog.stdlib.BoundLogger,
         # Cache the logger
-        cache_logger_on_first_use=True,
+        cache_logger_on_first_use=is_real,
```
dshemetov, line 115: Is this the main change and everything else is formatting? Any chance you could separate the formatting into a separate commit ahead of your other changes?

aysim319:
@dsweber2 I would suggest adding @minhkhul as a reviewer, I know that she's working with some elastic and logging stuff I feel like she would be able to give some insight on that part

_delphi_utils_python/delphi_utils/logger.py
dsweber2: other than adding `is_real` and some docs yeah

dsweber2:
Mostly unrelated, but I would definitely prefer if we dealt with formatting stuff all in one PR.

minhkhul:
If in your case, test runs are 1. also run on prod, and 2. the logged info must be output to the same log file as real runs, then the `is_real` addition makes sense.
But elastic allows us to distinguish different kinds of run using server names and/or the dir where log are output to. So far, that's been enough to help distinguish things. I think it's a good idea to separate where we put real log files and test log files anyway. 
So instead of adding a flag, I suggest doing those 2 things first if possible (run test on staging and/or output log of test to a separate location from real logging). 

If you want to, you can set up an elastic ingest pipeline to take advantage of elastic's log parsing by doing these:
1. Put your test log file into something like `/var/log/test-wastewater/`
2. Set up an elastic ingest pipeline like [this](https://logs-prod-01.delphi.cmu.edu/app/management/ingest/ingest_pipelines?queryText=filebeat-epidata-pipeline-staging&pipeline=filebeat-epidata-pipeline-staging) 
3. Add pickup configuration like [this](https://github.com/cmu-delphi/delphi-admin/blob/ef89898b548b36f1abce1773e25e919384443d10/ansible/templates/delphi-stage/delphi-stage.filebeat.yml.j2#L35-L39) so your test log are automatically picked up and put into the ingestion pipeline you created in step 3.
4. Then you can make dashboard like [this](https://logs-prod-01.delphi.cmu.edu/app/dashboards#/view/5f6f0490-7dff-42da-bd68-4c1f32ddde11?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-2w,to:now)))

melange396:
@minhkhul the point of this is to make it easier to evaluate the application's logging functionality in a sandboxed test environment; Elastic isnt part of the scope.

I would lean towards removing the formatting changes from this PR, they distract from the core diffferences being applied and the rationale.

_delphi_utils_python/delphi_utils/logger.py
```diff
@@ -33,21 +32,19 @@ def threading_exception_handler(args):
...
+            exception_handler(f"thread: {args.thread}", args.exc_type, args.exc_value, args.exc_traceback)
 
     sys.excepthook = sys_exception_handler
     threading.excepthook = threading_exception_handler
 
 
-def get_structured_logger(name=__name__,
-                          filename=None,
-                          log_exceptions=True):
+def get_structured_logger(name=__name__, filename=None, log_exceptions=True, is_real=True):
```
melange396, line 47:
```suggestion
def get_structured_logger(name=__name__, filename=None, log_exceptions=True, *, test_mode=False):
```
and then below:
```py
cache_logger_on_first_use=not(test_mode),
```

This name (and inverted truth value) makes the purpose of the argument clearer, and the `*` in the arguments list ensures its only ever triggered explicitly.

dsweber2:
Thanks all for the comments, I eventually figured out a way around it without having to modify the functionality of `get_structured_logger` while using `pytests`.

PR #2057, opened 2024-09-16
2034 add retry loop to Google symptoms data pull

### LINKED ISSUE #2034: Add a retry loop for google symptoms
When bigquery api call fails with a 500 error code*; we fail without retrying. We should add a retry loop when error code is 500 or the error message is backendError

There's a known issue with python request that 500 code gets converted into 400 (https://stackoverflow.com/questions/41129805/why-is-google-bigquery-api-returning-a-400-instead-of-a-500-status-code)
 
 and in the case that it returns a 400 error code we also need to check for the error message

**Context**

https://delphi-org.slack.com/archives/C01LZ3A2UMU/p1724504475248319

https://cloud.google.com/bigquery/docs/error-messages

aysim319:
### Description
Added retry loop to handle google api request more gracefully.
Also handles a case where a server error gets miscategorized as bad request error

### Associated Issue(s) 
- Closes #2034

### DIFF
--- a/google_symptoms/delphi_google_symptoms/constants.py
+++ b/google_symptoms/delphi_google_symptoms/constants.py
@@ -113,3 +113,5 @@ DC_FIPS = "11001"
 FULL_BKFILL_START_DATE = datetime.strptime("2020-02-20", "%Y-%m-%d")
 
 PAD_DAYS = 7
+
+NUM_RETRIES = 3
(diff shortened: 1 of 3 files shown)


minhkhul (approved):
To sum up, instead or raising an error immediately like current, this PR tries one more pull if encounter backend error or internal server error. If the second pull doesn't work, error out as usual.

Issue #2070, opened 2024-10-28
Add backup of source data to nchs and nssp


minhkhul:
Context: https://delphi-org.slack.com/archives/C013AH5N01E/p1728499163421519

We got this PR https://github.com/cmu-delphi/covidcast-indicators/pull/2065 for nchs
Just need to make nssp use the same mechanism and remove the current clunky mechanism on bigchunk-dev-02.

- [x] check nchs run on staging + set up staging stash while waiting for release
- [ ] check nchs run on prod
- [ ] move local + staging nchs stash to prod

- [ ] PR for nssp move
- [ ] check nssp on staging
- [ ] check nssp on prod
- [ ] move old stash on dev-2 to prod
- [ ] remove old mechanism from bigchunk-dev-02

PR #2072, opened 2024-10-29
2070 add backup of source data to nssp

minhkhul:
### Description
Context at #2070 

### Changelog
- Make nssp pull use delphi_utils.export.create_backup_csv() to save backup csv whenever it pulls new data in prod run.
- Adjust run, prod params.json and params.json.template accordingly
- Add test case accordingly


ansible/templates/nssp-params-prod.json.j2
```diff
@@ -1,6 +1,7 @@
 {
   "common": {
     "export_dir": "/common/covidcast/receiving/nssp",
+    "backup_dir": "./raw_data_backups",
```
aysim319, line 4: where would this location be? I feel like it would be better to be somewhere in /common/

minhkhul: This location would be `/home/indicators/runtime/nssp//raw_data_backups`. The location is because of precendents in [chng](https://github.com/cmu-delphi/covidcast-indicators/blob/3450dfc09ff8c4bbc9d89c28ddc78b81f52e2ee6/ansible/templates/changehc-params-prod.json.j2#L7), [hospital admissions](https://github.com/cmu-delphi/covidcast-indicators/blob/3450dfc09ff8c4bbc9d89c28ddc78b81f52e2ee6/ansible/templates/claims_hosp-params-prod.json.j2#L7C1-L7C37), [nchs](https://github.com/cmu-delphi/covidcast-indicators/blob/3450dfc09ff8c4bbc9d89c28ddc78b81f52e2ee6/ansible/templates/nchs_mortality-params-prod.json.j2#L3-L4), [quidel](https://github.com/cmu-delphi/covidcast-indicators/blob/3450dfc09ff8c4bbc9d89c28ddc78b81f52e2ee6/ansible/templates/quidel_covidtest-params-prod.json.j2#L8C3-L8C34). 
Basically, cache files of all sorts are historically stored within the respective dir where the code lives on prod, I assume because we don't want these files modified/disturbed. Meanwhile, common is for stuff that needs interaction with other processes, which is why the normal daily runs all output to common. It's so acquisition can pick them up. The permissions are set up accordingly as well. 
Considering we probably don't want these files modified and just stashed away, I think keeping them where they are is okie.

nssp/tests/test_pull.py
```diff
@@ -33,11 +33,27 @@ def test_pull_nssp_data(self, mock_socrata):
...
+        mock_logger = MagicMock()
+
         # Call function with test token
         test_token = "test_token"
-        result = pull_nssp_data(test_token)
+        result = pull_nssp_data(test_token, backup_dir, custom_run, mock_logger)
         print(result)
 
+        # Check logger used:
+        mock_logger.info.assert_called()
```
aysim319, line 51: you can also assert on the actual log message by using caplog: (https://pytest-with-eric.com/fixtures/built-in/pytest-caplog/)
```
 def test_merge_backfill_file(self, caplog, monkeypatch):
        caplog.set_level(logging.INFO)
        logger = get_structured_logger()

        fn = "quidel_covidtest_202008.parquet"
        assert fn not in os.listdir(backfill_dir)
        
        # Check when no daily file stored
        today = datetime(2020, 8, 20)
        merge_backfill_file(backfill_dir, today, logger, test_mode=True)
        assert fn not in os.listdir(backfill_dir)
        assert "No new files to merge; skipping merging" in caplog.text
       
```

nssp/delphi_nssp/pull.py
```diff
@@ -59,6 +65,7 @@ def pull_nssp_data(socrata_token: str):
         results.extend(page)
         offset += limit
     df_ervisits = pd.DataFrame.from_records(results)
+    create_backup_csv(df_ervisits, backup_dir, custom_run, logger=logger)
```
aysim319, line 68: I saw that there's no test for the actual create_backup_csv function for the last PR. can we add test for that in this PR?

nmdefries (approved):
looks good

_delphi_utils_python/tests/test_export.py
```diff
@@ -386,3 +387,18 @@ def test_export_sort(self, tmp_path):
...
+        today = datetime.strftime(datetime.today(), "%Y%m%d")
+        dtypes = self.DF.dtypes.to_dict()
+        del dtypes["timestamp"]
+        geo_res = "county"
+        metric = "test"
+        sensor = "deaths"
+        create_backup_csv(df=self.DF, backup_dir=tmp_path, custom_run=False, issue=None, geo_res=geo_res, metric=metric, sensor=sensor, logger=logger)
+        assert "Backup file created" in caplog.text
+
+        actual = pd.read_csv(join(tmp_path, f"{today}_{geo_res}_{metric}_{sensor}.csv.gz"), dtype=dtypes, parse_dates=["timestamp"])
```
aysim319, line 403: @nmdefries this gets me bit worried... when I was writing this, when we write to csv, the datatypes of the columns are not preserved for csv. Is there a particular reason we want to use csv.gz compared to parquet/other data format?

nmdefries: Basically, status quo. We use CSV for all of our other data storage (meaning we don't preserve data types anywhere else). Since these files are intended to be backups/not in regular use, having them mesh well with other systems doesn't matter so much. I don't have a strong preference. I'm worried, though, that using a different less-standard format will cause more engineering effort here. One of our goals for this was quick rollout.

The pro of using a data-type preserving format is that it lets us read these files in and continue the pipeline as if we were pulling from the source and doing our normal processing.

@minhkhul thoughts? Maybe y'all could briefly talk to Adam about this.

aysim319:
daily run size result:
7MB for nssp (5MB for csv 2MB for parquet) 600KB for nchs (400KB for csv 200KB for parquet)

_delphi_utils_python/tests/test_export.py
aysim319: talked with adam and decided on storing both csv and parquet for now; shouldn't be a problem for a while as both nssp and nchs is small

minhkhul:
@aysim319 🙏

aysim319:
Thanks for the double check :) @minhkhul

Issue #2073, opened 2024-11-06
Add signals from NSSP secondary source


minhkhul:
**Data details**
 
Basically we're adding 4 signals from [this](https://data.cdc.gov/Public-Health-Surveillance/2023-Respiratory-Virus-Response-NSSP-Emergency-Dep/7mra-9cq9/data_preview) secondary nssp source dataset. They report essentially the same data as the [main dataset](https://data.cdc.gov/Public-Health-Surveillance/NSSP-Emergency-Department-Visit-Trajectories-by-St/rdmq-nq56/data_preview), but only available at state level and above. However, it appears this secondary source updates quicker than the main source recently. 

**Additional context**
Context: https://delphi-org.slack.com/archives/C0130CSQRN3/p1730476295781409

TODO:
- [x] Confirm names with Roni
- [x] Test runs on staging
- [ ] Release
- [x] Update public facing docs
- [ ] Talk to Carlyn about signal sheets/app.

minhkhul:
seg faults on staging when testing. Seems deeper problem with python versioning.

minhkhul:
Things are up and running on staging again. Removed create_backup_csv due to unrelated issue. Then ran the indicator with secondary signals code + acquisition to staging db for testing. Result looks reasonable:
Indicator run on staging: https://cronicle-prod-01.delphi.cmu.edu/#JobDetails?id=jm3t9mtqzvq
Acquisition run on staging: https://cronicle-prod-01.delphi.cmu.edu/#JobDetails?id=jm3t9qfpfvr
Some querying on staging db:
```
mysql> SELECT `signal`, issue, geo_type, COUNT(*) FROM epimetric_full_v WHERE source 
= 'nssp' AND `signal` LIKE '%2023rvr' GROUP BY `signal`, issue, geo_type;
+---------------------------------+--------+----------+----------+
| signal                          | issue  | geo_type | COUNT(*) |
+---------------------------------+--------+----------+----------+
| pct_ed_visits_combined_2023rvr  | 202447 | state    |     5124 |
| pct_ed_visits_combined_2023rvr  | 202447 | hhs      |     1070 |
| pct_ed_visits_combined_2023rvr  | 202447 | nation   |      107 |
| pct_ed_visits_covid_2023rvr     | 202447 | state    |     5124 |
| pct_ed_visits_covid_2023rvr     | 202447 | hhs      |     1070 |
[...]

PR #2074, opened 2024-11-06
Add NSSP secondary source

minhkhul:
### Description
Add 4 signals: `pct_ed_visits_covid_2023RVR`, `pct_ed_visits_influenza_2023RVR`, `pct_ed_visits_rsv_2023RVR`, `pct_ed_visits_combined_2023RVR` to nssp.

These secondary signals are vailable at geo levels `state`, `hhs` and `nation`.

Adding 4 signals from [this](https://data.cdc.gov/Public-Health-Surveillance/2023-Respiratory-Virus-Response-NSSP-Emergency-Dep/7mra-9cq9/data_preview) secondary nssp source dataset. They report essentially the same data as the [main dataset](https://data.cdc.gov/Public-Health-Surveillance/NSSP-Emergency-Department-Visit-Trajectories-by-St/rdmq-nq56/data_preview), but only available at state level and above. However, it appears this secondary source updates quicker than the main source recently. 

More info at [DETAILS.md](https://github.com/cmu-delphi/covidcast-indicators/blob/f3cb5090106c80423c8900cbb4672e26004dfa5a/nssp/DETAILS.md)

Context: https://delphi-org.slack.com/archives/C0130CSQRN3/p1730476295781409

Things to think about:
- [x] need signal names approval from Roni
- [x] source stash? (different source -> different stash?)

### Associated Issue(s) 
- Addresses #2073

### DIFF
--- a/nssp/DETAILS.md
+++ b/nssp/DETAILS.md
@@ -5,12 +5,22 @@ We import the NSSP Emergency Department Visit data, including percentage and smo
 There are 2 sources we grab data from for nssp:
 - Primary source: https://data.cdc.gov/Public-Health-Surveillance/NSSP-Emergency-Department-Visit-Trajectories-by-St/rdmq-nq56/data_preview
 - Secondary source: https://data.cdc.gov/Public-Health-Surveillance/2023-Respiratory-Virus-Response-NSSP-Emergency-Dep/7mra-9cq9/data_preview
-There are 8 signals output from the primary source and 4 output from secondary. Secondary source data is only available from state-level geos and above, though secondary data might be updated more often.
+There are 8 signals output from the primary source and 4 output from secondary. There are no smoothed signals from secondary source.
+
+Note that the data produced from secondary source are mostly the same as their primary source equivalent, with past analysis shows around 95% of datapoints having less than 0.1 value difference and the other 5% having a 0.1 to 1.2 value difference. 
 
 ## Geographical Levels
-* `state`: reported using two-letter postal code
-* `county`: reported using fips code
-* `national`: just `us` for now
+Primary source:
+* `state`: reported from source using two-letter postal code
+* `county`: reported from source using fips code
+* `national`: just `us` for now, reported from source
+* `hhs`, `hrr`, `msa`: not reported from source, so we computed them from county-level data using a weighted mean. Each county is assigned a weight equal to its population in the last census (2020).
+
+Secondary source:
+* `state`: reported from source
+* `hhs`: reported from source
+* `national`: reported from source
+
 ## Metrics
 *  `percent_visits_covid`, `percent_visits_rsv`, `percent_visits_influenza`: percentage of emergency department patient visits for specified pathogen.
 *  `percent_visits_combined`: sum of the three percentages of visits for flu, rsv and covid.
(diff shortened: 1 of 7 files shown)


aysim319 (changes requested):

nssp/delphi_nssp/run.py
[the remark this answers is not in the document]
minhkhul: yup the weekly stuff is a bug. I'll fix that in a moment. 
The rounded shortness is from source. Definitely nees a note in the issue + readme changes for that.

nssp/tests/test_pull.py
```diff
@@ -55,6 +61,34 @@ def test_pull_nssp_data(self, mock_socrata):
...
+        # Call function with test token
+        test_token = "test_token"
+        result = secondary_pull_nssp_data(test_token)
+        # print(result)
+
+        # Check that Socrata client was initialized with correct arguments
+        mock_socrata.assert_called_once_with("data.cdc.gov", test_token)
+
+        # Check that get method was called with correct arguments
+        mock_client.get.assert_any_call("7mra-9cq9", limit=50000, offset=0)
```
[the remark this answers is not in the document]
minhkhul: The schema of the 2 sources are actually quite different, hence the discrepancy in the columns, and also why I didn't use the same test data file. But I like what you said about making tests for the 2 sources to get data from the same time frame, so Imma do that.

nssp/delphi_nssp/pull.py
```diff
@@ -79,3 +96,50 @@ def pull_nssp_data(socrata_token: str, backup_dir: str, custom_run: bool, logger
...
+        "Region ", ""
+    )
+    df_ervisits.loc[hhs_region_mask, "geo_type"] = "hhs"
+
+    df_ervisits["signal"] = df_ervisits["signal"].map(SECONDARY_SIGNALS_MAP)
+
+    df_ervisits = df_ervisits[SECONDARY_KEEP_COLS]
+
+    try:
+        df_ervisits = df_ervisits.astype(SECONDARY_TYPE_DICT)
```
nolangormley, line 141: [nitpick] You can specify column names and data types on load with Pandas. That might be better than the renaming and casting here.

nssp/delphi_nssp/run.py
```diff
@@ -139,5 +141,51 @@ def run_module(params):
             if len(dates) > 0:
                 run_stats.append((max(dates), len(dates)))
 
+    logger.info("Generating secondary signals")
+    secondary_df_pull = secondary_pull_nssp_data(socrata_token)
+    for signal in SECONDARY_SIGNALS:
+        secondary_df_pull_signal = secondary_df_pull[secondary_df_pull["signal"] == signal]
+        if secondary_df_pull_signal.empty:
+            continue
```
nolangormley, line 149: Think it would be worth it to log signals that are empty during the runs, or would that get too noisy?

nolangormley (commented):
Some small nitpicks, but overall looks good!

nssp/DETAILS.md
```diff
@@ -2,6 +2,11 @@
 
 We import the NSSP Emergency Department Visit data, including percentage and smoothed percentage of ER visits attributable to a given pathogen, from the CDC website. The data is provided at the county level, state level and national level; we do a population-weighted mean to aggregate from county data up to the HRR and MSA levels.
 
+There are 2 sources we grab data from for nssp:
```
aysim319, line 5: Mention that there are some difference that's significant; I actually got 4% of the data having more that .1 difference and I would also mention this to logan/daniel/etc as an fyi

aysim319 (approved):
Add the comment in the details that there are some difference; in the sample run there were 4% that was more than .1 difference and it should be good.

Issue #2089, opened 2024-12-19
Move source backup content to common


minhkhul:
**Context**

I think for next release maybe also change where the raw files are saved; maybe /common/indicator/<data_source>?

_Originally posted by @aysim319 in https://github.com/cmu-delphi/covidcast-indicators/pull/2000#discussion_r1888869042_

The current backup dir setup is placing sour backup files into a location super inconvenient for patching, so let's move it.

**Todo**

- [x] make dir on prod in common using automation + give appropriate permissions
- [x] PR to change to ansible template params.json + adjust local template accordingly
- [ ] BEFORE RELEASE: move old backup files from current place to new dir setup.

PR #2090, opened 2024-12-19
move source backup dir to common

minhkhul:
### Changelog
Change backup dir to `/common/covidcast/source_backup/` in prod

### Associated Issue(s) 
#2089


aysim319 (approved):

aysim319:
I noitced there was another directory /common/indicators/ do you know what that's about? but seeing that the batch_issue_upload is also under /common/covidcast/ it makes more sense.
we also probably have to move the existing backups and put it there

Issue #2085, opened 2024-12-10
Add proportions for NHSN (changed to just total reporting hospitals)


aysim319:
Due to time constraint the main metric that would be available once nhsn is deployed is raw counts. However, due to the reporting mandate not being consistent; the number of reporting hospital should be considered. 

This was mentioned in previous iterations, but at the time was not a factor as normalizing the data would result in noise and therefore would not add consider number of reporting hospitals.

At the final stages of implementing the indicator, it was decided that the proprotions should be added, but in a future release in order prevent more delays in initial release of nhsn

aysim319:
potential method was dicussed
> We could report [totalnumberreportinghospital]/max_per_geo([totalnumberreportinghospital]) (fraction of hospitals currently reporting compared to max ever seen in a location) as the signal and [totalnumberreportinghospital] as the sample size

but led to the conclusion of just introducing the total reporting hospital 
> One concern about using the concept of "max ever seen in a location" is that this value can change over time.  If you meant the max over the specific period when reporting wasn't mandatory (May 1 through Oct 31), then it feels like an ad-hoc, one-off kind of thing -- what do you do if there is another period of non-mandatory?  (and also, even during mandatory periods, the number of reporting hospitals can change, and should therefore be reported.
I think it would be cleaner to just report the total number of reporting hospitals (not only during the non-mandatory period), and let users derive reporting proportions on their own if they would like.

PR #2102, opened 2025-01-17
add proportions for nhsn

aysim319:
### Description
between April 30 2024 to October 31 2024, reporting was on a voluntary basis and therefore expect a dip during that time period. In order to take account of the fluctuating data, the new signals takes proportions on number of reporting hospitals into account

### Associated Issue(s) 
- Addresses #2085


aysim319:
need more exploratory work (haven't decided on how to caluculate the proportions

Issue #2091, opened 2024-12-24
add retry for 50x error for scorata api and also throw error when that happens


aysim319:
Saturday run for nhsn failed with 503 error; potentially missing updates that would've happened between the time. The error only showed up in logs, but marked as run succeeded in cronicle

PR #2111, opened 2025-01-28
2085 add proportions nhsn

aysim319:
### Description
address #2085 

### Changelog
added new function that checks the metadata for last update
-> the function also checks for 503 error
-> added signals for total reporting hospitals

### Associated Issue(s) 
- Addresses #2091


nhsn/delphi_nhsn/constants.py
```diff
@@ -6,29 +6,48 @@
...
+TOTAL_ADMISSION_COVID_COL = "totalconfc19newadm"
+TOTAL_ADMISSION_FLU_COL = "totalconfflunewadm"
+NUM_HOSP_REPORTING_COVID_COL = "totalconfc19newadmhosprep"
+NUM_HOSP_REPORTING_FLU_COL = "totalconfflunewadmhosprep"
+
+# signal name
+TOTAL_ADMISSION_COVID = "confirmed_admissions_covid_ew"
+TOTAL_ADMISSION_FLU = "confirmed_admissions_flu_ew"
+NUM_HOSP_REPORTING_COVID = "num_reporting_hospital_covid_ew"
+NUM_HOSP_REPORTING_FLU = "num_reporting_hospital_flu_ew"
```
nmdefries, line 18: praise: nice refactor!

nhsn/delphi_nhsn/pull.py
```diff
@@ -11,20 +15,45 @@
...
+    except HTTPError as err:
+        if err.code == 503:
+            time.sleep(2 + random.randint(0, 1000) / 1000.0)
+            response = client.get_metadata(dataset_id)
+        else:
+            raise err
+
+    updated_timestamp = datetime.utcfromtimestamp(int(response["rowsUpdatedAt"]))
+    now = datetime.utcnow()
+    recently_updated = (now - updated_timestamp) < timedelta(days=1)
```
nmdefries, line 31: issue: I think this "recently-updated" logic is sufficient but not robust. For example, if we fail to pull data for multiple days, the next day we run we would not pull data we had never seen before if it was not posted in the last day.

The more robust solution would be to save last pull's `updated_timestamp` to a local file. We would then load that and compare `updated_timestamp` to that -- if exactly equal, skip update; if unequal, pull data.

nhsn/delphi_nhsn/pull.py
```diff
@@ -11,20 +15,45 @@
...
-    limit = 50000  # maximum limit allowed by SODA 2.0
-    while True:
-        page = client.get(dataset_id, limit=limit, offset=offset)
-        if not page:
-            break  # exit the loop if no more results
-        results.extend(page)
-        offset += limit
-
-    df = pd.DataFrame.from_records(results)
+    recently_updated = check_last_updated(client, "ua7e-t2fy", logger)
```
nmdefries, line 43: suggestion:

```suggestion
    recently_updated = check_last_updated(client, dataset_id, logger)
```

nmdefries (commented):
Are you planning on adding the new RSV signals in this PR or in a separate one?

aysim319:
> Are you planning on adding the new RSV signals in this PR or in a separate one?

I was originally planning for a seperate one more so because when I tried to add just the new columns and locally ran the tests and ran across issues and looks like it might be more involved

nhsn/delphi_nhsn/pull.py
```diff
@@ -1,30 +1,87 @@
...
-        offset += limit
-
-    df = pd.DataFrame.from_records(results)
+    recently_updated = check_last_updated(client, dataset_id, logger)
+    df = pd.DataFrame()
+    if recently_updated:
+        results = []
+        offset = 0
+        limit = 50000  # maximum limit allowed by SODA 2.0
+        while True:
```
nolangormley, line 77: maybe something like this to get rid of the `while True`

```
page = client.get(dataset_id, limit=limit, offset=offset)
while page:
    results.extend(page)
    offset += limit

    page = client.get(dataset_id, limit=limit, offset=offset)
```

aysim319: yeah they're there from the copy pasta from nssp 😅

nhsn/delphi_nhsn/pull.py
```diff
@@ -1,30 +1,85 @@
...
+        logger.info(f"{prelim_prefix}NHSN data was recently updated; Pulling data", updated_timestamp=updated_timestamp)
+    elif not recently_updated_api:
+        logger.info(
+            f"{prelim_prefix}NHSN data is missing issue; Pulling data",
+            updated_timestamp=updated_timestamp,
+            issue=latest_issue_date,
+        )
+    else:
+        logger.info(f"{prelim_prefix}NHSN data is stale; Skipping", updated_timestamp=updated_timestamp)
+    return recently_updated_source or not recently_updated_api
```
aysim319, line 65: @nolangormley does the logic make sense here? 
i'm going by nat's comment:

> issue: I think this "recently-updated" logic is sufficient but not robust. For example, if we fail to pull data for multiple days, the next day we run we would not pull data we had never seen before if it was not posted in the last day.

I tried to resolve nat's concern, but i don't know if i'm just irrationally suspicous, but something doesn't seem right.

nhsn/delphi_nhsn/pull.py
nmdefries: Yeah, checking the API could make sense, too. The one thing I'd caution is timezones -- your previous approach explicitly used UTC on both "old" and "now" timestamps, but I don't know what the API uses.

Second, the API only has dates, not times. Would that ever cause problems? E.g. we want to check for updates multiple times a day.

aysim319: > Yeah, checking the API could make sense, too. The one thing I'd caution is timezones -- your previous approach explicitly used UTC on both "old" and "now" timestamps, but I don't know what the API uses
since the data and the dates are just date and not datetime, I didn't take timezones into account....hmm i also don't know for sure which timezone, i believe it's EST, but have to double check

> the API only has dates, not times. Would that ever cause problems? E.g. we want to check for updates multiple times a day.
since this is data that *generally* updates weekly, I was planning on just running once a day, so I thought timezone wouldn't be as much of an issue

nhsn/delphi_nhsn/pull.py
aysim319: @nmdefries hmm I'm also wondering if it's really necessary, should I get rid of it and run daily? if storage isn't that much of a concern, it is just better to have daily backups.....

nmdefries: hmmmm fair point

nmdefries: Don't we process issue dates for weekly data sources that would cause collisions?

nhsn/delphi_nhsn/pull.py
nmdefries: Okay, given these complications, I'm thinking reading/writing to a file is easier. We wouldn't need to keep a complete list of all update dates ever, just the single most recent date. So the file wouldn't keep getting bigger and bigger, we could just read a single line.

nhsn/delphi_nhsn/pull.py
aysim319: let me double check, but the behavior of the acquisition is to upsert

nmdefries: Oh, that would be convenient.

nhsn/delphi_nhsn/pull.py
aysim319: oh great I just went over with Nolan just now about this concern and already changed it to create backups daily, but still do simple check to see recently updated to actually continue processing and create the csv files

nhsn/delphi_nhsn/pull.py
```diff
@@ -11,20 +15,57 @@
...
+    try:
         page = client.get(dataset_id, limit=limit, offset=offset)
-        if not page:
-            break  # exit the loop if no more results
+    except HTTPError as err:
+        if err.code == 503:
+            time.sleep(2 + random.randint(0, 1000) / 1000.0)
+            page = client.get(dataset_id, limit=limit, offset=offset)
+        else:
+            raise err
```
nolangormley, line 57: this should probably log as well

nolangormley (changes requested):
Some small questions

nhsn/delphi_nhsn/pull.py
```diff
@@ -11,20 +15,57 @@
...
+        prelim_prefix = "Preliminary " if dataset_id == PRELIM_DATASET_ID else ""
+        if recently_updated_source:
+            logger.info(
+                f"{prelim_prefix}NHSN data was recently updated; Pulling data", updated_timestamp=updated_timestamp
+            )
+        else:
+            logger.info(f"{prelim_prefix}NHSN data is stale; Skipping", updated_timestamp=updated_timestamp)
+    # pylint: disable=W0703
+    except Exception as e:
+        logger.info("error while processing socrata metadata; continuing processing", error=str(e))
```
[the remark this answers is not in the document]
aysim319: Yeah because it's just checking if it's stale or not, if we already have the data I don't want this to be the reason that the indicator stops. This was a check to limit duplicating data

nhsn/delphi_nhsn/pull.py
aysim319: Ah yup, I missed that;

nhsn/delphi_nhsn/pull.py
```diff
@@ -89,25 +130,33 @@ def pull_nhsn_data(
...
+                logger.info("column not available in data", col_name=col_name)
+                keep_columns.remove(signal)
 
         df = df[keep_columns]
         df["geo_id"] = df["geo_id"].str.lower()
         df.loc[df["geo_id"] == "usa", "geo_id"] = "us"
-        df = df.astype(TYPE_DICT)
+        try:
+            df = df.astype(TYPE_DICT)
+        except KeyError:
```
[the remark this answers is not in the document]
aysim319: It shouldn't and probably wont when I need to patch older data (that was the whole reason I needed to merge this before RSV) 

I need to remove that line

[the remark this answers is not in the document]
aysim319: The idea was that some of the older data didn't have newer signals (rsv, reporting hospitals) in the source back up it would log, but resume the patching process.

Previously I tried modify TYPE_DICT in a patch run, but being mutable caused some issue. So this was the next solution...is it a good one ehhh....I log in the run.py that the signal is unavailable, and I thought I shouldn't log basically the same message twice

nolangormley (approved):
LGTM!

nhsn/delphi_nhsn/pull.py
```diff
@@ -11,20 +16,73 @@
...
     limit = 50000  # maximum limit allowed by SODA 2.0
-    while True:
+    # retry logic for 500 error
+    try:
         page = client.get(dataset_id, limit=limit, offset=offset)
-        if not page:
-            break  # exit the loop if no more results
+    except HTTPError as err:
+        if err.code == 503:
+            time.sleep(2 + random.randint(0, 1000) / 1000.0)
```
nmdefries, line 70: discussion: if you wanted to wait multiple times with increasingly long timeouts, you could do [something like this](https://github.com/cmu-delphi/covidcast/blob/c89e4d295550ba1540d64d2cc991badf63ad04e5/R-packages/evalcast/R/get_target_response.R#L38-L69) (in R). The `while` loop goes over the total # of attempts, exiting early if it gets a successful result.

Not expecting you to add this in and not sure it matters here, just bringing it up for future applications. An [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) (successive doublings) is the standard way to increase wait times.

nhsn/delphi_nhsn/pull.py
```diff
@@ -89,25 +147,33 @@ def pull_nhsn_data(
...
     # Pull data from Socrata API
     df = (
-        pull_data(socrata_token, dataset_id=MAIN_DATASET_ID)
+        pull_data(socrata_token, MAIN_DATASET_ID, backup_dir, logger)
         if not custom_run
         else pull_data_from_file(backup_dir, issue_date, logger, prelim_flag=False)
     )
 
-    keep_columns = list(TYPE_DICT.keys())
+    recently_updated = True if custom_run else check_last_updated(socrata_token, MAIN_DATASET_ID, logger)
```
nmdefries, line 155: suggestion: if we put this before the `pull_data` logic, we could avoid fetching from the source API in most cases (since this is or will be running every day but only updates once a week).

nhsn/delphi_nhsn/pull.py
```diff
@@ -144,24 +210,31 @@ def pull_preliminary_nhsn_data(
     pd.DataFrame
         Dataframe as described above.
     """
+    # Pull data from Socrata API
```
nmdefries, line 213: issue: `pull_preliminary_nhsn_data` and `pull_nhsn_data` are really similar. I think it will become a maintenance issue to keep both. We should probably keep these two functions as wrappers of a shared fn that takes a `is_prelim` flag (or similar).

Diff of the two fns:

Screen Shot 2025-02-11 at 15 20 55

nmdefries (changes requested):
A couple cleanup requests -- deduping `pull_nhsn...` and `pull_preliminary...` is the main one.

nhsn/delphi_nhsn/pull.py
aysim319: I only did one time for exponential back offs, I think I did the first iteration like google-symptoms, moslty cuz I wasn't sure if I can/should do more iterations.

nhsn/delphi_nhsn/pull.py
aysim319: that was the original thought, but previously you brought up if there's multiple failures in a row and the solution was to squirrel away for now, but at least avoid duplicating for both raw and processed

nmdefries (approved):

Issue #2097, opened 2025-01-13
[google-symptoms] Add conjunctivitis signal


carlynvandyke:
Per Roni's request (details to follow?), let's add a signal looking at search volume for [conjunctivitis/pink eye](https://www.aao.org/eye-health/tips-prevention/pink-eye-conjunctivitis-bird-flu-avian-influenza) in light of the H5N1 outbreak.

nmdefries:
@RoniRos Our other active `google-symptoms` signals are all [collections of related symptoms](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/google-symptoms.html#symptom-sets). For example, one for "runny nose" includes all the search terms Nasal congestion, Post nasal drip, Rhinorrhea, Sinusitis, Rhinitis, Common cold.

For this conjuctivitis signal, do you want it to include only a "conjuctivitis" search term or more general terms about eye irritation?

I'll look at available search terms and follow up with a list of related ones.

RoniRos:
Indeed we should include related terms.  The challenge is to include all those that are expected to sharpen a signal about conjunctivitis, without including those that would likely add noise.  I can't think of a realistic data driven way to to that, but maybe if you can do an initial pass over Google's terms and collect terms that are plausibly relevant, you and I can go over them and make an informed guess?  In parallel, we can try to learn more about the eye-related symptoms of H5N1.

RoniRos:
Check out[ what chatGPT 4o  says about it](https://chatgpt.com/share/67856db4-8b00-8000-91c8-8142bd172c96).

nmdefries:
*Common eye-related symptoms of H5N1 from various sources:* pink eye, conjunctivitis, tearing, teary eyes, eye redness, eye discharge, subconjunctival hemorrhage, eye discomfort, eye irritation, inflamed eyes, eye infection (from the American Optometric Association, American Academy of Ophthalmology, CDC, MyHealth Alberta, CDPH)

The chatGPT result additionally lists "itching or burning sensation", "swelling around the eyes", "blurred vision", and "sensitivity to light" (photophobia).

"itching..." is covered by "eye pain". I haven't seen the other three symptoms listed elsewhere, they may be uncommon or incorrect. "blurred vision" and "photophobia" are both available as search terms. "swelling around the eyes" might be considered a type of "angioedema" -- that search term is available but non-specific, so I don't think we have a good way to narrow down on it.

[...]

PR #2112, opened 2025-01-28
Define new conjunctivitis signal in `google-symptoms`

nmdefries:
### Description
Add new `s07` signal to `google-symptoms` to track novel eye-related symptoms of H5N1.

### Changelog
- Add new composite definition in `constants.py`
- Add to smoothed validator checks by adding new signal name to params templates.

### Associated Issue(s) 
Addresses https://github.com/cmu-delphi/covidcast-indicators/issues/2097


nmdefries:
We will need to patch this back to the beginning of `google-symptoms` history.

aysim319:
test data also needs to be updated to reflect the new signal

aysim319 (commented):
I tried running and I remember I kept the behavior of doing a full backfill on a regular run from previous code and it has potential cause some issues (would take multiple hours) and also would mess up the issue dates

aysim319 (changes requested):
I tried running and I remember I kept the behavior of doing a full backfill on a regular run from previous code and it has potential cause some issues (would take multiple hours) and also would mess up the issue dates

I can change the logic so it only logs and continues do the normal runs only. thoughts?

aysim319 (approved):

Issue #2116, opened 2025-02-03
Adding RSV for nhsn


aysim319:
previous scope for NHSN was limited for flu and covid, but external users have requested to also include RSV data for NHSN source.

PR #2121, opened 2025-02-12
first implimentation of rsv

aysim319:
### Description
add RSV for NHSN

### Associated Issue(s) 
- Addresses #2116


nhsn/tests/test_patch.py
```diff
@@ -13,7 +13,7 @@
 from delphi_nhsn.patch import filter_source_files, patch
 from delphi_nhsn.constants import TOTAL_ADMISSION_COVID_COL, TOTAL_ADMISSION_FLU_COL, \
     NUM_HOSP_REPORTING_FLU_COL, NUM_HOSP_REPORTING_COVID_COL, GEOS, TOTAL_ADMISSION_COVID, TOTAL_ADMISSION_FLU, \
-    NUM_HOSP_REPORTING_COVID, NUM_HOSP_REPORTING_FLU
+    NUM_HOSP_REPORTING_COVID, NUM_HOSP_REPORTING_FLU, NUM_HOSP_REPORTING_RSV_COL, TOTAL_ADMISSION_RSV_COL
```
nmdefries, line 16: suggestion: we don't seem to use any of the non-"_COL" `NUM_HOSP_REPORTING...` so remove those. I was confused why we weren't adding a "num_hosp_reporting" for RSV too.

```suggestion
    NUM_HOSP_REPORTING_RSV_COL, TOTAL_ADMISSION_RSV_COL
```

nhsn/tests/test_pull.py
```diff
@@ -146,7 +151,7 @@ def test_pull_prelim_nhsn_data_output(self, mock_create_backup, mock_socrata, ca
         # Mock Socrata client and its get method
         mock_client = MagicMock()
         mock_socrata.return_value = mock_client
-        mock_client.get.side_effect = [TEST_DATA, []]
+        mock_client.get.side_effect = [PRELIM_TEST_DATA, []]
```
nmdefries, line 154: praise: whoops, good find

question: if this wasn't failing, maybe the test should be stricter? Right now, it's comparing column names and if any values are missing. Should be also compare all values for equality in `result` vs `expected`?

nmdefries (approved):
This looks good.

We should do a brief statistical review on the new RSV signals. At minimum, we need national and state signals plotted over time, and RSV for some of those geos compared to flu and covid signals.

nhsn/tests/test_patch.py
aysim319: It was leftover from when i changed the variable names...

nmdefries:
Statistical review plots: [rsv_analysis.zip](https://github.com/user-attachments/files/18789033/rsv_analysis.zip)

Using one of those plots as an example,
Screenshot from 2025-02-13 14-12-17

RSV is shown as recently (in the current season) increasing maybe 50x above past values. This trend is shown in all plots.

Such a large increase is not believable and doesn't match other data sources, e.g.,

Screenshot from 2025-02-13 14-05-54

shows that hospitalizations this year are comparable to the last two years ([CDC RSV-NET](https://www.cdc.gov/rsv/php/surveillance/rsv-net.html), go to "Patient Characteristics" -> "Case Counts").

Emergency department admissions per capita show the same trend.

Screenshot from 2025-02-13 14-16-56

([CDC RSV-NET](https://www.cdc.gov/rsv/php/surveillance/rsv-net.html), select "All seasons" on the side bar).

Percent of emergency department admissions attributable to RSV (not shown) also have the same trend.

I'm assuming something went wrong with our handling of the data. It's also possible that the source started reporting new streams (e.g. previously only adults, now added in children), but honestly the increase is so large that this seems unlikely.

We need to

- Look at the raw data from the source to check values
- Compare those values to values reported by [CDC RSV-NET](https://www.cdc.gov/rsv/php/surveillance/rsv-net.html) -- I think they have national and state
- Same for our processed values

aysim319:
> This looks good.
> 
> We should do a brief statistical review on the new RSV signals. At minimum, we need national and state signals plotted over time, and RSV for some of those geos compared to flu and covid signals.

@nmdefries are we waiting to see how the statistical review before merging?

aysim319:
> Statistical review plots: [rsv_analysis.zip](https://github.com/user-attachments/files/18789033/rsv_analysis.zip)
> 
> Using one of those plots as an example, Screenshot from 2025-02-13 14-12-17
> 
> RSV is shown as recently (in the current season) increasing maybe 50x above past values. This trend is visible in all plots.
> 
> Such a large increase is not believable and doesn't match other data sources, e.g.,
> 
> Screenshot from 2025-02-13 14-05-54
> 
> shows that hospitalizations this year are comparable to the last two years ([CDC RSV-NET](https://www.cdc.gov/rsv/php/surveillance/rsv-net.html), go to "Patient Characteristics" -> "Case Counts").
> 
> Emergency department admissions per capita show the same trend.
> 
> Screenshot from 2025-02-13 14-16-56
> 
> ([CDC RSV-NET](https://www.cdc.gov/rsv/php/surveillance/rsv-net.html), select "All seasons" on the side bar).
> 
> Percent of emergency department admissions attributable to RSV (not shown) also have the same trend.
> 
> I'm assuming something went wrong with our handling of the data. It's also possible that the source started reporting new streams (e.g. previously only adults, now added in children), but honestly the increase is so large that this seems unlikely. Edit: possibly a plotting bug?
> 
> We need to
> 
> * Look at the raw data from the source to check values
> * Compare those values to values reported by [CDC RSV-NET](https://www.cdc.gov/rsv/php/surveillance/rsv-net.html) -- I think they have national and state
> * Same for our processed values

one of the main issues is that at least in NHSN data, people didn't report RSV until 2023/2024 that's the main reason for zero, and then peaking in the tail end

there's also spearman through time
Screenshot 2025-02-13 at 5 43 51 PM

Once I filter out the dates the comparison does seem more reasonable -- minus the plotting bug
not sure how to fix it
Screenshot 2025-02-13 at 5 48 35 PM

national level looks fine
Screenshot 2025-02-13 at 5 49 03 PM

nmdefries:
> in NHSN data, people didn't report RSV until 2023/2024 that's the main reason for zero

Ah, I see, it was because of the reporting mandate changes:

> All RSV data through October 31, 2024 represent voluntarily reported data; as such, all voluntarily reported data included in this dataset represent reporting hospitals only for a given week and might not be complete or representative of all hospitals during the specified reporting periods.

from the [NHSN data description](https://data.cdc.gov/Public-Health-Surveillance/Weekly-Hospital-Respiratory-Data-HRD-Metrics-by-Ju/ua7e-t2fy/about_data). We'll need to note this in the documentation, because the numbers prior to this season are not useful.

The national numbers match RSV-NET well for the most recent season. (RSV-NET is a representative sample covering only 9% of the population, so I multiplied it by 10 to compare values.)

This is good to go.

Issue #2129, opened 2025-02-27
NSSP data for HRRs is not summed properly


melange396:
[it is computed with a basic sum](https://github.com/cmu-delphi/covidcast-indicators/blob/454ac565d0a0f2b5cf557e4efb2278c278c528a9/nssp/delphi_nssp/run.py#L119-L122) instead of a weighted one like the other geo types we aggregate ourselves.  as a result the scale is off and can be >100% (see `nssp:pct_ed_visits_influenza` @`hrr:366` )

minhkhul:
Probably adjust some wording/add warning in geomapper here too. The original instruction was why I implemented weighting for msa but not hr.
 
https://github.com/cmu-delphi/covidcast-indicators/blob/454ac565d0a0f2b5cf557e4efb2278c278c528a9/_delphi_utils_python/delphi_utils/geomap.py#L99-L104

melange396:
I just did another read-through of some GeoMapper code...  It looks like `.replace_geocode()` can/should produce weighted sums for certain geo-level conversions (through a sneaky and convoluted path), but the [precomputed fips-->hrr weights do not add up the way they should](https://github.com/cmu-delphi/covidcast-indicators/blob/454ac565d0a0f2b5cf557e4efb2278c278c528a9/_delphi_utils_python/delphi_utils/data/2020/fips_hrr_table.csv?plain=1#L3955-L3966).  Using your other approach that calls `.aggregate_by_weighted_sum()` w/ population should get around this.

dshemetov:
I'm pretty familiar with GeoMapper, having written much of the current core logic in #217 and in #1960, so I figure I should provide some context. I don't think the fips->hrr weights are wrong. I think the TL;DR of the problem here is that

- `replace_geocode()` is appropriate for aggregating unnormalized counts data and implicitly considers missing geos to be zero
- `aggregate_by_weighted_sum()` is intended for normalized data and implicitly extrapolates to missing geos from present neighbor geos

We should definitely use `aggregate_by_weighted_sum()` in this code path, not sure why we didn't originally.

Context: 

("Source geo" and "target geo" below refer to the geos being translated. So for the bug in question, source geos are fips and target geos are hrrs).

[...]

PR #2131, opened 2025-03-03
2129 fix hrr for nssp

aysim319:
### Description
cleaning up test and fixing geo aggregation for hrr

### Associated Issue(s) 
- Addresses #2129


nssp/delphi_nssp/run.py
```diff
@@ -117,11 +117,14 @@ def run_module(params, logger=None):
...
                 df = df[["fips", "val", "timestamp"]]
-                # fips -> hrr has a weighted version
-                df = geo_mapper.replace_geocode(df, "fips", "hrr")
-                df = df.rename(columns={"hrr": "geo_id"})
+                df = geo_mapper.add_population_column(df, geocode_type="fips", geocode_col="fips")
+                df = geo_mapper.add_geocode(df, "fips", "hrr", from_col="fips", new_col="geo_id")
+                df = geo_mapper.aggregate_by_weighted_sum(df, "geo_id", "val", "timestamp", "population")
+                df = df.rename(columns={"weighted_val": "val"})
             elif geo == "msa":
+                df = df[df["county"] != "All"]
```
aysim319, line 127: @nmdefries the fact that this wasn't included in the previous runs kinda makes me wonder about data quality.... I tried to dig up where the original one for the statical analysis was done and I couldn't find it this time.

nssp/raw_data_backups/.gitignore
melange396, line 1: removing this file will cause the directory to become empty, which will then cause git to delete the directory...  is that the intention?

nssp/delphi_nssp/run.py
melange396: This isnt necessary.  The only rows with county=="All" are national-level or for entire states, and these all have FIPS codes that are integer multiples of 1000 (the full USA data is given with FIPS code "0" in the source data set, which is technically also a multiple of 1000).  The GeoMapper knows that none of these FIPS codes are part of any [HRR](https://github.com/cmu-delphi/covidcast-indicators/blob/main/_delphi_utils_python/delphi_utils/data/2019/fips_hrr_table.csv?plain=1) or [MSA](https://github.com/cmu-delphi/covidcast-indicators/blob/main/_delphi_utils_python/delphi_utils/data/2019/fips_msa_table.csv?plain=1).

melange396 (changes requested):

melange396 (commented):
the logic in the `elif geo == "hrr":` section looks good...  can you describe whats going on with the changes to tests and with the txt/json files?

minhkhul (approved):
lgtm! Appreciate the addition of full run module tests. 
[nonblocking] suggesting some extra docstrings for clarity.

Issue #2137, opened 2025-03-24
nhsn skips run if data updated after scheduled run


aysim319:
on march 19th, nhsn was scheduled to run on 12:30pm EST, but the update for the preliminary nhsn data didn't arrive until 12:54pm EST. The next scheduled date, being 21st, skipped the preliminary data being more than a day old. 

need to patch preliminary data and also adjust the staleness deadline

aysim319:
Another incident happened where the main data was updated 18 minutes later than usual.

PR #2142, opened 2025-04-07
using epidata metadata to determine stalenes for nhsn pulling

aysim319:
### Description
using metadata api to determine staleness

### Associated Issue(s) 
- Addresses #2137


nhsn/tests/test_pull.py
```diff
@@ -169,8 +170,9 @@ def test_check_last_updated(self, mock_socrata, dataset, updatedAt, caplog):
         check_last_updated(mock_client, dataset["id"], logger)
 
         # Check that get method was called with correct arguments
-        now = time.time()
-        if now - updatedAt < 60:
+        now_datetime = datetime.utcfromtimestamp(time.time())
```
minhkhul, line 173: nit: My Pylance was complaining that utcfromtimestamp is deprecated as noted here https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp and that `datetime.fromtimestamp(timestamp, tz=timezone.utc)` should be used instead.

minhkhul (approved):
LGTM. One nit was the utcfromtimestamp thing but everything else looks good.

nhsn/delphi_nhsn/pull.py
```diff
@@ -38,17 +47,31 @@ def check_last_updated(socrata_token, dataset_id, logger):
...
+
+        # pulling last updated from the api
+        meta_df = pd.DataFrame(Epidata.covidcast_meta()["epidata"])
+        signal_suffix = "prelim" if dataset_id == PRELIM_DATASET_ID else "ew"
+        nhsn_meta_df = meta_df[(meta_df["data_source"] == "nhsn") & (meta_df["signal"].str.endswith(signal_suffix))]
+        est = timezone(timedelta(hours=-5))
+        last_updated = datetime.fromtimestamp(nhsn_meta_df["last_update"].min(), tz=est)
+
+        # currently set to run twice a week, RECENTLY_UPDATED_DIFF may need adjusting based on the cadence
+        recently_updated_source = (updated_timestamp - last_updated) > RECENTLY_UPDATED_DIFF
```
melange396, line 60: i dont think this math is quite right...  why wouldnt we want to proceed any time socrata has a newer timestamp than we do?  the form you have here has the potential to delay processing if updates are frequent enough or if `RECENTLY_UPDATED_DIFF` is too large.
```suggestion
        socrata_ts = updated_timestamp
        delphi_ts = last_updated
        recently_updated_source = socrata_ts > delphi_ts
```

melange396 (commented):
in theory, i really like the idea of using our own metadata, but in practice, its going to have problems that make it unworkable.  The biggest is that there is an indeterminate delay between data getting inserted into the database and when the metadata is updated to include that data (the delay is bounded, but long enough to throw a wrench in this).  Another is that patches will affect the metadata, thereby disrupting scheduling.

PR #2175, opened 2026-02-25
version pinning osqp==1.1.0 because 1.1.1 breaks the build

melange396:
`osqp` is a dependency of `cvxpy` which is a dependency of `delphi-utils` which gets imported all over the place :-/

pursuant to https://github.com/cmu-delphi/epidata-etl/issues/594


korlaxxalrok (approved):
Looks good to me 👍

aysim319 (approved):

PR #2184, opened 2026-08-18
Make date/time separator optional in claims filenames

minhkhul:
Fixes the claims filename handling that broke the `claims_hosp` pipeline in prod on 2026-08-17, and ports [epidata-etl#886](https://github.com/cmu-delphi/epidata-etl/pull/886). Applied to both `claims_hosp` and `doctor_visits`, which carry identical copies of this code.

### The prod failure

The FTP served `EDI_AGG_INPATIENT_08162026_0209CDT.csv.gz` — an **MMDDYYYY** date. The download succeeded; the rename is what broke. `change_date_format` assumed YYYYMMDD unconditionally:

```python
flip_date = date[6:] + date[4:6] + date[:4]   # "08162026" -> "26200816"
```

so drops landed as `..._26200816_...`, and `get_latest_filename` died on one of them:

```
ValueError: unconverted data remains: 15
  get_latest_claims_name.py:20 in get_latest_filename
```

(`"262008071415"` under `%d%m%Y%H%M` consumes `%d`=26, `%m`=2, `%Y`=0081, `%H`=6, `%M`=02, then chokes.)

### Changes

**1. Date flip handles MMDDYYYY** — using the same `> 1240` year heuristic as the upstream PR, which `get_timestamp` already relied on implicitly. `08162026` now becomes `16082026`, which parses. This is the fix for the prod crash.

**2. Separator between date and time is optional** — the actual upstream port. `EDI_AGG_INPATIENT_060620260000CDT.csv.gz` is now a valid drop name; previously the regex rejected it outright, so the file got a 1900-01-01 timestamp and was silently skipped at download. `change_date_format` reinserts the separator so local files keep the `..._DDMMYYYY_HHMM{tz}.csv.gz` shape that `get_latest_filename` and `run.py` parse.

```python
FILENAME_TIMESTAMP = re.compile(
    r".*EDI_AGG_INPATIENT_(?:[0-9]_)?(?P<ymd>[0-9]{8})_?(?P<hm>[0-9]{4})[^0-9]*")
```

The `len(name.split("_")) > 5` dispatch had to go: the new name is 4 fields and a chunked no-separator name is 5, so the field count no longer distinguishes the shapes. Fixed widths make the chunk and the separator unambiguous instead. Chunked names still pass through the rename untouched, as before.

**3. `get_latest_filename` skips drops whose timestamp won't parse**, logging a warning. This one is load-bearing, not defensive polish: `retrieve_files/` still holds the names the old flip mangled, and `get_latest_filename` walks the whole directory every run. Without this, tomorrow's run names today's drop correctly and then still dies on a leftover from last week. Now those files are ignored and a single bad name degrades a run to the existing "no drop for today" assert instead of a stack trace.

Recovering the mangled files is a backfill concern and is out of scope here. They stay unusable and noisy in the logs until that happens, but they no longer break anything.

### Testing

Test coverage added for both indicators (`doctor_visits` had none for the download helpers). I could **not** run pytest locally — `claims_hosp/env` and `doctor_visits/env` are dead symlinks to a `/opt/homebrew/bin/python3.8` that no longer exists, and system 3.8 has no numpy/pytest — so CI is the first real run of the suite.

What I did verify, by exercising the functions directly: old, current, and new-format filenames all parse and rename correctly; and end-to-end against the two filenames from the failing run, both rename correctly, an unrepaired leftover is warned-and-skipped, and the 14:09 drop is selected as latest.

Stays Python 3.8 compatible; no new dependencies.


minhkhul:
The FTP served `EDI_AGG_INPATIENT_08162026_0209CDT.csv.gz` — MMDDYYYY, separator present, so the download was fine. The rename mangled it to `..._26200816_...`, and `get_latest_filename` died on an older mangled leftover:

```
ValueError: unconverted data remains: 15
  get_latest_claims_name.py:20 in get_latest_filename
```

`"262008071415"` under `%d%m%Y%H%M` consumes `%d`=26, `%m`=2, `%Y`=0081, `%H`=6, `%M`=02 and chokes on the trailing `15`. The `> 1240` heuristic in the first commit produces `16082026` instead, which parses.

Two additions, because the rename fix alone does not restart prod:

- **`get_latest_filename` skips unparseable names** instead of raising. `retrieve_files/` is already full of mangled names; without this, the corrected rename changes nothing, since the run still dies on the leftovers. One bad name now degrades to the existing "no drop for today" assert plus a warning.
- **`claims_hosp/repair_claims_filenames.py`** renames the files already on disk (`26200807` → `07082026`). Dry-run by default, `--apply` to commit. It only touches names that fail to parse as DDMMYYYY *and* whose repair parses, so already-correct names, chunked names, and non-drops are left alone and reruns are no-ops.

End-to-end check against the two filenames from the failing run: both rename correctly, an unrepaired leftover is warned-and-skipped, and the 14:09 drop is selected as latest.

Deploy order: repair script first (or at least before the next scheduled run), then the code. The corrected flip maps recent drops to different names than what is on disk, so anything still inside the 24h window gets re-fetched under its correct name — a couple of duplicate files to clean up.

minhkhul:
Dropped `repair_claims_filenames.py` — recovering the already-misnamed drops is a backfill concern and will be handled separately. Nothing in this PR is a manual step anymore; it is all code that runs on its own.

The skip in `get_latest_filename` stays, and it is load-bearing rather than defensive polish: `retrieve_files/` still holds the mangled `26200xxx` names, and without the skip the run keeps dying on them even though new downloads are named correctly. Those files are now ignored with a warning each run until someone cleans them up.

nolangormley (approved):

PR #2187, opened 2026-08-19
Parse service dates explicitly instead of via parse_dates

minhkhul:
Fixes the `doctor_visits` prod failure on the 2026-08-18 drop:

```
TypeError: '>=' not supported between instances of 'str' and 'datetime.datetime'
  update_sensor.py:120 in update_sensor
```

### Cause

`read_csv`'s `parse_dates` gives up **silently** when a column contains anything it cannot convert — it hands back the raw strings for the entire column rather than raising. So `ServiceDate` arrived as `str` and died on the first comparison against a datetime, 30 lines after the actual problem.

Today's outpatient drop supplies two things it can't convert:

- **77,283 leading rows with a blank service date** (`""`, usually with blank FIPS too), before the dated rows start.
- **One row dated in the year 2608.** `datetime64[ns]` only spans 1677-09-21 to 2262-04-11, so that single row out of 97.7M is enough to turn the whole column into strings. It is invisible to any format check — every value in the column is ISO-shaped.

### Fix

Parse the column explicitly with `to_datetime(errors="coerce")`, so unusable values become `NaT` and are dropped, and everything else is a real timestamp. `doctor_visits` logs how many rows it drops; `claims_hosp` relies on its existing date-range filter to discard them.

The format is pinned to `%Y-%m-%d` (via a new `Config` constant, alongside the existing date column config) rather than left to inference. Without a `format`, `to_datetime` guesses one from the first non-null value and falls back to per-element `dateutil` parsing when that guess doesn't hold for the rest of the column — on 97.7M rows that fallback is the difference between ~10 seconds and roughly an hour, with nothing in the run to say why. Pinning it means an unexpected format costs a column of `NaT`, which the coerce and date-range filters already handle and `doctor_visits` already logs, instead of a silent stall.

`claims_hosp` loads its drops with the identical `dtype` + `parse_dates` pattern, so it is fixed too — the inpatient feed is generated by the same process and is a crash waiting to happen.

### Testing

Verified against the real 97,707,136 row drop pulled from the FTP server, on pandas 2.0.3 / Python 3.8 to match prod:

| | `ServiceDate` dtype |
|---|---|
| as shipped | `object` → reproduces the prod `TypeError` |
| with this fix | `datetime64[ns]`, 77,284 rows dropped, 93,968,049 kept |

Also benchmarked, since replacing `parse_dates` with a separate pass looks like it should cost something. On a synthetic 10M row drop matching the real column set and `DTYPES`:

| | read_csv | to_datetime | dropna | total |
|---|---|---|---|---|
| `parse_dates` (clean file) | 7.3s | — | — | **7.3s** |
| this PR (clean file) | 4.2s | 1.0s | 0.0s | **5.2s** |
| this PR (file with bad dates) | 4.6s | 0.6s | 0.5s | **5.7s** |

`read_csv` gets cheaper when the date parsing comes off it, by more than the separate pass costs, so this is a small net speedup rather than a regression. Pinning the format is a wash on the happy path (0.9s vs 1.0s); its value is bounding the bad path. Scaled to the real drop that is ~10s of `to_datetime` against a `read_csv` that already costs ~45s and a sensor-fitting stage that dominates both.

Regression test added for `claims_hosp` covering the blank-date case. `doctor_visits` has no unit-level coverage of `update_sensor`'s loading path, and building a fixture for it is a bigger job than this fix warrants — flagging rather than silently skipping.

Incidentally: on pandas 2.2 the pre-PR code returns `object` dtype even for a perfectly clean file, because `dtype={"ServiceDate": str}` and `parse_dates=[...]` conflict and the explicit dtype wins. Prod is on 2.0.3 where `parse_dates` wins, but neither `setup.py` pins pandas, so this was going to start failing on every drop and not just malformed ones.

Note the drop also contains dates as early as 1753 and as late as 2121, which are in range and parse fine; the existing `FIRST_DATA_DATE` and `dropdate` filters discard them as before.

### DIFF
--- a/claims_hosp/delphi_claims_hosp/config.py
+++ b/claims_hosp/delphi_claims_hosp/config.py
@@ -38,6 +38,7 @@ class Config:
     # data columns
     CLAIMS_COUNT_COLS = ["Denominator", "Covid_like", "Flu1"]
     CLAIMS_DATE_COL = "ServiceDate"
+    CLAIMS_DATE_FORMAT = "%Y-%m-%d"  # drops carry ISO service dates
     FIPS_COL = "fips"
     DATE_COL = "timestamp"
     AGE_COL = "age_group"
--- a/claims_hosp/delphi_claims_hosp/load_data.py
+++ b/claims_hosp/delphi_claims_hosp/load_data.py
@@ -34,8 +34,12 @@ def load_claims_data(claims_filepath, dropdate, base_geo):
     # Drops carry blank service dates and sentinels outside the datetime64 range
     # (e.g. 1753-01-01), which make read_csv's parse_dates give up and hand back
     # the raw strings for the whole column. Parse it ourselves so those rows turn
-    # into NaT and are dropped by the date range filter below.
-    claims_data[Config.CLAIMS_DATE_COL] = pd.to_datetime(claims_data[Config.CLAIMS_DATE_COL], errors="coerce")
+    # into NaT and are dropped by the date range filter below. The format is
+    # pinned so an unexpected one costs a column of NaT rather than silently
+    # falling back to per-element dateutil parsing over the whole drop.
+    claims_data[Config.CLAIMS_DATE_COL] = pd.to_datetime(
+        claims_data[Config.CLAIMS_DATE_COL], format=Config.CLAIMS_DATE_FORMAT, errors="coerce"
+    )
 
     # standardize naming
     claims_data.rename(columns=Config.CLAIMS_RENAME_COLS, inplace=True)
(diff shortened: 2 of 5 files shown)


minhkhul:
The exact offending row, for the record — line 58,211,887 of `EDI_AGG_OUTPATIENT_08182026_1411CDT.csv.gz`:

```
2608-08-03,36061,NJ - Hackensack,284,20-24,1.0,0.0,0.0,0.0,0.0
```

One row out of 97,707,136, with a valid FIPS, a real HRR, and a plausible count — the only thing wrong with it is a service date 582 years in the future, which is past what `datetime64[ns]` can represent. That alone silently turned the entire `ServiceDate` column into strings.

Worth passing to Optum alongside the 77,283 blank-date rows: both are upstream data-quality issues. This PR keeps the pipeline running through them rather than fixing them at the source.

dshemetov:
> the only thing wrong with it is a service date 582 years in the future

Wow, that's quite a data error! :upside_down_face:

dshemetov (approved):
Seems like a fine hotfix and will make the date parsing on DATE_COL more robust

